PR #20, opened 2026-04-16
feat: set default HSTS and CSP headers for direct-exposure deployments (#14)

### LINKED ISSUE #14: Add default HSTS and CSP headers for direct-exposure deployments
## Summary

\`Strict-Transport-Security\` and \`Content-Security-Policy\` headers are currently delegated to a reverse proxy. For self-hosted/OSS deployments where the server may be exposed directly, sensible defaults should be set by the API server.

## Recommendation

Add HSTS and CSP with defaults that can be overridden by proxy headers or configuration. Document the recommended reverse proxy setup.

## Priority

P3 — deployment hardening

mgoldsborough:
## Summary

- Previous policy of delegating HSTS/CSP to a reverse proxy left direct-exposure self-hosted deployments with no hardening. Direct-HTTPS is a supported topology, so defaults should protect it.
- Added conservative defaults in \`securityHeaders()\`:
  - HSTS: \`max-age=31536000; includeSubDomains\` (no \`preload\` — that is a deliberate operator choice)
  - CSP: \`default-src 'none'; frame-ancestors 'none'; base-uri 'none'\` (does not affect JSON/SSE; bundle UI HTML is consumed via \`fetch+srcdoc\` so the response CSP does not break the iframe bridge)
- Overrideable via \`NB_HSTS\` / \`NB_CSP\` env vars or middleware options. Env var takes precedence over option. Empty string disables the header — preserves the "delegate to proxy" behavior for operators who want it.

## Test plan

- [x] 13 unit tests covering defaults, option override, env var override, empty-string disable, POST and 404 paths
- [x] \`bun run lint\` / \`bun run check\` — clean
- [x] README updated with new env var documentation

Closes #14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - `manage_workspaces.list` now returns the requesting user's role within each workspace (`userRole?: "admin" | "member"`) so the web client can gate workspace-admin UI without an extra `list_members` round-trip.
 - `POST /v1/chat` validation error wording. Hand-rolled checks (`"metadata must be a JSON object"`, `"allowedTools must be an array of strings"`) replaced with TypeBox path-prefixed errors (`"/metadata: Expected object"`, `"/allowedTools: Expected array"`). HTTP status (400) and `error: "bad_request"` are unchanged. External callers asserting on the exact wording need to update; the field name still appears in the message.
 - Skill manifest writes now always include `metadata.keywords` and `metadata.triggers` as arrays (defaulting to `[]` when omitted by the caller). Previously a partial `metadata: { category: "X" }` could write a manifest with no `keywords`/`triggers` keys at all; the loader's domain type required them, so the divergence was a latent type lie. The on-disk JSON shape is now what the type always claimed.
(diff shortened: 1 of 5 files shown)


mgoldsborough:
Rebased onto current main. One adjustment from the original PR:

The proxy route in main strips upstream CSP and asserts `Content-Security-Policy` is `null` on iframed bundle responses. The PR's global default CSP `default-src 'none'` would block iframed bundle dev-server scripts/styles — and the test caught it.

Resolution: middleware now respects route-level intent for HSTS/CSP using the same pattern as X-Frame-Options. The proxy route opts out by setting an internal `X-NB-Skip-Security-Defaults` header, which the middleware strips before egress. The parent shell's `frame-ancestors 'none'` remains the real protection vector for iframed responses; CSP on the iframe content itself isn't the boundary that matters here.

Diff now touches `src/api/routes/proxy.ts` (1 import + 1 `outHeaders.set`) in addition to the original middleware changes. `bun run verify` passes.

PR #19 (the workspace generic-error PR) was also rebased — clean, no adjustments needed.

PR #21, opened 2026-04-16
feat: enforce per-file size limit in files.write tool handler (#12)

### LINKED ISSUE #12: Add file size limit enforcement at tool handler layer
## Summary

File upload limits are configured at the API layer, but tool handlers that write files from within the agentic loop don't independently enforce size limits. Adding a size check before writing provides defense-in-depth against resource exhaustion.

## Priority

P3 — defense-in-depth

mgoldsborough:
## Summary

- The API-layer \`bodyLimit\` middleware caps HTTP uploads, and \`src/files/ingest.ts\` enforces \`maxFileSize\` on the user-upload path. The \`files__write\` tool handler reached by the agentic loop does not traverse either — so the "no file larger than \`maxFileSize\`" invariant did not hold at the point of write.
- Plumbed \`runtime.getFilesConfig().maxFileSize\` into \`handleWrite\` and reject the call with a descriptive error when the decoded payload exceeds it.
- Reads from runtime config (not a hard-coded constant) so operators who raise the limit in \`nimblebrain.json\` see it applied consistently across the HTTP and tool-call paths.

## Test plan

- [x] New \`test/unit/tools/platform-files.test.ts\` — 3 tests covering accept-within-limit, reject-over-limit (error message includes filename and mentions the limit), and respects runtime config override
- [x] \`bun run lint\` / \`bun run check\` — clean

Closes #12

### DIFF
--- a/src/tools/platform/files.ts
+++ b/src/tools/platform/files.ts
@@ -202,11 +202,27 @@ interface WriteInput {
   description?: string;
 }
 
-function handleWrite(registryPath: string, filesDir: string, args: WriteInput): object {
+function handleWrite(
+  registryPath: string,
+  filesDir: string,
+  args: WriteInput,
+  maxFileSize: number,
+): object {
   ensureDir(filesDir);
 
   const id = generateId();
   const decoded = Buffer.from(args.base64_data, "base64");
+
+  // Defense-in-depth: enforce the configured per-file size limit at the point
+  // of write. The API body-limit middleware caps HTTP uploads, but
+  // agent-initiated tool calls reach this function without traversing that
+  // middleware — so the limit must hold here independently.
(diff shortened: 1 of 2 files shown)


mgoldsborough:
Closing — superseded by main's store-unification refactor. The file got rewritten: `files__write` → `files__create`, args shape changed from flat snake_case to `{ manifest, body }`, and the handler now delegates to `FileStore.saveFile()` instead of inline `writeFileSync`. Patch no longer applies and tests won't compile against the new signature.

Underlying gap from #12 is still real (agent-initiated writes bypass both `bodyLimit` middleware and `src/files/ingest.ts` size enforcement; `FileStore.saveFile()` doesn't check size). Issue #12 stays open; will redo on top of main.

PR #27, opened 2026-04-16
Scope bundle instances and placements by workspace ID

mgoldsborough:
## Summary

Fixes a cross-workspace data leak and a nav-visibility bug, both rooted in single-tenant residue that survived the migration to multi-tenant workspaces. The codebase had two structures claiming to know "what's installed where": `lifecycle.instances` (compound `${serverName}|${wsId}` key) and per-workspace `ToolRegistry` (keyed by `serverName`). The read-side joined them by `serverName` only, so when two workspaces installed the same bundle the filter returned both instances. Each instance carries its own `entityDataRoot`, so downstream consumers — briefing facet collector, apps list — read entity data from every matching workspace and surfaced it to the caller.

Three related defects, all fixed in this PR:

- **`Runtime.getBundleInstancesForWorkspace`** now requires `inst.wsId === wsId` in addition to registry visibility. Global/protected instances (`wsId === undefined`) remain visible in every workspace.
- **`PlacementRegistry.register`/`unregister`** are scoped by `wsId`. The previous global-per-`serverName` unregister caused re-registration in a second workspace to silently wipe the first workspace's nav entries (last seeded won). `unregister(name, "*")` removes across all workspaces for full-bundle uninstalls.
- **Install/lifecycle APIs** (`installNamed`/`installLocal`/`installRemote`/`uninstall`/`startBundle`/`stopBundle`/`recordCrash`/`recordRecovery`/`recordDead`) now require `wsId`. Instances are stored under compound keys, `inst.wsId` is populated on the record, and `installNamed` scopes its bundle data dir to `{workDir}/workspaces/{wsId}/data/{bundle}` to match `seedInstance`.

## Test plan

- [x] `bun run check` — clean
- [x] `bun run lint` — clean (only pre-existing import-type warnings)
- [x] `bun run test:unit` — 1725 pass, 0 fail
- [x] `bun run test:integration` — failure set identical to `main`; remaining 28 failures are a pre-existing echo-MCP subprocess env issue, verified by diff against the base branch
- [x] New regression tests in `test/integration/workspace-security.test.ts` (14 pass):
  - PlacementRegistry: registering the same bundle in a second workspace preserves the first's entries
  - PlacementRegistry: `unregister` is scoped to the given workspace
  - PlacementRegistry: re-registering a global source does not wipe workspace entries
  - BundleLifecycleManager: two workspaces seeded with the same bundle keep distinct `entityDataRoot` values

## Follow-ups (separate PR)

The underlying conflation that made this bug possible — `BundleInstance` mixing catalog-level fields (name, version, manifest) with runtime-level per-workspace fields (state, wsId, entityDataRoot) — is worth splitting into `BundleCatalogEntry` (global) and `WorkspaceBundleInstance` (per-ws). That change would force the scope question at the type level rather than leaving it to discipline, preventing the same class of bug from reappearing in a new projection (automations, health-monitor attribution, etc.). Flagging but out of scope for this hotfix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,30 @@
 
 ## [Unreleased]
 
+### Security
+
+- **Scope bundle instances and placements by workspace.** When two workspaces had the same bundle installed, `Runtime.getBundleInstancesForWorkspace` returned instances from both (filtering only by `serverName`), causing briefing facets and the apps list to read entity data from other workspaces. `PlacementRegistry.unregister` was also global-per-serverName, so re-seeding in a second workspace silently wiped the first workspace's nav entries. Both paths are now workspace-scoped.
+
+### Breaking (internal API)
+
+Downstream forks or consumers that extend `BundleLifecycleManager` will need to update callsites:
+
+- `installNamed(name, registry, env?)` → `installNamed(name, registry, wsId, env?)`
+- `installLocal(bundlePath, registry, env?)` → `installLocal(bundlePath, registry, wsId, env?)`
+- `installRemote(url, serverName, registry, transportConfig?, ui?, trustScore?)` → `installRemote(url, serverName, registry, wsId, transportConfig?, ui?, trustScore?)`
+- `uninstall(nameOrPath, registry)` → `uninstall(nameOrPath, registry, wsId)`
+- `startBundle(serverName, registry)` → `startBundle(serverName, wsId, registry)`
+- `stopBundle(serverName, registry)` → `stopBundle(serverName, wsId, registry)`
+- `recordCrash(serverName)` / `recordRecovery(serverName)` / `recordDead(serverName)` each gain a required `wsId` second argument.
+- `BundleInstance.wsId` is now required (was optional). Every instance belongs to exactly one workspace; global/platform sources are represented as `InlineSource`, not `BundleInstance`.
+- `PlacementRegistry.unregister(serverName, wsId?)` now scopes to `(serverName, wsId)` only. Passing no `wsId` removes only global entries; passing a specific `wsId` leaves other workspaces untouched.
+
+### Migration
+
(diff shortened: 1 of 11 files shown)


mgoldsborough:
Addressed QA review items in commit 02c9c72:

**Warnings**
- ✅ CHANGELOG entry added covering the security fix, the breaking internal API surface on `BundleLifecycleManager`, and the `installNamed` data-dir path migration.

**Suggestions**
- ✅ **Direct regression test for the fix site.** New tests in `test/integration/security/workspace-isolation.test.ts` under §V6 spin up `Runtime.start`, seed the same serverName into two workspaces, register matching sources in each workspace's registry (so the pre-fix `visible.has(serverName)` condition holds for both), and assert `getBundleInstancesForWorkspace` returns exactly one instance with the correct `wsId` and `entityDataRoot`. Also covers the orphaned-lifecycle case (instance in lifecycle, no source in registry).
- ✅ **`"*"` sentinel.** Rather than test it, I removed it. Zero callers today — preemptively added bandaid. If a future caller legitimately needs "unregister across all workspaces," a clearly-named second method is the right shape; a magic-string overload is not.
- 🗒️ **`entityDataRoot` not populated on install paths.** Tracked for the `BundleInstance` type-split follow-up PR (where `entityDataRoot` belongs on `WorkspaceBundleInstance`, not the catalog half).
- ✅ **Type split.** Remains the next PR.

**Additional first-principles tightening (not in review but called out in earlier audit)**
- `BundleInstance.wsId` is now **required** (was optional). The optional shape was the type-level enabler of the leak — nothing was constructing instances without a wsId in practice, so the `undefined` branch in `getBundleInstancesForWorkspace` was dead code. Removed that branch and the optional type; the filter is now simply `inst.wsId === wsId && visible.has(inst.serverName)`.

**Verification**
- `bun run test:unit` — 1725 pass / 0 fail
- `test/integration/workspace-security.test.ts` + `test/integration/security/workspace-isolation.test.ts` — 27 pass / 0 fail (the 13 new checks include the direct Runtime fix-site test)

Issue #2, opened 2026-04-14
Refactor identity to be per-workspace


mgoldsborough:
## Problem

Agent identity (`identity.md`) is currently a single global file at `{workDir}/identity.md`. It's loaded at startup and on reload as a synthetic context skill (priority 1) and injected into the system prompt for **every conversation, every workspace, every user** on the runtime instance.

The workspace layer (`src/workspace/`) has zero awareness of identity. Workspaces have isolated bundles, conversations, and members — but the agent persona is shared globally.

In a multi-tenant deployment, this doesn't work. A law firm workspace and a marketing workspace need different agent personas.

## Current Flow

```
buildSkills() / reloadSkills() (runtime.ts:1447-1460, 826-835)
  → loadIdentityOverride(workDir) (runtime.ts:1474-1494)
    → reads {workDir}/identity.md
    → returns synthetic Skill { type: "context", priority: 1 }
  → injected into contextSkills for ALL requests
```

Write path: `nb__manage_identity` (core-source.ts:414-478) and `settings__identity` both write to the same global file.

## Proposed Change

Move identity to the workspace store so each workspace gets its own persona:

1. **Extend workspace schema** — Add an `identity` field (markdown string) to the workspace record in `WorkspaceStore`.

2. **Make skill loading workspace-aware** — `buildSkills()` and `reloadSkills()` currently operate globally. The identity override needs to be resolved per-request based on the active workspace context. The `chat()` method already has workspace context available.

3. **Update prompt composition** — `composeSystemPrompt()` receives context skills globally. It needs to accept a workspace-scoped identity skill instead.

4. **Migrate settings tool** — `settings__identity` should read/write the workspace's identity field, not the global file.

5. **Fallback chain** — `workspace.identity ?? global identity.md ?? default`. This preserves backward compatibility for single-workspace deployments.

6. **Admin gating** — Only workspace admins (or org admins) should be able to write workspace identity.

## Files to Modify

- `src/workspace/workspace-store.ts` — Add `identity` field to workspace schema
- `src/runtime/runtime.ts` — Make `loadIdentityOverride()` workspace-aware, update `buildSkills()` and `reloadSkills()`
- `src/prompt/compose.ts` — Accept workspace-scoped identity
- `src/tools/platform/settings.ts` — Read/write workspace identity instead of global file
- `src/tools/core-source.ts` — Remove `manage_identity` system tool (replaced by settings)

## Context

[...]

Issue #1, opened 2026-04-09
Support version pinning for bundles in nimblebrain.json


mgoldsborough:
## Problem

Bundle entries in `nimblebrain.json` only support `name` (registry) or `path` (local), with no way to pin a specific version:

```json
{ "name": "@nimblebraininc/synapse-collateral" }
```

The runtime calls `mpak.prepareServer({ name })` which always resolves to the latest published version. There's no mechanism to pin a bundle to a specific version (e.g., `0.2.0`) or constraint (e.g., `^0.2.0`).

## Impact

- **No reproducible deployments** — redeploying a tenant can silently pick up a new bundle version
- **No rollback path** — if a bundle release is broken, every tenant gets it on next restart
- **Tenant values can't lock versions** — `config.bundles[].version` in Helm values would be ignored

## Proposed Solution

Add an optional `version` field to bundle config:

```json
{ "name": "@nimblebraininc/synapse-collateral", "version": "0.2.0" }
```

Pass it through to the mpak SDK: `mpak.prepareServer({ name, version })`. The mpak SDK already supports version resolution — we just need to wire it through.

### Places to update

- `src/runtime/bundles.ts` — `buildRegistrySource()` reads `ref.version` and passes to `prepareServer`
- `src/bundles/lifecycle.ts` — same pattern for lifecycle-managed bundles
- `src/tools/system-tools.ts` — `installBundle()` / `installBundleViaLifecycle()`
- `src/config/nimblebrain.schema.json` — add `version` to bundle schema
- Helm chart ConfigMap template — pass `version` through from values

PR #40, opened 2026-04-18
feat(config): workspace credential store + permission hardening

mgoldsborough:
## Summary

- Adds `src/config/workspace-credentials.ts` — per-bundle credential files at `{workDir}/workspaces/{wsId}/credentials/{bundle-slug}.json`. Atomic writes, explicit `0o600`/`0o700` perms, defensive parsing.
- Hardens `scaffold.ts` and `workspace-store.ts` so the workspace root and `credentials/` subdirectory are created `0o700`, and `workspace.json` is written `0o600`.
- **No production wiring.** This PR is file-level primitives only — nothing reads from these yet. The resolver (task 003) and startup integration (task 005) come next.

## Context

Part of the credential-resolution restructuring. The goal is to satisfy a bundle's `user_config` requirements from a workspace-scoped store instead of `~/.mpak/config.json` (which is globally shared and doesn't support per-workspace values). The new store is tier 1 of the 4-tier hierarchy documented in `.tasks/credential-resolution/SPEC_REFERENCE.md`.

## Why split permissions + credential store into one PR

They're both file-level primitives with no production wiring. Reviewing the credential store without the hardening would risk shipping a credential directory that inherits umask-derived perms. Combined, the security story is complete in one unit.

## Test plan

- [x] 22 unit tests for `workspace-credentials.ts` — save/retrieve roundtrip, merge semantics, workspace isolation, slug derivation, `0o600`/`0o700` verification via `fs.stat`, ENOENT handling, permission warnings
- [x] 33 unit tests for `scaffold.ts` + `workspace-store.ts` (including 5 new) — credentials dir is `0o700`, other dirs track umask, workspace dir is `0o700`, `workspace.json` is `0o600`, updates preserve `0o600`
- [x] `bun run check` clean
- [x] `bun run lint` clean (pre-existing warning in `url-validator.ts` unrelated)
- [x] `bun run format:check` clean
- [x] `bun run test:unit` — 1744 pass, 0 fail, no regressions

### DIFF
--- a/src/config/workspace-credentials.ts
+++ b/src/config/workspace-credentials.ts
@@ -1,5 +1,6 @@
 import { chmod, mkdir, readFile, rename, stat, unlink, writeFile } from "node:fs/promises";
 import { join } from "node:path";
+import { WORKSPACE_ID_RE } from "../workspace/workspace-store.ts";
 
 /**
  * Workspace-scoped credential store.
@@ -10,13 +11,18 @@ import { join } from "node:path";
  * File format is plain JSON key-value — no metadata envelope:
  *   { "api_key": "sk-...", "workspace_id": "ws-..." }
  *
- * This is tier 1 of the 4-tier credential resolution hierarchy described in
- * `.tasks/credential-resolution/SPEC_REFERENCE.md`. This module implements the
- * file-level CRUD only. The tier resolver lives in `resolveUserConfig` (task 003).
+ * This is the tier-1 primitive for NimbleBrain's workspace-scoped credential
+ * resolution. This module implements the file-level CRUD only; the tier
+ * resolver layered on top is `resolveUserConfig` in the same module.
  *
- * Security: files are written with `0o600`, the `credentials/` directory is
- * created with `0o700`, and writes are atomic (temp file + rename). Credential
- * values are never logged; only keys appear in diagnostics.
+ * Security posture:
+ *   - Files are written with `0o600`, the `credentials/` directory is created
+ *     with `0o700`, and writes are atomic (temp file + rename).
+ *   - `wsId` is validated against `WORKSPACE_ID_RE` on every call because the
+ *     path derived from it is a filesystem path — a caller passing `../evil`
+ *     would otherwise escape the workspace tree. We don't trust the call site.
+ *   - Credential values are never logged; only keys and paths appear in
+ *     diagnostics.
  */
 
 // ── Path helpers ──────────────────────────────────────────────────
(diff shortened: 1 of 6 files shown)


mgoldsborough:
## QA Fixup Response

Pushed `e1f03e9` addressing warnings #1, #2 and suggestions S1, S2, S3, S6, S7.

### Warnings — fixed

**#1 — `wsId` / `bundleName` validation (defense in depth).** The credential store writes to filesystem paths derived from `wsId` and `bundleName`; a filesystem-writing module shouldn't rely on callers validating upstream.
- Exported `WORKSPACE_ID_RE` from `workspace-store.ts` and validate it in every public entry point (`save` / `get` / `clear` / `clearAll` / `credentialPath`).
- `bundleSlug` now rejects `..`, `.`, null bytes, backslashes, spaces, and any character outside `[A-Za-z0-9._-]` after the `@scope/` prefix is stripped.
- 7 new tests verify rejections across all entry points; 6 new tests cover `bundleSlug` rejections.

**#2 — Concurrent save race.** Implemented `withFileLock` — an in-process per-file promise chain. Each read-modify-write on a given path waits for the previous one to settle before starting. Lock is per-file so writes to different bundles in the same workspace proceed in parallel. `clearAllWorkspaceCredentials` also takes the lock so an unlink can't race with an in-flight rename.

Since NimbleBrain is single-process, in-process serialization is sufficient — no flock / O_EXCL needed.

4 new concurrent-write tests: paired saves, save + clear of different keys, 50-way fanout stress, and proof that writes to different bundles don't serialize.

### Suggestions — addressed

**S1 — TOCTOU on mode check** is advisory *by design*. A mismatched mode still returns credentials (we've already read the file; refusing wouldn't prevent our own access). The check exists to nudge operators toward fixing permissions before the file leaks via backup / sync / other readers. Expanded the docblock to make this explicit.

**S2 — mkdir leaf-only mode.** The credentials dir gets `0o700` via explicit chmod, but intermediate `workspaces/{wsId}/` relies on `WorkspaceStore.create` having run first. Documented the invariant in the `ensureCredentialsDir` docblock rather than duplicating the parent create — tests already verify the parent mode separately, so drift would be caught fast.

**S3 — silent chmod failure** now warns to stderr with a clear recovery hint. File contents stay protected via the explicit `0o600` on write; only the directory listing might be readable, and an operator will now see a warning.

**S6 — dead `.tasks/` doc reference.** Dropped. The security story in the module docblock reads self-contained.

**S7 — coverage gaps.** Added tests for:
- Malformed JSON throws with file path in message
- Non-object JSON (array) throws with "not a JSON object"
- File with `0o644` produces a single stderr warning, still returns creds, and the warning does NOT contain the credential value

### Skipped

**S4 — shared `fs-utils` for the tmp suffix counter.** Two callers isn't worth an abstraction layer. YAGNI.

**S5 — test file relocation.** Pure convention churn. Defer.

### About the earlier integration test failures

Diagnosed as missing `@modelcontextprotocol/sdk` in the stale worktree's `node_modules` — the worktree was created before a dep update. After `bun install` in the worktree, `bun run test:integration` passes 338/338. The code didn't regress; the environment was stale.

### Verify

- `bun run check` clean
- `bun run lint` clean (1 pre-existing `url-validator.ts` warning)
- `bun run format:check` clean
- `bun run test:unit` — 1787 pass, 0 fail (was 1766; added 21 tests)
- `bun run test:integration` — 338 pass, 6 skip (pre-existing), 0 fail

Issue #8, opened 2026-04-16
Default to restricted tool set for unnamed delegate sub-agents


mgoldsborough:
## Summary

When \`nb__delegate\` is called without explicit tool globs and without a named agent profile, the child engine receives the parent's full tool router. This is bounded by iteration limits and the delegate preamble, but a more restrictive default would reduce blast radius.

## Recommendation

Consider defaulting unnamed delegates to a restricted tool set that excludes administrative tools (\`nb__manage_app\`, \`nb__manage_users\`, \`nb__manage_workspaces\`). Named agent profiles with explicit tool lists are unaffected.

## Priority

P2 — defense-in-depth, bounded by existing iteration limits

Issue #7, opened 2026-04-16
Bundle startup failures not logged to workspace or surfaced to UI


mgoldsborough:
## Problem

When a bundle fails to start (e.g., wrong platform binary, missing deps, MCP connection refused), the error is written to `process.stderr` only (`src/runtime/workspace-runtime.ts:182`). It does not:

- Appear in workspace logs (`/data/workspaces/{ws}/logs/`)
- Emit an SSE event to connected clients
- Show up in `/v1/health` bundle status (the bundle simply doesn't exist in the registry)

Operators must check container logs to discover startup failures. Users see "App X is not available" with no explanation.

## Root cause

`startWorkspaceBundles()` catches errors from `startBundleSource()` and writes to stderr, but has no reference to the EventSink to emit a persistent event.

## Fix

1. Pass the EventSink into `startWorkspaceBundles()`
2. On catch, emit a new `bundle.start_failed` event with `{ name, error, wsId }`
3. Add `bundle.start_failed` to the `WORKSPACE_EVENTS` set in `src/adapters/workspace-log-sink.ts`
4. Add `bundle.start_failed` to the SSE forwarding list in `src/api/events.ts`
5. Register the failed bundle in HealthMonitor as `state: "dead"` so `/v1/health` reflects it

## Key files

- `src/runtime/workspace-runtime.ts` lines 173-185 — catch block
- `src/adapters/workspace-log-sink.ts` lines 6-21 — WORKSPACE_EVENTS set
- `src/api/events.ts` lines 93-111 — SSE forwarding filter
- `src/tools/health-monitor.ts` — HealthMonitor state tracking

## Acceptance criteria

- `bundle.start_failed` events appear in workspace JSONL logs
- `bundle.start_failed` events are forwarded via SSE
- `/v1/health` shows failed bundles as `dead`
- Existing startup-continues-on-failure behavior is preserved

Issue #6, opened 2026-04-16
Add custom header requirement for state-changing API requests


mgoldsborough:
## Summary

State-changing endpoints (`/v1/chat`, `/v1/tools/call`) rely on `SameSite=Lax` cookies and CORS for CSRF protection. There are no explicit CSRF tokens or custom header requirements. While JSON-body APIs behind CORS are generally safe from form-based CSRF, a custom header check provides defense-in-depth.

## Recommendation

Require a custom header (e.g., `X-Requested-With: NimbleBrain`) on all POST requests that use cookie authentication. This is a lightweight check that blocks CSRF from non-JS contexts (forms, image tags) without requiring token management.

## Priority

P2 — defense-in-depth, current protection via SameSite + CORS is adequate for JSON APIs

Issue #5, opened 2026-04-16
Rate limiter should support trusted proxy IP resolution


mgoldsborough:
## Summary

Login rate limiting currently uses a static key for all requests rather than per-client IP tracking. This is intentional to avoid trusting spoofable headers, but it means all clients share a single rate limit bucket.

## Recommendation

Add a `TRUSTED_PROXY_HEADER` configuration option. When set (e.g., to `X-Real-IP`), rate limiting keys on that header value. When unset, current behavior is preserved. Document that this should only be set when deployed behind a reverse proxy that sets the header from the socket address.

## Priority

P2 — defense-in-depth

PR #42, opened 2026-04-18
feat: wire credential resolution into production paths (startup, configure tool, CLI)

mgoldsborough:
## Summary

Wires `resolveUserConfig` (#41) into the three production paths that actually use it, plus the CLI migration. This is the user-visible payoff of the credential resolution work — after this merges, the original newsapi bug is fixed end-to-end.

Three commits, one per task in the implementation plan:

- **[005] `bundles`**: `startBundleSource` resolves credentials before `mpak.prepareServer` and passes them via the new `userConfig` option (mpak PR #81). Covers `@nimblebraininc/newsapi` and any other registry bundle with required `user_config`.
- **[006] `tools`**: `nb__manage_app configure` now prompts via TUI and persists to the workspace credential store — no longer writes `~/.mpak/config.json`.
- **[007] `cli`**: `nb config set/get/clear` require `-w <wsId>`. Workspace is validated before writing. `~/.mpak/config.json` is no longer touched.

> **Stacked on #41.** Base is `feat/resolve-user-config`. Merging #40 → #41 → this in order auto-retargets each to `main`.

## End-to-end fix

Before this PR (and the stack behind it), a workspace with:
```json
{ "bundles": [{ "name": "@nimblebraininc/newsapi" }] }
```
and no value in `~/.mpak/config.json` fails with:
```
[workspace-runtime] Failed to start newsapi in ws_mat: Missing required config for @nimblebraininc/newsapi: NewsAPI API Key
```

After this PR, the same workspace starts cleanly if credentials live in any of:
1. Workspace credential store (`nb config set @nimblebraininc/newsapi api_key=sk-xxx -w ws_mat`)
2. Process env (`NB_CONFIG_NIMBLEBRAIN_NEWSAPI_API_KEY=sk-xxx`)
3. The manifest's default value

## `wsId` is always required

Per SPEC_REFERENCE, NimbleBrain is workspace-scoped — no default workspace, no global store. This PR threads `wsId` through `startBundleSource`, `workspace-ops`, `workspace-runtime`, `lifecycle.installNamed`, `configureBundle`, and all three CLI commands. Callers that don't have a workspace fail fast with actionable errors.

## Test plan

- [x] 4 new integration tests for startup — every tier of the resolution hierarchy + missing-wsId precondition, exercised against a real mpak cache layout with a tiny echo-env MCP server (no mocking of `prepareServer`)
- [x] 5 new unit tests for `configureBundle` — non-interactive error format, prompted values persisted to workspace store, `forcePrompt` behavior
- [x] 12 new unit tests for CLI — `-w` required, workspace validation, masked display, per-key clear, no `.mpak/` ever created
- [x] `bun run check` clean
- [x] `bun run lint` clean (1 pre-existing warning in `url-validator.ts`, unrelated)
- [x] `bun run format:check` clean
- [x] `bun run test:unit` — 1811/1811 pass, no regressions
- [x] `bun run test:integration` — 342 pass, 6 skip (pre-existing), 0 fail

## Manual verification needed after merge

The mpak SDK change (PR #81) is locally linked in my dev environment. Reviewers testing locally should:
1. Build the SDK: `cd /path/to/mpak-pr-81/packages/sdk-typescript && pnpm build`
2. Link: from that dir, `bun link`; from nimblebrain, `bun link @nimblebrain/mpak-sdk`
3. Test: create a workspace with `@nimblebraininc/newsapi`, set credentials via `nb config set ... -w <wsId>`, start nimblebrain

Once mpak PR #81 merges and is published, the linking step goes away.


mgoldsborough:
## QA Fixup Response

Pushed `ab026d1` + cherry-picked `8e776d2` (task 009) on top of the existing stack. Summary of QA addressed, with first-principles justification:

### Critical — fixed

**#1 — Restart call omits `wsId`/`workDir`:** Fixed, but went deeper than the direct patch. The root cause was that `startBundleSource` accepted `wsId` and `dataDir` as independent opts for named bundles — but they're *redundant*: the data dir is always derivable from `wsId`+`workDir` via `resolveBundleDataDir`. Every caller computed `dataDir` separately, which is the shape of bug waiting to recur. Moved the derivation *inside* `startBundleSource` so there is now a single source of truth. `opts.dataDir` retained as a test-fixture override only. `configureBundle` stops computing data dirs entirely and just passes `wsId`+`workDir`.

**#2 — Test masks regression:** New integration test `test/integration/configure-credentials.test.ts` exercises the full `manage_app configure` flow with a real subprocess MCP server (reusing the fixture pattern from `startup-credentials.test.ts`). Asserts `isError === false` AND `registry.hasSource(...) === true` after restart — both invariants the QA-flagged bug would violate. Left the existing unit test as-is since it still validates credential persistence, which is separate behavior.

### Warnings — fixed

**#3 — Data dir drift:** Resolved by the same architectural change as #1. All four call sites (`lifecycle.installNamed`, `workspace-ops`, `workspace-runtime`, `configureBundle`) now produce identical workspace-scoped paths because `startBundleSource` owns the derivation.

**#4 — Dead module:** Deleted `src/config/credentials.ts` (114 lines). `git grep` confirmed zero live references across `src/` and `test/`. (This was also task 008's scope, which would have shipped in a follow-up PR — folded here since QA asked for it on this PR.)

**#5 — README drift:** Updated `README.md:286` to call out the `-w <wsId>` requirement on `nb config set|get|clear`.

**#6 — Error message leaks internal identifier:** Rephrased the guard in `startup.ts` from `startBundleSource: "wsId" is required...` to `Cannot start <name>: a workspace ID is required (platform bug — please report).` Reads as what it is — an internal invariant violation, not a user mistake. The test's regex updated to match.

### Suggestion — addressed

**#7 — `inc`-stripping collision:** Added a JSDoc comment on `normalizeScope` calling out that the rule applies to *all* scopes (so a hypothetical `@fooinc/...` would collide with `@foo/...`), with guidance on when to promote to an explicit alias map. No runtime behavior change.

### Deferred

**#8 — `installNamed` direct env read:** Pure code-quality. Three callers compute `NB_WORK_DIR` three different ways; harmless today. Cleanup-PR material.

### Bonus — task 009 folded in

Cherry-picked `e0fe1da` (task 009 — credential cleanup on bundle uninstall) since it was ready in a sibling branch and completes the lifecycle story: credentials are now created in the workspace store at configure time AND removed from it when the bundle is uninstalled (best-effort; uninstall never fails on cleanup errors).

### Verify

- `bun run check` clean
- `bun run lint` clean (1 pre-existing `url-validator.ts` warning)
- `bun run format:check` clean
- `bun run test:unit` — 1824 pass, 0 fail
- `bun run test:integration` — 343 pass, 6 skip (pre-existing), 0 fail

PR #43, opened 2026-04-20
refactor(config): drop NB_CONFIG_* convention, use SDK-declared env aliases

mgoldsborough:
## Summary

Removes NimbleBrain's host-specific \`NB_CONFIG_{SCOPE}_{NAME}_{FIELD}\` env-var convention. The mpak SDK (0.4.0, from mpak#82) now does the resolution itself by reading each bundle's canonical \`mcp_config.env\` declaration in reverse: a bundle that maps \`\"NEWSAPI_API_KEY\": \"\${user_config.api_key}\"\` is satisfied by a host \`NEWSAPI_API_KEY\` export. No host convention, no special-case \`inc\`-stripping, no duplicate source of truth.

## End-to-end behavior

**Before:**
\`\`\`bash
export NB_CONFIG_NIMBLEBRAIN_NEWSAPI_API_KEY=sk-xxx    # ugly, host-specific
bun run dev   # newsapi starts
\`\`\`

**After:**
\`\`\`bash
export NEWSAPI_API_KEY=sk-xxx                          # matches the bundle's declared mapping
bun run dev   # newsapi starts
\`\`\`

Same pattern works for any bundle that declares canonical env vars — \`ANTHROPIC_API_KEY\`, \`OPENAI_API_KEY\`, \`GITHUB_TOKEN\`, etc. No renaming required.

## Architectural shift

Host-side resolution shrinks to two concerns:
1. Read the workspace credential store.
2. In the TUI configure flow, prompt via gate + persist responses.

Everything else — \`mcp_config.env\` reverse lookup, manifest defaults, required-field validation — lives in the SDK's \`gatherUserConfig\`. The result is:

- **Every mpak host benefits uniformly.** The convention is in one place.
- **Bundle manifest is the single source of truth.** It already had to declare the field-to-env mapping for spawn time; the SDK just uses it both ways.
- **No special-case scope stripping.** The \`inc\`-stripping rule that was a warning sign of wrong design is gone.

## Changes

- Bump \`@nimblebrain/mpak-sdk\` 0.3.0 → 0.4.0
- Delete \`envVarName()\` and \`normalizeScope()\`
- Remove the process-env tier from \`resolveUserConfig\`
- \`resolveUserConfig\` returns a partial map (no throw on missing required) — the SDK owns that contract
- Add \`friendlyMpakConfigError()\` helper: translates SDK's \`MpakConfigError\` into the \`nb config set -w <wsId>\` hint
- \`startBundleSource\` catches \`MpakConfigError\` via the helper so user-facing messages stay the same
- Net diff: +332 / -687

## Test plan

- [x] \`workspace-credentials.test.ts\` rewritten for the simplified contract. 61 tests covering the primitives (slug, paths, CRUD, permissions, concurrent writes), the new \`resolveUserConfig\` shape, and \`friendlyMpakConfigError\` translation.
- [x] \`integration/startup-credentials.test.ts\` "env path" test now uses the bundle-declared \`NBTEST_API_KEY\` directly.
- [x] \`bun run test:unit\` — 1832 pass, 0 fail.
- [x] \`bun run test:integration\` — 351 pass, 0 fail.
- [x] \`bun run check\` clean.
- [x] \`bun run lint\` clean (1 pre-existing \`url-validator.ts\` warning).
- [x] \`bun run format:check\` clean.

## Manual verification

\`\`\`bash
# With newsapi bundle declaring:
#   \"NEWSAPI_API_KEY\": \"\${user_config.api_key}\"
export NEWSAPI_API_KEY=sk-xxx
bun run dev
# → newsapi starts, no NB_CONFIG_* needed, no nb config set required
\`\`\`

Closes the ergonomics gap raised in the post-merge critique of #42.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -10,7 +10,7 @@
         "@ai-sdk/openai": "^3.0.48",
         "@modelcontextprotocol/ext-apps": "^1.3.1",
         "@modelcontextprotocol/sdk": "^1.29.0",
-        "@nimblebrain/mpak-sdk": "0.3.0",
+        "@nimblebrain/mpak-sdk": "0.4.0",
         "@nimblebrain/synapse": "^0.4.0",
         "@workos-inc/node": "^8.12.1",
         "ai": "^6.0.138",
@@ -84,7 +84,7 @@
 
     "@nimblebrain/mpak-schemas": ["@nimblebrain/mpak-schemas@0.2.0", "", { "dependencies": { "zod": "^4.1.12" } }, "sha512-hGE6RV0krHrQr0wF2zi+0/DwzLwxDago4hyEZbUPbKlfmC2ZFT0hIrT60xcEVEAg3jRZewJX4FaCkvbIMVMStw=="],
 
-    "@nimblebrain/mpak-sdk": ["@nimblebrain/mpak-sdk@0.3.0", "", { "dependencies": { "@nimblebrain/mpak-schemas": "0.2.0", "zod": "^4.3.6" } }, "sha512-jAW46HaPEKtFtWT8ToB6NHMwWCfU9fF1NyPUDk2TwcFlw4a7onv8kAOexBMLXbpGpW2NASd2yktlx8ANA0EGdg=="],
+    "@nimblebrain/mpak-sdk": ["@nimblebrain/mpak-sdk@0.4.0", "", { "dependencies": { "@nimblebrain/mpak-schemas": "0.2.0", "zod": "^4.3.6" } }, "sha512-/DIOjaktfR5rUheCyLR188xcnC+krmz0lws4umvzWoT2fkHb4RTEeGtiqaQJD9gzRjtvxoDucWLJY1gbilnjMg=="],
 
     "@nimblebrain/synapse": ["@nimblebrain/synapse@0.4.0", "", { "peerDependencies": { "@modelcontextprotocol/ext-apps": "^1.3.1", "react": "^18.0.0 || ^19.0.0" }, "optionalPeers": ["react"], "bin": { "synapse": "dist/codegen/cli.js" } }, "sha512-7VLuwPhAgp7n7Ok9IVRs/qtK64KTl7RAPMWR2IGmiyjXW/552vwuMrAgDkksuPrIiZECalXSx08Zaks/+SRsxg=="],
 
--- a/bun.lock
+++ b/bun.lock
@@ -10,7 +10,7 @@
         "@ai-sdk/openai": "^3.0.48",
         "@modelcontextprotocol/ext-apps": "^1.3.1",
         "@modelcontextprotocol/sdk": "^1.29.0",
-        "@nimblebrain/mpak-sdk": "0.4.0",
+        "@nimblebrain/mpak-sdk": "0.4.1",
         "@nimblebrain/synapse": "^0.4.0",
         "@workos-inc/node": "^8.12.1",
         "ai": "^6.0.138",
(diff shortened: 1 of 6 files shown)


mgoldsborough:
## Follow-up on QA feedback

Three items evaluated:

**1. Drop \`allowedEnv: [\"ANTHROPIC_API_KEY\"]\` from workspace.json** — correct but user-side hygiene, not PR work. Credentials now flow via \`user_config\` resolution, not \`allowedEnv\` passthrough. After merging, users should remove redundant credential entries from their \`allowedEnv\`; keep \`allowedEnv\` for genuine non-credential vars like \`HTTP_PROXY\`.

**2. \`instanceof MpakConfigError\` vs duck-typing** — done (\`0274d6c\`). The SDK is already a direct dep so there's no import cost; instanceof gives proper TS narrowing, can't be typo'd, and fails at compile time if the SDK ever removes the export. Also caught a latent bug in the process: the old \`?? f.key\` title fallback didn't handle empty-string titles (\`??\` treats \`\"\"\` as present). Fixed.

**3. Smoke on main after merge** — process advice, will run naturally once merged.

Tests: 62 passing in workspace-credentials.test.ts (new regression test documents that look-alike objects no longer translate — intentional behavior change, that's the point of instanceof). Full \`bun run test:unit\` 1833 pass, \`test:integration\` 351 pass.

Ready to merge.

mgoldsborough:
## Enriching the onboarding error

Worth doing. Reasoning: this isn't a \"rarely-hit path\" — it's the **first error** any new user sees when they install a bundle without credentials. Making it actionable matters more here than almost anywhere else.

No extra fetch required — \`startBundleSource\` already reads the manifest from the cache before calling \`prepareServer\`. Threading it into \`friendlyMpakConfigError\` is free.

### Before

\`\`\`
Missing required config \"API Key\" for @scope/bundle.
Run one of:
  nb config set @scope/bundle api_key=<value> -w ws_mat
Or export the env var declared in the bundle's mcp_config.env.
\`\`\`

### After (with manifest)

\`\`\`
Missing required config \"API Key\" for @scope/bundle.
Run one of:
  nb config set @scope/bundle api_key=<value> -w ws_mat
  export ANTHROPIC_API_KEY=<value>  # satisfies \"api_key\"
  export CLAUDE_API_KEY=<value>  # satisfies \"api_key\"
\`\`\`

Implementation (\`cf6d890\`):

- \`friendlyMpakConfigError\` gains an optional third arg — the manifest (or a narrow subtype with just \`server.mcp_config.env\`).
- Only whole-value \`\${user_config.X}\` substitutions are surfaced, matching the SDK's resolver rule — we don't suggest names that wouldn't actually satisfy the field.
- Literal entries (\`\"LOG_LEVEL\": \"info\"\`) and composite templates (\`\"PREFIXED\": \"prefix-\${user_config.x}\"\`) are skipped.
- Manifest omitted/null/undefined → unchanged behavior (just the \`nb config set\` line).

Tests: 5 new cases pinning the shape — generic fallback, concrete alias naming, literal/partial-template skip, alias-less field, and null/undefined manifest equivalence. 1836 unit + 351 integration green.

mgoldsborough:
## Dropping duplicated env-alias derivation

With \`@nimblebrain/mpak-sdk@0.5.0\` (mpak#85), \`MpakConfigError\` self-describes — each missing field carries its \`envAliases: string[]\` list derived by the SDK. NimbleBrain's consumer no longer needs to re-derive from the manifest.

### What dropped

- \`envAliasesForField\` (duplicated regex + whole-value-substitution rule)
- \`EnvAliasSource\` interface
- \`manifest\` arg on \`friendlyMpakConfigError\`
- The \"keep in sync with SDK — see mpak/pull/83\" comment
- Manifest threading in \`startBundleSource\`'s catch block

Net: 54 insertions, 118 deletions. 54 of that is test shape updates for the \`envAliases: []\` required field on MpakConfigError.missingFields.

### Follow-up tracked separately

mpak-schemas drift from canonical MCPB v0.4 (optional \`title\`/\`description\`, missing \`directory\`/\`file\` types, etc.) is tracked in [mpak#84](https://github.com/NimbleBrainInc/mpak/issues/84). Not gating this PR.

### Verify

- \`bun run check\` / \`lint\` / \`format:check\` clean
- \`test:unit\` — 1835/1835
- \`test:integration\` — 351/351

Ready to merge.

PR #50, opened 2026-04-20
Add "analyzing" streaming state to cover post-tool inference gap

mgoldsborough:
## Summary

- Adds an `analyzing` streaming state to `useChat` that fills the previously-silent gap between the last `tool.done` and the next `text.delta` / `tool.start`
- Parallel tool batches now correctly stay in `working` until the final call lands — previously, any `tool.done` in a batch flipped the state to `streaming` even while siblings were still running
- `ToolAccordion` renders an "Analyzing…" sub-row (spinning Loader2 + animated ellipsis) when it's the active block and the model is inferring post-tool; composer footer mirrors this with an "Analyzing..." label

## Why

Real-world example from `conv_8f474335077b4ca4`: `webfetch__web_fetch` returned an error at `20:20:44.645`, `files__search` had already completed at `20:20:43.921`, and the next LLM response didn't start streaming until `~20:20:46.0` — a **~1,370ms** dead window during which the user saw a red "Couldn't fetch" headline and nothing else. That's well above both the `useMinDisplayTime` smoothing threshold and human perceptual threshold, so the UI genuinely looked frozen. Failure paths expose this worst because error recovery tends to require more thinking than happy-path synthesis.

## State machine

```
null → thinking → streaming ↔ working → analyzing → streaming → null
                                     ↘ working (if the model calls another tool)
```

- `tool.done` → `analyzing` only when `toolCallsRef.current.some(tc => tc.status === "running") === false`
- `text.delta` flips `analyzing → streaming` via the existing `prev !== "streaming"` guard
- `tool.start` flips `analyzing → working` via the existing unconditional set
- `done` / `error` still flush to `null`

## Test plan

- [x] `bun run verify:static` (format, lint, tsc, check:cycles)
- [x] `bun run test:unit` — 1835 pass
- [x] `bun run test:web` — 96 pass, including new coverage in `streamingState.test.tsx`:
  - `working → analyzing` on last `tool.done` (was: `working → streaming`)
  - `working` holds while parallel tools are still in flight; flips to `analyzing` on the final `tool.done`
  - `analyzing → streaming` on the next `text.delta`
  - `analyzing → working` when the model calls another tool
  - Updated full-cycle test to include the `analyzing` phase
- [ ] Manual: replay a prompt that causes a `web_fetch` redirect failure (e.g. "Search the web for the latest Anthropic model pricing…" hitting `anthropic.com/pricing` → `claude.com/pricing` redirect) and verify the "Analyzing…" sub-row appears under the red headline for the ~1.3s window before recovery text streams in
- [ ] Manual: send a multi-tool-call prompt and verify the indicator does NOT appear until *all* parallel tools complete

### DIFF
--- a/web/src/components/MessageInput.tsx
+++ b/web/src/components/MessageInput.tsx
@@ -147,7 +147,8 @@ export function MessageInput({
     [addFiles],
   );
 
-  const isActive = streamingState === "thinking" || streamingState === "working";
+  const isActive =
+    streamingState === "thinking" || streamingState === "working" || streamingState === "analyzing";
   const canSend = (text.trim() || attachedFiles.length > 0) && !disabled;
 
   return (
@@ -233,7 +234,11 @@ export function MessageInput({
       <div className="flex items-center justify-center gap-3 mt-2 text-[10px] text-muted-foreground">
         {isActive && (
           <span className="text-processing font-mono font-medium">
-            {streamingState === "working" ? "Working..." : "Thinking..."}
+            {streamingState === "working"
+              ? "Working..."
+              : streamingState === "analyzing"
+                ? "Analyzing..."
+                : "Thinking..."}
           </span>
         )}
         {onNewConversation && (
--- a/web/src/components/MessageList.tsx
+++ b/web/src/components/MessageList.tsx
@@ -341,6 +341,11 @@ export function MessageList({
                               <ToolAccordion
                                 calls={block.toolCalls}
                                 displayDetail={displayDetail}
+                                pending={
+                                  streamingState === "analyzing" &&
+                                  idx === messages.length - 1 &&
+                                  blockIdx === msg.blocks!.length - 1
+                                }
                               />
                               {blockWidgets.map((tc) => {
                                 const match = tc.resourceUri!.match(/^ui:\/\/[^/]+\/(.+)$/);
--- a/web/src/components/ToolAccordion.tsx
+++ b/web/src/components/ToolAccordion.tsx
@@ -28,11 +28,19 @@ import { describeBatch } from "../lib/tool-display";
 interface ToolAccordionProps {
   calls: ToolCallDisplay[];
   displayDetail: DisplayDetail;
+  /**
+   * True when this block's tools have all finished and the model is now
+   * inferring on the results (streamingState === "analyzing"). Renders an
+   * "Analyzing…" footer so the UI doesn't appear frozen between tool.done
+   * and the next text.delta / tool.start.
+   */
+  pending?: boolean;
 }
 
 export const ToolAccordion = memo(function ToolAccordion({
   calls,
   displayDetail,
+  pending = false,
 }: ToolAccordionProps) {
   // Visual statuses smooth the running→done transition so quick tools don't flash.
   const visualStatuses = useMinDisplayTime(calls);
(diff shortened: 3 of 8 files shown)


mgoldsborough:
## QA feedback — adjudication and follow-up commit

Thanks for the review. Walked through each item:

| # | Item | Verdict | Action |
|---|---|---|---|
| 1 | Dead `run_in_progress` branch | **Fix** | You're right — pre-existing uncommitted work that got captured when I staged `useChat.ts`. References an error code that isn't emitted anywhere in the codebase, and is unrelated to this PR's scope. Reverted in 3e8093e (both the branch and the `ApiClientError` import). |
| 2 | No UI tests for the new visual surfaces | **Fix (partial)** | Added two new test files in 3e8093e: <br>• `web/test/ToolAccordion.test.tsx` — 3 tests: pending footer renders only when `pending={true}`, absent when `false`, fully suppressed in quiet mode.<br>• `web/test/MessageList.pending.test.tsx` — 3 tests covering exactly the indexing regression you called out: the indicator must appear only on the last tool block of the latest message. Tests include the multi-turn case (earlier assistant's tool block must NOT get the indicator even though it's the last block of that turn). <br>Skipped: MessageInput label test — the logic is a trivially-correct 3-branch ternary; the state-machine test already guards the state value. |
| 3 | Double "Analyzing…" indicator (accordion + composer) | **No change** | Intentional. Accordion sub-row = "what the model is doing right now, below the tool that just landed." Composer footer = "what this turn is doing overall." Same rationale as the existing "Working..." / "Thinking..." labels in the composer coexisting with in-message spinners. I'll confirm it doesn't feel noisy in manual QA before merge. |
| 4 | `quiet` hides pending footer — worth a comment | **Fix** | Added a comment at `ToolAccordion.tsx:55` in 3e8093e explaining the intent: in quiet mode the composer's "Analyzing..." is the authoritative signal, which matches how quiet mode suppresses all tool surfaces. |
| 5 | Tab/space drift in test file | **Skip** | Biome passes format:check; `streamingState.test.tsx` is tab-indented consistently within itself. Not worth touching. |

### New test counts
- web: 96 → 102 pass (added 6)
- `bun run verify:static`: ✓
- `bun run test:unit`: 1835 pass (the fs.watch flake you saw is unrelated)

Ready for re-review.

mgoldsborough:
## QA feedback round 2 — adjudication + follow-up

Thanks for the thorough review. Walked through each item:

| # | Item | Verdict | Action |
|---|---|---|---|
| W1 | Sub-600ms tools: pending footer overlaps `useMinDisplayTime` smoother | **Fix** | Real visual regression — head shows spinner + "Researching · X", footer shows spinner + "Analyzing" at the same time. Gated the footer on `visualStatuses.every(vs => vs.status !== "running")` at `ToolAccordion.tsx:97`. Added a regression test in 365e218. |
| S1 | State diagram missing `analyzing → working` edge | **Fix** | Diagram now shows the edge (`useChat.ts:34-42`). The edge was already tested; doc was incomplete. |
| S2 | `role="status"` on the pending row | **Fix** | Added alongside `aria-live="polite"` — idiomatic for a live status indicator. |
| S3 | Extra `flushToMessage` on `tool.done` | **Skip — false alarm** | The diff only *moved* the `setStreamingState` call (before vs after the updater loop). Count of `setState` + `flushToMessage` invocations per `tool.done` is unchanged, and React 18 batches both setState calls into one render. No extra work. |

### Test counts after round 2
- web: 102 → 103 pass (added 1 new regression test for the visualStatuses gate)
- `bun run verify:static`: ✓
- `bun run test:unit`: 1835 pass

Ready for re-review.

PR #68, opened 2026-04-23
Add OAuth client for outbound remote MCP sources

mgoldsborough:
## Summary

- Closes the gap where workspace bundles with `{ url }` refs against OAuth-gated remote MCP servers returned `401 invalid_token` and failed to start
- Adds `WorkspaceOAuthProvider` (file-backed per `(workspace, serverName)`), `/v1/mcp-auth/callback` route, and a one-shot retry in `McpSource.start()` — the MCP TS SDK's `OAuthClientProvider` path is now wired end-to-end
- Handles headless flows (Reboot `Anonymous` dev provider, client-credentials-style auth) by probing the authorize chain up to 10 hops and resolving in-process when it lands at our callback. Interactive browser OAuth (Granola, Claude.ai hosted) fails fast with a clear `InteractiveOAuthNotSupportedError` — staged for a follow-up iteration so the extension point is visible, not hidden behind TODOs

See the commit message for the full per-file breakdown.

## Test plan

- [ ] `bun run verify` passes on this branch alone (verified locally; 1881 unit + 123 web + 370 integration + 16 smoke, all green)
- [ ] New unit tests cover: DCR roundtrip, tokens roundtrip, multi-hop headless detection, interactive-reject, missing-verifier, scope invalidation (10 tests for the provider + 4 for the flow registry)
- [ ] Existing `remote-integration.test.ts` and `remote-lifecycle.test.ts` still pass — covers bundles with static `transport.auth` (which now bypasses the provider) and url-only bundles
- [ ] Manual: added a Reboot-framework app to a workspace as `{ url: "http://localhost:9991/mcp" }` with no static auth, restart NB — bundle registers, tokens persist under `workspaces/<ws>/credentials/mcp-oauth/reboot-hello/`, tools callable from the agent

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -108,6 +108,8 @@ web/               Vite + React + TypeScript SPA (separate package.json)
 
 All tool handlers that access data must be workspace-scoped. Use `runtime.requireWorkspaceId()` (never `getCurrentWorkspaceId()`). In dev mode it returns `"_dev"` — no special-case logic needed.
 
+When adding a new code path that touches workspace-scoped credentials or identity, match the existing precedent: **hard-error on missing `wsId`, don't silently default**. `startBundleSource`'s named-bundle branch throws; the URL-bundle branch does too (for OAuth-provider paths). A `?? "ws_default"` fallback would pool credentials across tenants.
+
 ## Debug Logging
 
 Hot-path diagnostics are gated behind namespace flags so they're available when you need them without editing source. Use for tracing across the runtime ↔ SSE ↔ browser ↔ iframe chain.
--- a/src/api/app.ts
+++ b/src/api/app.ts
@@ -9,6 +9,7 @@ import { conversationEventRoutes } from "./routes/conversation-events.ts";
 import { eventRoutes } from "./routes/events.ts";
 import { healthRoutes } from "./routes/health.ts";
 import { mcpRoutes } from "./routes/mcp.ts";
+import { mcpAuthRoutes } from "./routes/mcp-auth.ts";
 import { resourceRoutes } from "./routes/resources.ts";
 import { toolRoutes } from "./routes/tools.ts";
 import { wellKnownRoutes } from "./routes/well-known.ts";
@@ -29,6 +30,11 @@ export function createApp(
   app.route("/", wellKnownRoutes(ctx));
   app.route("/", healthRoutes(ctx));
   app.route("/", authRoutes(ctx));
+  // Outbound-OAuth callback for remote MCP servers. Unauthenticated by
+  // design — state param guards against unsolicited codes. Must be
+  // reachable before any authenticated middleware; ordering alongside
+  // authRoutes keeps that invariant obvious.
+  app.route("/", mcpAuthRoutes(ctx));
 
   // MCP routes BEFORE other authenticated routes — prevents other sub-app
   // wildcard middleware from intercepting /mcp requests. Hono runs use("*")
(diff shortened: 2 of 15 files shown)


mgoldsborough:
Thanks for the thorough review. Fix-up landed in db8dfbf. PR #69 rebased onto the updated base.

## Adjudication

**Critical #1 (wsId drop)** — Confirmed and fixed. `installRemote` now threads `wsId` + `workDir` into `startBundleSource`; the URL-branch hard-errors on missing `wsId` when no static auth is set (matching the named-bundle precedent). No more `ws_default` silent fallback. Saved a memory note on "don't `?? 'ws_default'` workspace ids on credential boundaries" so future sessions don't step on the same mine.

**Critical #2 (SSRF)** — Confirmed and fixed. `redirectToAuthorization` now validates every hop through `validateBundleUrl` — initial URL, every redirect Location. Blocks surface as explicit `[workspace-oauth-provider] SSRF block …` errors so operators see the real cause, not the generic "interactive not supported." `allowInsecureRemotes` propagated into `WorkspaceOAuthProviderOptions` so local-dev (Reboot on localhost:9991) still works when the flag is explicitly on.

**Critical #3 (callback match)** — Confirmed and fixed. Precomputed canonical form at constructor time (lowercased origin, pathname stripped of trailing `/`). Regression test in integration covers trailing-slash + uppercase-host round-trip.

**Bonus bug found writing W5's test:** my original retry-path was broken — `StreamableHTTPClientTransport` rejects a second `start()` on the same instance, so the retry would have failed whenever it actually fired. Reboot happened to work because tokens landed on disk via a prior partial run, so the first connect succeeded with cached tokens. Fix mirrors the SDK's own `simpleOAuthClient` pattern: retry tears down the first transport+client, rebuilds via `createRemoteTransport`, reconnects on the fresh instance. Integration test at `test/integration/mcp-source-oauth-retry.test.ts` exercises the full 401→OAuth→200 flow.

**Warnings 4, 5, 6, 7, 9** — all applied:
- W4: Bun.serve cases moved from unit → integration per AGENTS.md classification
- W5: new `mcp-source-oauth-retry.test.ts` — full mock of discovery + DCR + authorize + token + MCP endpoints, drives the SDK through the retry end-to-end
- W6: connectWithTimeout timer captured + cleared in finally (was leaking 15–30s timer per start, doubled under retry)
- W7: `rename` moved to top static import
- W9: `Cache-Control: no-store` + `Pragma: no-cache` on callback responses

**W8 (token_endpoint_auth_method)** — acknowledged, deferred with TODO. "One-line check after DCR" understates it — switching between `"none"` and `"client_secret_basic"` post-registration requires implementing `OAuthClientProvider.addClientAuthentication` to inject the credential on token requests. Ship "none" unconditionally for now (works for Reboot Anonymous and AS's that issue refresh tokens to public clients, which the ext-apps examples do); follow-up ticket for confidential-client switching.

**Suggestions 10, 12, 13, 14** — all applied:
- S10: 15-min TTL per registration in flow-registry with `clearTimeout` on resolve/reject/_clearAll (prevents intra-process leaks); `timeout.unref?.()` so stuck flows don't keep CLIs alive
- S12: dropped redundant `this.dead = false` in retry path (matches happy path)
- S13: narrower `Transport & { finishAuth?: (...) }` cast — SSE has finishAuth too, old cast was a narrowing lie
- S14: startup warning when `NB_API_URL` is unset and a URL-ref bundle is being wired

**S11 (Hono html template)** — deliberately skipped. Hand-rolled escape is minimal and auditable; the swap adds Hono API surface coupling for cosmetic gain.

## Verify

1874 unit + 117 web + 378 integration + 16 smoke. Clean across the stack.

## Process note

Your review caught two landable bugs my PR didn't — cross-tenant leak (`ws_default` fallback) and the SSRF. Both were specifically introduced by adding the new OAuth plumbing; pre-existing code didn't have these vectors. The lesson I'm writing down for future me: when adding a new code path behind a credential / identity / isolation boundary, the "sensible default" instinct for missing inputs is an anti-pattern. Hard error, not fallback.

PR #69, opened 2026-04-23
Thread _meta.ui through resource reads and fix latent bridge bugs

mgoldsborough:
## Summary

Stacked on top of #68 (OAuth client for outbound remote MCP sources). Review #68 first; this PR's base will auto-retarget to `main` once that merges.

- Surfaces server-declared ext-apps UI metadata (CSP domain allowlists, permissions, layout hints) to the iframe bridge and honors it end to end
- `handleResourceProxy` now returns an MCP `ReadResourceResult`-shaped JSON envelope (`{ contents: [{ uri, mimeType?, text?, blob?, _meta? }] }`) instead of raw HTML — same endpoint, richer contract. `handleReadResource` picks up the same `_meta` pass-through so both endpoints are consistent.
- `buildCSP` and `createAppIframe` take the full `_meta.ui.csp.*` surface (`connect`/`resource`/`frame`/`baseUri` domains), map `_meta.ui.permissions` to the iframe `allow` attribute, and honor `_meta.ui.prefersBorder`. `_meta.ui.domain` is plumbed but deferred to the sandbox-proxy follow-up (see `nimblebrain-ops/research/SANDBOX_PROXY_ARCHITECTURE.md` for that work).
- Also fixes three latent bridge bugs that only surfaced against a strict spec-conformant client (Reboot's React runtime): `MessageList` regex dropped load-bearing URI segments, `ui/initialize` rejected numeric JSON-RPC ids, and the host sent out-of-spec theme tokens that strict Zod-validating clients reject.

See the commit message for the full per-file breakdown.

## Test plan

- [ ] `bun run verify` passes on this branch (verified locally; 1881 unit + 123 web + 370 integration + 16 smoke, all green — one flaky integration test `body-limit-routes` that passes in isolation; unrelated to this change)
- [ ] Integration test `core-registration.test.ts` updated to assert the new JSON envelope
- [ ] New unit tests: 3 for `McpSource.readResource` `_meta` passthrough, 6 for `buildCSP` with each new options field, 2 for bridge (numeric id + theme filter)
- [ ] Manual: Reboot clicker UI renders end-to-end in NimbleBrain — iframe resolves, CSP allows WebSocket to `:9991`, `+`/`−` buttons update the counter value live via `useCounter().useGet()` subscription

## Contract change note

`GET /v1/apps/:name/resources/:path` previously returned `text/html` with a raw HTML body. It now returns `application/json` with an MCP `ReadResourceResult` envelope. One integration test (`core-registration.test.ts`) was updated. No external consumers of this endpoint exist in the codebase — only NimbleBrain's web client calls it.

## Follow-up

- Interactive OAuth UI for Granola-class remote servers: `nimblebrain-ops/research/INTERACTIVE_OAUTH_UI.md`
- Per-member vs per-workspace credential scoping: `nimblebrain-ops/research/CREDENTIAL_SCOPING_DESIGN.md`
- Sandbox-proxy double-iframe pattern (consumes this PR's `_meta.ui.domain` plumbing unchanged): `nimblebrain-ops/research/SANDBOX_PROXY_ARCHITECTURE.md`


mgoldsborough:
Thanks for the thorough review. Fix-up landed in cd6838d.

## Adjudication summary

**Critical #1 (PDF download)** — Dead code, not a regression. A trace-through showed `triggerResourceDownload` only fires for tools ending in `__export_pdf`, the only such tool (synapse-collateral's `export_pdf`) registers at `ui://collateral/preview.pdf`, and the hardcoded URL this function builds resolves to `ui://synapse-collateral/preview.pdf` — never matched, 404'd before and after this PR. User-visible PDF downloads flow through `ResourceLinkView` via the POST endpoint, which this PR didn't touch. **Action taken:** removed the dead function + single call site rather than "fixing" it. CHANGELOG notes the removal.

**Criticals #2, #3, #4** — all confirmed blockers, all fixed:
- #2 CSP injection: `isValidCspDomain()` added; every declared domain validated against a scheme-URL regex; CSP keywords / metacharacters rejected with a `console.warn` (loud rather than silent). Bundle declaring `["https://x; script-src *"]` now sees its injection entry dropped and a warning emitted.
- #3 Permissions: default-OFF via new `CreateIframeOptions.honorServerPermissions` (defaults `false`). Server-declared camera/mic/geolocation ignored unless the host opts in. `clipboard-write` stays always-on (gesture-gated, not device access).
- #4 Theme filter: moved inside `bridge.setHostContext` so the SlotRenderer theme-toggle path (and any future caller) can't bypass. Single enforcement point.

**Warnings 1–6** — all applied. W7 (MessageList render test) deferred: `web/src/components/` has no React render-test infrastructure (no happy-dom test config for components); the behavior is trivially verifiable from the diff (regex removed, URI passed through unchanged), and the extracted `uiPathFromUri` helper consolidates the strip logic as a stronger guard. Flagged in the commit as explicit deferral.

**Suggestions 3, 4** — both applied (`uiPathFromUri` + `buildResourceEnvelopeEntry` extracted). Suggestions 1, 2, 5 deferred as noted in the commit.

## Full verify

1887 unit + 130 web + 370 integration + 16 smoke — all green.

## Process note for me

One process lesson I owe the thread: my PR description's "no external consumers" claim came from a `getResources()` caller grep, not a full audit of every `<a href>` / `<iframe src>` / direct fetch targeting the endpoint URL pattern. That's how I missed `triggerResourceDownload` in my own review. Your catch forced me to trace the actual user-visible flow rather than accept the claimed regression at face value — which is how the dead-code framing emerged. Noted for next time.

PR #85, opened 2026-04-25
Fix unclickable code-block copy/download buttons in chat (#81)

### LINKED ISSUE #81: Code block copy and save buttons in chat do not work
## Summary

The copy and save/download buttons rendered on formatted code blocks in the chat UI do not function. Clicking them has no visible effect — no clipboard write, no file download, no feedback state.

## Repro

1. Start a conversation and ask the agent to produce a code block (e.g., "write a Python hello world").
2. Hover over the rendered code block — copy and save buttons appear.
3. Click the copy button.
4. Observe: clipboard is not updated (paste confirms nothing was written).
5. Click the save/download button.
6. Observe: no file download is triggered.

## Root cause

Code blocks are rendered by the `<Streamdown>` component (`web/src/components/MessageList.tsx:317`), which provides its own built-in copy and download buttons on code fences. These buttons are part of the library's default rendering and are not wired to any custom handlers in the NimbleBrain integration.

The custom `CopyButton` component in `MessageList.tsx:32` operates at the message level (copies the full raw markdown text of the message), not at the individual code block level. It is a separate button from the ones Streamdown renders inside each code block.

The Streamdown code block buttons likely rely on browser clipboard and download APIs that either:
- Require a user gesture in a Secure Context, which may not be satisfied in the current render path, or
- Are broken by the way the Streamdown component is mounted (e.g., inside an iframe, shadow DOM, or focus-trapped container that intercepts events)

No error handling exists anywhere in the copy flow — `navigator.clipboard.writeText()` calls are fire-and-forget with no try/catch (`MessageList.tsx:36`, `ToolAccordion.tsx:218`). Silent failures leave users with no feedback.

## What would help

- Determine whether Streamdown's built-in buttons work in isolation (outside NimbleBrain's layout)
- If the issue is library-level: replace Streamdown's default code block rendering with a custom component that calls `navigator.clipboard.writeText()` directly with proper error handling and user feedback
- If the issue is context-level (iframe / CSP / Secure Context): ensure the chat panel is served in a context where clipboard and download APIs are permitted
- Add try/catch around all `navigator.clipboard.writeText()` calls and display a visible error state when the write fails

## Key files

- `web/src/components/MessageList.tsx:317–327` — Streamdown instantiation (code block rendering)
- `web/src/components/MessageList.tsx:32–51` — message-level CopyButton (separate from code block buttons)
- `web/src/components/ToolAccordion.tsx:213–242` — tool result CopyButton (same pattern, no error handling)
- `web/src/index.css:298–300` — Streamdown code block styling

mgoldsborough:
Fixes #81.

## Summary

- Override Streamdown's `position: sticky` + `pointer-events-none` wrapper on code-block action buttons with `position: absolute`, so clicks reach the buttons in nested scroll containers. This is the documented workaround for upstream [vercel/streamdown#494](https://github.com/vercel/streamdown/issues/494) — Streamdown's own `navigator.clipboard.writeText()` call is fine; hit-testing on the sticky wrapper was eating every click.
- Add real error handling to the message-level `CopyButton` in `MessageList.tsx` and the tool-result `CopyButton` in `ToolAccordion.tsx`. Both were calling `navigator.clipboard.writeText()` fire-and-forget, so any failure was invisible to the user. They now await the write and surface an AlertCircle "failed" state on error.

## Root cause

The issue's Secure-Context / clipboard-API theories are red herrings — I reproduced Streamdown's buttons in isolation on `localhost` and they work correctly, and `localhost` is a Secure Context. The actual failure is CSS layering: Streamdown renders code-block actions as

```html
<div class="pointer-events-none sticky top-2 -mt-10">
  <div class="pointer-events-auto">[buttons]</div>
</div>
```

In a layout with 2+ nested scroll containers (the chat panel has one), hit-testing on the inner `pointer-events-auto` div fails and clicks register on the outer `<div data-streamdown="code-block">`. DevTools "Select element" on the button area picks the wrapper, not the button, confirming the interception.

## Trade-off

With `position: absolute`, the action buttons no longer follow the viewport when you scroll past long code blocks — they stay pinned to the top of the block. For chat prose that's a fine trade, and upstream considers this the standard workaround.

## Test plan

- [ ] Start a conversation, ask for a long code block, hover over it, click Copy — paste confirms the code was copied.
- [ ] Same flow for Download — file downloads with correct extension.
- [ ] Hover over a message, click the message-level copy button — paste confirms full-message copy, and the icon briefly shows Check.
- [ ] Temporarily block clipboard (e.g., DevTools `navigator.clipboard = undefined`), click either custom copy button — icon shows AlertCircle "failed" state instead of silently doing nothing.
- [ ] `bun run verify` passes.

### DIFF
--- a/web/bun.lock
+++ b/web/bun.lock
@@ -16,7 +16,7 @@
         "react-dom": "^19.0.0",
         "react-router-dom": "^7.1.0",
         "shadcn": "^4.1.0",
-        "streamdown": "^2.5.0",
+        "streamdown": "~2.5.0",
         "tailwind-merge": "^3.5.0",
         "tw-animate-css": "^1.4.0",
       },
--- a/web/package.json
+++ b/web/package.json
@@ -22,7 +22,7 @@
     "react-dom": "^19.0.0",
     "react-router-dom": "^7.1.0",
     "shadcn": "^4.1.0",
-    "streamdown": "^2.5.0",
+    "streamdown": "~2.5.0",
     "tailwind-merge": "^3.5.0",
     "tw-animate-css": "^1.4.0"
   },
--- a/web/src/components/ToolAccordion.tsx
+++ b/web/src/components/ToolAccordion.tsx
@@ -235,7 +235,7 @@ function CopyButton({ content }: { content: string }) {
       type="button"
       onClick={onClick}
       className="tool-accordion__copy"
-      aria-label="Copy to clipboard"
+      aria-label={state === "failed" ? "Copy failed" : "Copy to clipboard"}
     >
       {state === "copied" ? (
         <>
--- a/web/test/MessageList.copy.test.tsx
+++ b/web/test/MessageList.copy.test.tsx
@@ -1,0 +1,140 @@
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import { act, fireEvent, render, waitFor } from "@testing-library/react";
+import type { ChatMessage } from "../src/hooks/useChat.ts";
+import { MessageList } from "../src/components/MessageList.tsx";
+
+/**
+ * Locks in the message-level copy button's success and failure feedback.
+ * Issue #81 was about Streamdown's per-block buttons (fixed via CSS), but
+ * the same PR added try/catch + AlertCircle feedback to the custom copy
+ * button — easy to silently regress if someone ever drops the await or
+ * the failure-state branch. This test fails noisily if either happens.
+ */
+
+const assistantMsg: ChatMessage = {
+	role: "assistant",
+	content: "hello world",
+};
+
+let originalClipboard: PropertyDescriptor | undefined;
+
+beforeEach(() => {
+	originalClipboard = Object.getOwnPropertyDescriptor(globalThis.navigator, "clipboard");
+});
+
+afterEach(() => {
+	if (originalClipboard) {
+		Object.defineProperty(globalThis.navigator, "clipboard", originalClipboard);
+	} else {
+		// happy-dom's navigator may not have clipboard by default; remove what we set
+		// biome-ignore lint/performance/noDelete: cleanup of test mock
+		delete (globalThis.navigator as { clipboard?: unknown }).clipboard;
+	}
+});
+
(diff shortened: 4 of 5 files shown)


mgoldsborough:
@Ovaculos can you take a look at this? i can't recreate it locally, but claude said it was able to determine the issue and identify the fix. need some help here.

mgoldsborough:
QA review adjudicated — three of the seven items addressed in 0d8c8c9; the rest declined with reasoning below.

## Accepted

**W1 — Pin streamdown.** Pinned to `~2.5.0` (patch only). The CSS workaround targets Streamdown's internal `data-streamdown` attribute, so a minor bump could silently re-break #81. The RTL alternative the reviewer offered is weaker — happy-dom doesn't fully implement CSS layout, so `getComputedStyle().position` checks are unreliable. Pin is decisive and cheap; revert to caret once vercel/streamdown#494 lands a real fix upstream.

**W2 — ToolAccordion a11y parity.** Fixed. `aria-label` now flips to "Copy failed" on the error branch, matching the message-level button.

**S1 — Tests.** Added `web/test/MessageList.copy.test.tsx` and `web/test/ToolAccordion.copy.test.tsx`. Each covers success, `writeText` rejection, and missing Clipboard API, asserting the visible failure affordance. 136 web tests pass.

## Declined

**S2 — Other clipboard sites.** Filed as #88. Out of scope for #81 (which is specifically about chat code blocks); the three sites in `ChatPanel.tsx` / `WorkspaceDetailPage.tsx` / `ProfileTab.tsx` deserve their own pass.

**S3 — `useClipboard` hook.** Two callsites isn't enough to justify the hook; the duplicated state machine is ~10 lines. If a third site appears while doing #88, that's the natural moment to extract — noted in the issue.

**S4 — Throw-to-catch style.** Stylistic preference. The current pattern unifies "API not available" and "writeText threw" into one catch; the suggested alternative duplicates the state-and-timer logic. Not changing.

**S5 — Rapid-click race.** Pre-existing across all clipboard sites (not introduced by this PR). Best fixed in the same place as #88's hook extraction.

Ovaculos:
This works for me

PR #91, opened 2026-04-25
Unify platform sources on MCP via in-memory transport (#90)

### LINKED ISSUE #90: InlineSource.readResource has incompatible signature with ResourceReader interface
## Problem

The `ResourceReader` interface and `InlineSource.readResource` have incompatible signatures, but the type guard `isResourceReader` only checks for method existence. Sources that implement `readResource` with the wrong shape pass the guard and silently misbehave at runtime.

## Specifics

`src/tools/types.ts:71`:
```ts
export interface ResourceReader {
  readResource(uri: string): Promise<ResourceData | null>;  // async, structured return
}

export function isResourceReader(source: unknown): source is ToolSource & ResourceReader {
  return source != null && typeof (source as Record<string, unknown>).readResource === "function";
  // ↑ only checks the method exists; not the signature or return type
}
```

`src/tools/inline-source.ts:85`:
```ts
readResource(resourcePath: string): string | null   // sync, raw string return
```

`McpSource.readResource` returns the proper async `ResourceData` shape. Both are valid implementations under the type guard.

## Impact

`nb__read_resource` (added in #25) iterates sources via `isResourceReader` and processes results as `ResourceData`:

```ts
const data = await source.readResource(uri);
if (data == null) continue;
if (typeof data.text === "string") { /* return text */ }
if (data.blob) { /* return binary placeholder */ }
// otherwise falls through to "not found"
```

When called against an `InlineSource`, `data` is a plain string. `data.text` is `undefined`, `data.blob` is `undefined`, and the source is silently skipped. **`nb__read_resource` cannot resolve any URI served by an `InlineSource`** — including the `ui://` resources several existing platform sources publish (settings, home, automations, conversations, files, usage).

This isn't surfaced by tests in #25 because the test harness uses a mock source that returns the structured `ResourceData` shape, mimicking `McpSource`.

## Why this matters now

The bundle-instructions Phase 1 spec ([nimblebrain-ops/research/BUNDLE_INSTRUCTIONS_PHASE1.md](https://github.com/NimbleBrainInc/nimblebrain-ops/blob/main/research/BUNDLE_INSTRUCTIONS_PHASE1.md)) needs the platform to publish resources at `instructions://*` and `prompt://composed/*`, reachable via `nb__read_resource`. The natural pattern would be to add resources to the existing settings `InlineSource` — but that path is silently broken. Phase 1 will add a new `ResourceReader`-conforming source class as a workaround, but the underlying inconsistency remains.

## Possible fixes

1. **Fix the type guard.** Make `isResourceReader` validate signature shape (e.g., probe with a sentinel call, or require an explicit marker like `static readonly __isResourceReader = true`). Tightens correctness; doesn't fix existing InlineSources.

2. **Bring `InlineSource.readResource` to the `ResourceReader` shape.** Change return type to `Promise<ResourceData | null>` and wrap the existing string returns as `{ text: html }`. Update the one caller in `handlers.ts:330` that consumes the string return. ~10 LOC of changes; unifies the interface.

3. **Split into two interfaces.** `IframeResourceReader` (sync, string for HTML iframes) vs `MCPResourceReader` (async, `ResourceData`). Type guard for each. Existing patterns for each survive; explicit at every call site.

Lean: **option 2.** The two patterns aren't conceptually different; the divergence is historical. Aligning on the MCP-shaped return makes `InlineSource` work uniformly with `nb__read_resource`, the `/mcp` resources/read endpoint, and any future `ResourceReader`-aware tool. The HTML-iframe path adapts trivially (`source.readResource(path).then(d => d?.text)`).

## Non-goals

- Bigger refactors of the resource model.
- Changing the public MCP resource shape.

## References

- #3 — original `nb__read_resource` issue
- #25 — merged PR adding `nb__read_resource`
- [BUNDLE_INSTRUCTIONS_PHASE1.md](https://github.com/NimbleBrainInc/nimblebrain-ops/blob/main/research/BUNDLE_INSTRUCTIONS_PHASE1.md) — first downstream consumer to surface the issue

mgoldsborough:
## Summary

- Every tool/resource provider — platform built-ins and user-installed bundles alike — is now an MCP server. Built-ins run in-process over `InMemoryTransport`; bundles continue as subprocess or remote MCP. One contract, one shape, one set of capabilities.
- Closes [#90](https://github.com/NimbleBrainInc/nimblebrain/issues/90) by construction: with one source class and one resource shape, the structural type-guard divergence the issue described is no longer expressible.
- Future MCP capabilities (prompts, sampling, elicitation) work for in-process built-ins automatically the day the SDK supports them — no parallel implementation to keep in sync.

## Why

`InlineSource.readResource` returned `string | null` (sync); `McpSource.readResource` returned `Promise<ResourceData | null>` (async, structured). The type guard `isResourceReader` only checked for method existence, so both passed. Consumers — `nb__read_resource`, `POST /v1/resources/read`, `runtime.getAppSkillResource` — accessed `.text` / `.blob` / `.mimeType` on what was sometimes a string. Result: every `ui://` resource published by the platform built-ins (settings panel, home dashboard, automations panel, conversations browser, files browser, usage dashboard, nb core resources) was unreachable to the agent. The bug had already bitten once and been worked around in `briefing-collector.ts` with a dual-shape branch.

The fix scope is "make the unsoundness impossible": remove the second implementation. Every source goes through the MCP SDK's `Server` over a `Transport`. Built-ins use `InMemoryTransport.createLinkedPair()`; nothing else changes externally.

## What changed

**New**
- `src/tools/in-process-app.ts` — `defineInProcessApp({ name, version, tools, resources, placements, instructions })` builds an `McpSource` backed by an in-process `Server` over an `InMemoryTransport`. Same authoring ergonomic as the former `InlineSource` (function + JSON schema). JSON Schema input validation preserved at the call boundary.
- `inProcess` mode on `McpSource.McpTransportMode` — factory that produces a fresh Server + linked transport pair per `start()` for restart safety.
- `McpSource.getPlacements()` exposes UI placements declared on the inProcess mode.

**Migrated to in-process MCP**
- `nb` core (system tools), `settings`, `home`, `automations`, `conversations`, `files`, `usage`.
- All resources keyed by full `ui://<name>/<path>` URIs (the protocol's identity).
- `createPlatformSources(runtime, eventSink)` returns sources already-started.
- `createSystemTools(...)` is now async — returns a started `McpSource`.

**Deletions**
- `src/tools/inline-source.ts` (`InlineSource`, `InlineToolDef`).
- `ResourceReader` interface and `isResourceReader` from `src/tools/types.ts`.
- `instanceof InlineSource` branch in `handleResourceProxy` — both built-ins and bundles now route through `runtime.readAppResource`.
- Dual-shape `string | object` branching in `briefing-collector.ts`.

**Behavior change**
- `runtime.readAppResource` no longer gates on `lifecycle.getInstance` — registry membership is the authoritative "is this app available to this workspace?" check. Without this, platform sources (which have no lifecycle entry) would never resolve through the unified path.

## Public API

- `InlineSource` / `InlineToolDef` removed. External callers should switch to `defineInProcessApp` (returns an `McpSource`); `InlineToolDef` becomes `InProcessTool` with the same field shape.
- `Tool.source` for migrated platform sources changes from `inline:<name>` to `mcpb:<name>` (label only — no logic depended on the prefix).

## Test plan

- [x] `bun run verify` clean (format, lint, type, cycles, 1908 unit tests, 383 integration tests, smoke).
- [x] New regression test in `system-tools.test.ts`: `nb__read_resource` resolves a `ui://` resource published by an in-process source — the exact #90 failure mode.
- [x] `defineInProcessApp` contract tests (`test/unit/tools/in-process-app.test.ts`) verify the schema-validation and unknown-tool-error guarantees the former `InlineSource` enforced.
- [ ] Manual: hit `/v1/apps/settings/resources/primary` in a running dev instance and confirm the panel HTML comes back (was previously broken via `runtime.readAppResource` because the lifecycle gate excluded platform sources).
- [ ] Manual: invoke `nb__read_resource` with `ui://settings/panel` from a chat and confirm the body is returned (the original #90 failure case).


mgoldsborough:
QA review fixes pushed in [`b57f56a`](https://github.com/NimbleBrainInc/nimblebrain/pull/91/commits/b57f56a).

| QA item | Action |
|---|---|
| W1: `automations.ts` `stop()` ordering | **Fixed** — `try/finally` so transport cleanup runs even if scheduler ever throws |
| W2: stale `InlineSource` doc comments (6 files) | **Fixed** — zero `InlineSource` / `InlineToolDef` mentions remain in `src/` |
| S1: duplicated `bytesToBase64` | **Fixed** — extracted to `src/util/base64.ts`, both call sites import from it (avoids upward `tools/ → api/` dep) |
| S2: `getCoreResource()` test-only | **Fixed by deletion** — `core-resources.test.ts` now asserts via `buildCoreResourceMap()` |
| S3: performance baseline | **Deferred** — tracked in #92 |

`bun run verify` clean (exit 0).

PR #118, opened 2026-04-28
fix: translate enabled→adaptive+effort for Opus 4.7

mgoldsborough:
## Summary

Anthropic rejects `thinking.type=enabled` for Claude Opus 4.7 with:

> "thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior.

The engine now gates on `supportsEnabledThinking()` and translates the platform's enabled+budget into adaptive+effort for adaptive-only models. Budget→effort mapping (≤4K→low, ≤16K→medium, ≤32K→high, else max) preserves the operator's intended thinking spend without leaking provider-specific knowledge into the resolver.

Models that still accept `enabled` (Sonnet 4.6 and older) keep current behavior verbatim. Adding future adaptive-only models is a one-line edit to `ADAPTIVE_ONLY_THINKING_MODELS` in `src/model/catalog.ts`.

## Test plan

- [x] `bun test test/unit/resolve-thinking.test.ts test/unit/engine.test.ts` — 71 pass
- [x] `bun run check` — clean
- [x] `bun run verify:static` — clean (pre-existing biome warnings on unrelated files)
- [ ] Manual: `ws_nimblebrain_shared` chat in prod hq with thinking=Default after deploy


mgoldsborough:
Pushed `e347dad` addressing W1 + S1. Reasoning on the two I skipped:

**S2 (threshold boundary comment):** The existing doc comment in `budgetToEffort` already shows the full `maxOutputTokens → budget → tier` calibration table, which is more actionable than a "thresholds are upper edges" disclaimer would be. The "decimal-K vs binary-K" framing in the review is also off — 4096/16384/32768 are all binary K, same as the `safeThinkingBudget` outputs. If `MIN_VISIBLE_OUTPUT_TOKENS` changes, the calibration table in the comment becomes the wrong-answer signal, not the threshold constants.

**S3 (redundant provider re-parse in `supportsEnabledThinking`):** The function is exported. Its correctness shouldn't depend on the caller already being inside an `if (provider === "anthropic")` block — a future caller (admin tool, debug surface, refactored engine) shouldn't have to know that gate exists. Defensive code on an exported API boundary is appropriate, and the cost is one `indexOf` + slice.

Issue #121, opened 2026-04-28
Phase 3 gap: implement skills__history / skills__diff / skills__rollback


mgoldsborough:
## Background

Per the skill management design doc Phase 3, three versioning query tools were specified alongside the mutation tools:

- `skills__history(id)` — list all snapshot versions of a skill
- `skills__diff(id, version_a, version_b?)` — diff two versions
- `skills__rollback(id, version)` — restore a skill to a prior version

Phase 3 (#112) shipped `skills__create` / `update` / `delete` / `activate` / `deactivate` / `move_scope` and the `_versions/` snapshot mechanism, but not the query/restore tools. The data is being captured (every mutation snapshots to `_versions/<name>/<utc-ts>.md`) — there's just no tool surface to query it.

## Why it matters

Versioning without restore is half a feature. Operators can't answer \"what did this skill say last week?\" or \"undo my last edit\" without filesystem access. The doc's claim that \"versioning means edits are reversible\" is currently aspirational.

## Acceptance

- `skills__history(id)` → list of `{version, ts, size}` for the snapshot files in `_versions/<name>/`
- `skills__diff(id, version_a, version_b?)` → unified diff text. Default `version_b` to current.
- `skills__rollback(id, version)` → snapshot the current state, then `cp _versions/<name>/<ts>.md <id>`. Returns the new current version.
- All three are RBAC-gated and hidden from `tools/list` (mutation tools are `[HIDDEN]` per the design).
- `_versions/_versions/` infinite recursion shouldn't happen (snapshot the snapshot is meaningless) — guard against it.

## Related

- Design doc lines 327-340
- The `_versions/` directory layout and naming convention is already in place

Issue #120, opened 2026-04-28
Phase 2 gap: implement skills__search


mgoldsborough:
## Background

Per the skill management design doc Phase 2, `skills__search` was specified as a content-search tool over the skill catalog:

> `skills__search(query, scope?, layer?)`
> → `[{id, snippet, line_no}]`
> # filesystem: greps files
> # bundle:     fetches each skill resource and searches in-memory (cache the content briefly)

Phase 2 (#102) shipped `skills__list` / `skills__read` / `skills__active_for` / `skills__loading_log` but not `skills__search`.

## Why it matters

Without search, an agent that wants to answer \"is there a skill that talks about X\" has to call `skills__list` then `skills__read` on every entry. With ~10-20 skills per workspace this is wasteful but tolerable; with scaled customization it becomes prohibitive.

Phase 4 (`skills__author`, audit automation) will rely on search to detect duplicate or contradictory content cheaply.

## Acceptance

- New tool `skills__search(query, scope?, layer?)` returning matches with `{id, snippet, line_no}`
- Filesystem skills: grep with line numbers + ~80-char snippet around match
- Bundle skills: load via `resources/read`, search in-memory; cache briefly per discovery cycle
- Honors `scope` and `layer` filters from existing `skills__list` patterns

## Related

- Design doc lines 266-269

Issue #119, opened 2026-04-28
Phase 1 gap: implement compose_effective_context debug surface


mgoldsborough:
## Background

Per the [skill management design doc](https://github.com/NimbleBrainInc/nimblebrain-ops/blob/main/research/SKILL_MANAGEMENT_DESIGN.md) Phase 1, `compose_effective_context` was specified as the unified debug surface for context assembly:

> `compose_effective_context(conversation_id?, bundle?)`
> → `{text, layers: [{source, kind, id, tokens, text}]}`
>
> Returns the fully composed system prompt as it would be (or was) emitted for this conversation. Includes Layer 1 + Layer 2 (bundle instructions) AND Layer 3 (skills) AND static sources. Crosses both `bundle_instructions.loaded` and `skills.loaded` plus the snapshot — single tool, all layers. Read-only. **The single most useful debug surface for \"why is the agent doing X?\"**

Phase 1 (#98) shipped layered composition with provenance markers but did not implement this tool. It's been a notable gap during recent prod debugging sessions where reconstructing \"what was in the agent's prompt for run X\" requires manual jsonl spelunking.

## Acceptance

- New tool `nb__compose_effective_context` (or sibling under the platform source) accepting optional `conversation_id` and `bundle` filters
- Returns the fully composed prompt text + per-layer breakdown with `source`, `kind`, `id`, `tokens`, body excerpt
- Reads from `context.assembled` events when given a conversation; otherwise simulates assembly with current state
- Available to agents AND web shell (matches the dual-surface model the doc specifies)

## Related

- Design doc lines 275-283
- Surfaces this would replace: ad-hoc grep through conversation jsonl + manual prompt reconstruction

PR #122, opened 2026-04-28
fix: skills tool surface honest about its input contract

mgoldsborough:
## Summary

Five related fixes (F1, F2, F4, F6, F8 from the audit in conv `3f1d097145134000`) so the platform skills tools are usable from an agent loop without prior knowledge of the filesystem layout. None are new features — each closes a gap between what the [skill management design doc](https://github.com/NimbleBrainInc/nimblebrain-ops/blob/main/research/skill-management-design.md) specifies and what shipped in Phase 2 / Phase 3.

| Fix | Before | After |
|---|---|---|
| **F1** | `skills__list` returns `"12 skills (12 org)"` — counts only | One row per skill with id, layer, scope, type, priority, status, truncated description |
| **F2** | `scopeOfPath` falsely classifies any unknown path as `"bundle"` → bare ids hit "Bundle (Layer 1) skills are vendored" | Explicit bundle-roots classification + `null` for unclassified |
| **F4** | Error says `(platform/workspace/user/builtin)` — stale scope names + wrong-format hint | Honest message describing the real contract (filesystem path or bundle URI) |
| **F6** | Mutation tool descriptions: `"id (filesystem path)"` with no source | Point at `skills__list` and reject bare/scope-prefixed forms in the description |
| **F8** | `bundleNotMutable()` returns bare prose | Returns the `{error, layer, suggested_action, message}` shape from the design doc |

## Why this matters

The audited conversation showed the agent making 17 doomed tool calls in increasingly desperate guess-shapes, ultimately giving up and offering to "write the six skill bodies as text here" for the user to paste manually. Total session cost: 382K tokens. The bugs were structural — the surface forced the agent to guess paths it had no way to discover, and every guess returned a misleading error.

F1 alone unblocks the agent (paths visible in list output → directly feedable into read/update/delete). F2 + F4 stop the misleading errors. F6 + F8 reinforce.

## Out of scope (filed as separate issues)

- `compose_effective_context` (Phase 1 deliverable not shipped) — #119
- `skills__search` (Phase 2 deliverable not shipped) — #120  
- `skills__history` / `diff` / `rollback` (Phase 3 deliverables not shipped) — #121
- `skills__active_for` defaulting to current conversation (F7 in audit) — small follow-up
- `skill://NAME` URI resolution for L3 (F3 in audit) — deferred per discussion: L3 stays path-only, `skill://` reserved for bundle resources
- `update_deal` schema/coercion bugs in synapse-crm — NimbleBrainInc/synapse-crm#3

## Test plan

- [x] `bun test test/unit/skills.test.ts test/unit/skills/ test/integration/skill-lifecycle.test.ts test/integration/skills-read-tools.test.ts` — 107 pass (was 105; added 2 new tests for F1 textContent and F2 unrecognized-id error path)
- [x] `bun run check` — clean
- [x] `bun run verify:static` — clean (pre-existing biome warnings on unrelated files)
- [ ] Manual: retry the audited workflow in `ws_nimblebrain_shared` once deployed; agent should be able to list → read → update org-tier skills without guessing paths


mgoldsborough:
QA pass — pushed `84c04a4`.

**(1) Stale `assertSymlinkBoundaryOrThrow` comment:** fixed. The function's behavior didn't change — it computes `realScope` locally at lines 767-771 and the local fallback to `"bundle"` still applies for outside-workdir paths. Comment now describes that local logic instead of attributing it to `scopeOfPath`.

**(2) Missing `bundle` / `bundleVersion` in `bundleNotMutable()`:** TODO comment added. Populating those properly means threading bundle context through three mutation handlers (parse `skill://<bundle>/<name>` URI authority, derive bundle from path for filesystem-rooted bundle skills) plus a runtime bundle-registry lookup for the version. That's a substantive enough refactor to warrant its own PR — agreeing with your "acceptable to defer" framing.

**(3) `summarizeList` truncation at scale:** acknowledged, no action. Current largest tenant: 12 skills. The right fix when scale matters is pagination on the schema (`skills__list` already takes filters; cursor would be the natural extension), not a row cap on text content. That's a Phase-4-ish concern for the platform; not blocking the agent today.

PR #124, opened 2026-04-28
fix(ui): show "Calling X…" while model streams tool input

mgoldsborough:
## Summary

- The chat UI went visually dark for the entire duration of large tool-input streams. Reproduced in production `conv_6cf2bbacee2745cd` (hq tenant): a 45 KB `set_source` call took 4m18s with no breathing border, no label, and no inline placeholder — only the static "Waiting for response…" textarea hint. Cause: `src/model/stream.ts` was discarding the AI SDK's `tool-input-{start,delta,end}` parts that fire while the model assembles a tool call within an iteration.
- Surface those parts as two new engine events (`tool.preparing` on `tool-input-start`, `tool.preparing.done` on `tool-input-end`) and forward them through the existing `EventSink` → SSE path. `useChat` adds a `"preparing"` `StreamingState` plus a `preparingTool { id, name }` field; `MessageInput` renders **`Calling <tool>...`** with the breathing border, and `MessageList` shows an inline placeholder under the last assistant block.
- Provider-agnostic: AI SDK V3's `LanguageModelV3StreamPart` is the unified abstraction — verified that Anthropic, OpenAI, and Google adapters all normalize their native streaming protocols into `tool-input-{start,delta,end}` (against installed `@ai-sdk/*` dist). Gateway passes them through. Zero per-provider code. Per-char `tool-input-delta` is intentionally swallowed in `stream.ts` — only `start`/`end` propagate, so steady-state SSE traffic is unchanged.

## Design notes

Three options were considered: (A) UI-side idle timeout, (B) forward AI SDK part events end-to-end, (C) engine heartbeat. **B** chosen — A and C fix the visible symptom but leave information loss in place; B surfaces the actual phase the model is in, names the tool 4 minutes earlier than today on large calls, and is provider-agnostic. The bigger refactor of `streamingState` into a full phase union is left for follow-up; this PR adds `"preparing"` as a new value plus a sibling `preparingTool` field for the smallest blast radius that still fixes the bug and surfaces the tool name.

## Test plan

- [x] `bun run verify` green (static + 416 unit + 24 web + integration + 17 smoke)
- [x] New `model-stream.test.ts` cases: `tool-input` callbacks fire once per tool with correct ordering; not invoked for text-only responses
- [x] Echo model fixture updated to emit `tool-input-{start,delta,end}` parts so existing tool-call tests exercise the new code path automatically
- [ ] Manual: send a prompt that triggers a long tool-args stream (e.g. set_source with a large body) — confirm "Calling set_source..." appears immediately after the model starts the tool block, persists for the full streaming window, and transitions to "Working..." once the engine begins execution
- [ ] Manual: confirm short / fast tool calls (set_source on a small doc, get_workspace) flicker through `"preparing"` → `"working"` cleanly with no UI artifacts
- [ ] Manual: verify shared-conversation remote stream (`processRemoteStreamEvent`) shows the same indicator for a participant watching another user's session

## Files

- `src/engine/types.ts` — new event types in the union
- `src/model/stream.ts` — consume `tool-input-*`; new optional callbacks
- `src/engine/engine.ts` — wire callbacks → emit events
- `src/api/handlers.ts` — forward to SSE allow-list
- `web/src/types.ts` — wire event types and `ChatStreamEventMap` entries
- `web/src/hooks/useChat.ts` — `"preparing"` state + `preparingTool` field; local and remote handlers
- `web/src/components/{ChatPanel,MessageInput,MessageList}.tsx` — thread state, render label and inline placeholder
- `test/helpers/echo-model.ts` — emit `tool-input-{start,delta,end}` parts before `tool-call`
- `test/unit/model-stream.test.ts` — callback ordering tests

### DIFF
--- a/src/api/handlers.ts
+++ b/src/api/handlers.ts
@@ -147,6 +147,8 @@ export async function handleChatStream(
           event.type === "chat.start" ||
           event.type === "text.delta" ||
           event.type === "reasoning.delta" ||
+          event.type === "tool.preparing" ||
+          event.type === "tool.preparing.done" ||
           event.type === "tool.start" ||
           event.type === "tool.done" ||
           event.type === "llm.done" ||
--- a/src/engine/engine.ts
+++ b/src/engine/engine.ts
@@ -372,6 +372,8 @@ export class AgentEngine {
             },
             (text) => this.events.emit({ type: "text.delta", data: { runId, text } }),
             (text) => this.events.emit({ type: "reasoning.delta", data: { runId, text } }),
+            (id, name) => this.events.emit({ type: "tool.preparing", data: { runId, id, name } }),
+            (id) => this.events.emit({ type: "tool.preparing.done", data: { runId, id } }),
           ),
         );
         const llmMs = Math.round(performance.now() - llmStart);
--- a/src/engine/types.ts
+++ b/src/engine/types.ts
@@ -45,6 +45,8 @@ export type EngineEventType =
   | "run.start"
   | "text.delta"
   | "reasoning.delta"
+  | "tool.preparing"
+  | "tool.preparing.done"
   | "tool.start"
   | "tool.done"
   | "tool.progress"
--- a/src/model/stream.ts
+++ b/src/model/stream.ts
@@ -18,6 +18,19 @@ export async function callModel(
   options: LanguageModelV3CallOptions,
   onTextDelta: (text: string) => void,
   onReasoningDelta?: (text: string) => void,
+  /**
+   * Called once when the model begins emitting a tool-call block, with
+   * the tool name already known. Lets the engine surface "Calling X…"
+   * during the dark gap where the model is streaming a large tool
+   * input — `tool.start` only fires after `callModel` returns.
+   *
+   * Provider-agnostic: AI SDK V3 normalizes `tool-input-start` across
+   * Anthropic / OpenAI / Google. Providers that never emit it simply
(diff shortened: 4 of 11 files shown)


mgoldsborough:
## QA review adjudication

**Note 1 — Resolved** in fa1747e. Cleared `preparingTool` in `text.delta` and `reasoning.delta` handlers (both local and remote paths). Cheap insurance, makes the invariant `preparingTool ↔ streamingState === "preparing"` explicit.

**Note 2 — Deferred to follow-up.** SSE allow-list missing `error` / `run.error` is a pre-existing gap (same risk has existed for `working` / `analyzing`). Affects shared-conversation remote viewers across all stream states, not specific to `"preparing"`. Will file a follow-up issue.

**Note 3 — Deferred to follow-up.** No existing chat-stream integration test exercises tool calls in SSE — adding one needs new fixtures (tool router + mock MCP setup). Manual test plan covers the wire path; the unit test in `model-stream.test.ts` locks the engine→callback contract. The full wire-level lock-in is worth doing but is more scaffolding than this PR warrants.

**Note 4 — Skipped.** A doc comment on `EngineEventType` listing persisted vs transient consumers goes stale fast and isn't enforced. The convention is already explicit at four consumer-side filters (`event-sourced-store`, `posthog-sink`, `workspace-log-sink`, `handlers.ts` SSE allow-list). If we want to enforce persistence intent in the type system, the right move is a separate `PersistedEvent` union, not a comment — that's a separate refactor.

PR #125, opened 2026-04-29
feat: emit contentHash on skills.loaded events

mgoldsborough:
## Summary

Adds SHA-256 hex of each skill's body to `SkillsLoadedEntry`, populated whenever the runtime emits a `skills.loaded` event. **Telemetry-only change — no behavior impact today.** Foundation for the upcoming `compose_effective_context` debug tool (#119).

## Why

We're shipping #119 next — a tool that returns "what was in the system prompt for run X, with provenance per layer." For historical-mode reconstruction, the tool needs to detect whether a skill's body has been mutated between when it loaded and when the operator inspects the run. Without that, the tool would silently display the *current* skill content as if it were the historical content (the body that actually fed the model).

The cheapest reliable detector is a content hash recorded at load time:
- **Hash matches current source** → display the body verbatim, full fidelity.
- **Hash differs** → look up the recorded hash against `_versions/` snapshots to find the actual loaded body, or surface an explicit "this skill changed since this run" warning.

Doing this without the hash would require either (a) embedding the full body in every `skills.loaded` event (10-100× event size; rejected as too expensive), or (b) accepting silent fidelity loss (rejected as user-hostile during debugging).

## Cost

~64 bytes per skill per turn (SHA-256 hex is 64 chars). For a typical conversation with 10 skills loading per turn × 30 turns, that's ~20 KB of additional telemetry over the conversation's lifetime. Negligible against the existing event volume.

## Contract: `contentHash` is **required**, not optional

The platform isn't deployed anywhere with persisted `skills.loaded` events, so there's no historical data to be backward-compatible with. Required field gives clean, first-principles types: emitters guarantee the field, consumers can rely on it. Test fixtures across three files (`skills/events`, `tools/platform/skills-tools`, `runtime/skills-loaded-payload`) updated with `TEST_HASH` sentinels for the cases that don't verify hash math.

(An earlier commit on this branch made the field optional. That decision was reversed — see commit `f2a8501`. Don't read the optional pattern as a guide to how new fields should be modeled.)

## Refactor included

Extracted `buildSkillsLoadedPayload` and `hashSkillBody` from `runtime.ts` into a small focused module (`src/runtime/skills-loaded-payload.ts`) so the function can be unit-tested directly without polluting `runtime.ts`'s public surface. Pure function over inputs; no behavior change beyond the new field.

## Test plan

- [x] `bun test test/unit/runtime/skills-loaded-payload.test.ts` — 9 new tests (hash determinism, body sensitivity, empty-string sentinel, propagation through `buildSkillsLoadedPayload`)
- [x] `bun run test:unit` — 2137 pass (was 2128; added 9)
- [x] `bun test test/integration/skill-lifecycle.test.ts test/integration/skills-read-tools.test.ts` — 7 pass (no regression)
- [x] `bun run verify:static` — clean (the gate CI runs; matches `format:check + lint + check + check:cycles`)

## Out of scope

- The debug tool that consumes the hash → #119 (next PR, depends on this landing).
- Hashes for bundle instructions / overlays / app instructions → separate PRs if needed; L3 skills are the most volatile so the highest-leverage layer to instrument first.
- Snapshotting full layer bodies into events → not doing it. Hashes give enough fidelity for the debugging use cases we care about, at 1/100th the cost.
- Validation of persisted-event shape on read (the trust boundary at `event-sourced-store.ts:607`) → its own audit; flagged with a comment at the cast site.

## Refs

- `SKILL_MANAGEMENT_DESIGN.md` § "Provenance" (Phase 5 deterministic recomputation pattern)
- Closes the telemetry prerequisite for #119

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 
 ### Added
 
-- `skills.loaded` events now carry a `contentHash` (SHA-256 hex) per skill — telemetry foundation for the upcoming `compose_effective_context` debug tool (#119), which uses the hash to detect whether an on-disk skill body has been mutated between when the skill loaded and when an operator inspects the run. ~64 bytes per skill per turn; backward-compatible (field is optional on the type so old persisted events still parse).
+- `skills.loaded` events now carry a `contentHash` (SHA-256 hex) per skill — telemetry foundation for the upcoming `compose_effective_context` debug tool (#119), which uses the hash to detect whether an on-disk skill body has been mutated between when the skill loaded and when an operator inspects the run. ~64 bytes per skill per turn.
 - Extended-thinking config (`thinking: off | adaptive | enabled` + `thinkingBudgetTokens`) in runtime config and Settings → Model; defaults to `adaptive` for catalog-flagged reasoning models, `off` otherwise ([#109](https://github.com/NimbleBrainInc/nimblebrain/pull/109)).
 - Reasoning (extended-thinking) content is now captured end-to-end: `stream.ts` handles `reasoning-*` parts, the engine emits `reasoning.delta` SSE events, and the chat UI renders a collapsed "Thoughts" block above the assistant text. Empty turns with non-stop finishReason now emit a placeholder so replay surfaces truncations.
 - `nb__read_resource` system tool — the agent can now load `skill://` / `ui://` resources advertised by an installed bundle's MCP server ([#3](https://github.com/NimbleBrainInc/nimblebrain/pull/25)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Added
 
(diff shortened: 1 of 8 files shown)


mgoldsborough:
QA pass — pushed `d742a95`.

**C1 (lint will fail CI):** Fixed via `biome check --write` on the two affected files. Process miss: I ran `bun run check` locally instead of `verify:static`, which runs only `tsc --noEmit` and skips biome. Updating my pre-PR habit to `verify:static` (the script CI actually invokes).

**W1 (PR body out of sync):** Updated via `gh pr edit`. The new body explicitly documents the "required, not optional" decision and points at commit `f2a8501` for context.

**W2 (orphaned JSDoc):** Removed.

**W3 (unsafe cast at the trust boundary):** Added a comment at the cast site — but framed as "the cast is the existing pattern for every event type, broader event-shape validation is its own audit." A contentHash-specific note would have misrepresented the underlying issue (the cast trusts arbitrary on-disk JSON regardless of which field changed). The comment names the assumption and points future contributors at validation-at-read as the right defense.

`verify:static` green, 2137 unit tests pass.

PR #128, opened 2026-04-29
feat: HTTP proxy primitive for bundles

mgoldsborough:
## Summary

Adds an opt-in same-origin HTTP proxy so a bundle can expose its own loopback HTTP server (e.g. `astro preview`, a Jupyter kernel) inside the platform UI via `/v1/ws/<wsId>/apps/<bundle>/<mount>/*`. The bundle declares `_meta["ai.nimblebrain/http-proxy"]` in its manifest; the platform proxies browser requests to its loopback server with credentials stripped, X-Frame-Options set to SAMEORIGIN, and per-workspace kill switch.

Driven by `synapse-apps/synapse-astro-editor`, which uses it to embed a built Astro site in the editor preview iframe — but the primitive is generic.

### Two commits

- `eb07bee` — initial route + manifest plumbing
- `d6c18e1` — workspace-in-URL refactor + hardening pass (header strip, gzip handling, X-Frame default, loopback restriction, NB_PUBLIC_ORIGIN env injection, Hono ordering fix)

### Trust model

Documented at the top of `src/api/routes/proxy.ts`:

- A bundle declaring http-proxy runs as same-origin code in the user's session — the iframe is sandboxed `allow-scripts allow-same-origin`. Treat these bundles like browser extensions.
- Defenses enforced: loopback-only target, request-header strip (Authorization/Cookie/X-Workspace-Id), response-header strip (Set-Cookie/CSP/X-Frame-Options), per-request membership check, `Workspace.allowHttpProxy = false` kill switch.
- Defenses NOT today: cross-origin isolation per bundle (would need subdomain-per-bundle + COEP). For untrusted-bundle marketplaces, that's the next investment.

### Why workspace in the URL

Browser iframe loads can't set custom headers, so the original `X-Workspace-Id` header pattern 400'd every preview load. Workspace ID lives in the URL path now, validated with `WORKSPACE_ID_RE` and membership-checked per request inside the handler. Proxy route is mounted before other sub-apps in `app.ts` because their `.use("*", requireWorkspace)` middleware would otherwise intercept `/v1/ws/*`.

### Bundle env

Bundles get three new env vars at spawn:

- `NB_WORKSPACE_ID` — which workspace this instance is for
- `NB_PROXY_PREFIX` — `/v1/ws/<wsId>/apps/<name>/<mount>` for use as `astro --base` or equivalent
- `NB_PUBLIC_ORIGIN` — operator-set platform origin so bundles can declare it in `_meta.ui.csp.{frameDomains,connectDomains}` on their UI resources

## Test plan

- [x] `bun run verify` — 418 unit/integration + 17 smoke pass, 0 fail
- [x] Manual end-to-end via `synapse-astro-editor`: bundle loads target site (Bayze) with full styling in editor preview iframe; edits trigger rebuild; preview reflects changes
- [x] Verified loopback-only restriction: external host in `target` is rejected with warning at manifest parse
- [x] Verified header stripping: bundle's loopback server sees no Authorization/Cookie headers
- [x] Verified gzip handling: gzipped upstream responses render correctly in browser
- [x] Verified per-workspace kill switch: `allowHttpProxy: false` on workspace returns 403

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - `UserMenu` in the shell sidebar (avatar + display name + popover with Profile settings + Sign out).
 - Top-level `/profile` route. Identity isn't a setting; un-nested from `/settings/*`.
 - Org-admin gate on `set_model_config` — backend now refuses non-org-admin writes (was UI-only via RouteGuard). Distinguishes "no identity" (cron, automations) from "wrong role" so debug logs make non-user code paths obvious.
+- HTTP proxy primitive (`_meta["ai.nimblebrain/http-proxy"]`). Bundles can expose a loopback HTTP server (e.g. `astro preview`, Jupyter kernel) through the platform at `/v1/ws/<wsId>/apps/<bundle>/<mount>/*`. Loopback-only target, credentials and `Accept-Encoding` stripped on forward, `Set-Cookie`/CSP/X-Frame-Options stripped on response, per-workspace kill switch via `Workspace.allowHttpProxy`. Bundles get `NB_WORKSPACE_ID`, `NB_PROXY_PREFIX`, `NB_PUBLIC_ORIGIN` in their env at spawn ([docs](https://docs.nimblebrain.ai/apps/http-proxy/)).
 
 ### Changed
 
--- a/src/api/app.ts
+++ b/src/api/app.ts
@@ -10,6 +10,7 @@ import { eventRoutes } from "./routes/events.ts";
 import { healthRoutes } from "./routes/health.ts";
 import { mcpRoutes } from "./routes/mcp.ts";
 import { mcpAuthRoutes } from "./routes/mcp-auth.ts";
+import { proxyRoutes } from "./routes/proxy.ts";
 import { resourceRoutes } from "./routes/resources.ts";
 import { toolRoutes } from "./routes/tools.ts";
 import { wellKnownRoutes } from "./routes/well-known.ts";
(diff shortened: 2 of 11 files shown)


mgoldsborough:
QA feedback addressed in two follow-up commits:

- `ad3d635` — source fixes: XFF stripping, identity-encoding strategy, dev-mode trust-model note, reserved-mount removal, stale path comments, CHANGELOG entry
- `797ec4a` — tests: 16 unit cases covering every \`extractHttpProxy\` rejection branch + 16 integration cases covering each defense in the trust-model header (handler validation, kill switch, membership, header strip on forward + response, X-Frame-Options: SAMEORIGIN)

Notable architectural decisions per the review thread:

**X-Forwarded-For** — added to \`REQUEST_HEADERS_STRIPPED\` rather than rewritten. Browser-supplied XFF is unverifiable; bundles run on loopback and don't need it. If a future use case needs it, we'll set it from a server-side source.

**Content-encoding** — moved from response-side strip to request-side \`Accept-Encoding: identity\`. The previous strip relied on \"Bun's fetch decompresses gzip\" — true today but couples our correctness to Bun's per-encoding behavior. Identity in, identity out, no encoding state to track. Discovered via the integration test that simply stripping isn't enough: Bun's outbound \`fetch\` auto-injects \`Accept-Encoding: gzip, deflate, br, zstd\` when absent. We now SET \`identity\` explicitly to override that default. Updated the trust-model comment to reflect the actual mechanism.

**Reserved-mount list** — deleted. With workspace-in-URL, \`/v1/ws/<wsId>/apps/<bundle>/resources\` doesn't collide with \`/v1/resources\`, so the guard had no purpose. If a future route shape collides, we'll re-add with a comment naming the collision.

**Dev-mode bypass** — left as-is (consistent with every other workspace-scoped route on the platform), but documented in the trust-model header so operators don't assume membership is always enforced.

**CHANGELOG** — added under \`[Unreleased] → Added\` with the operator-relevant bits (route shape, env vars, kill switch).

\`bun run verify\` clean: 2141 unit + 213 web + integration + 17 smoke pass, 0 fail.

PR #129, opened 2026-04-29
feat: compose__effective_context debug tool

### LINKED ISSUE #119: Phase 1 gap: implement compose_effective_context debug surface

mgoldsborough:
## Summary

Single-call answer to "what's in the system prompt for this conversation, with provenance per layer." Replaces the manual jsonl-grep workflow that the recent prod incidents (Opus 4.7 thinking, skills tools surface) each required for diagnosis. Closes #119.

## Two modes, one tool

```
compose__effective_context({
  conversation_id?,   // defaults to current conv (the F7 pattern)
  run_id?,            // historical mode trigger
  bundle?,            // filter to one bundle's contributions
})
```

**Live (default).** Re-gathers the same inputs `runtime.chat()` uses (workspace, identity, registry, skills, instructions, apps, prefs) and runs `composeSystemPromptTraced` to produce the full per-layer breakdown for the current state.

**Historical (\`run_id\` set).** Reads the `skills.loaded` event for the given run from the conv jsonl and audits each layer-3 skill:
- Re-reads the on-disk body, computes its SHA-256, compares to the recorded `contentHash` (from #125).
- **Match** → display the body verbatim (\`hashStatus: "match"\`).
- **Drift + matching `_versions/` snapshot** → display the snapshot body (\`hashStatus: "recovered"\`).
- **Drift, no snapshot** → display current with a warning that the body may differ (\`hashStatus: "drift"\`).
- **File deleted** → \`hashStatus: "missing"\`.

## Three structural changes

1. **New traced compose function.** Added `composeSystemPromptTraced` to `prompt/compose.ts` returning `{text, layers, totalTokens}`. The existing `composeSystemPrompt` is now a thin wrapper that calls the traced variant and returns `.text`. Joined layer texts are byte-identical to the legacy string output — verified by a property test.

2. **New `compose` platform source.** New file `src/tools/platform/compose.ts` exposing one tool (\`compose__effective_context\`). Visible to agents; read-only; no security implications.

3. **Two private runtime helpers made public.** `buildAppsList(wsId)` and `readPromptOverlays(wsId)` — needed by the tool's live-mode gather path. Workspace-scoped, no privilege escalation.

## Layer kinds

| kind | granularity | subItems |
|---|---|---|
| `core_skill` | one per priority ≤ 10 context skill | — |
| `default_identity` | one (only when no core skills) | — |
| `user_context_skill` | one per priority > 10 context skill | — |
| `user_prefs` | one | — |
| `participants` | one (shared convs only) | — |
| `workspace_context` | one | — |
| `org_overlay` / `workspace_overlay` | one each (when non-empty) | — |
| `layer3_skills` | one section | per-skill (hash-verified, bundle-attributed) |
| `apps` | one section | per-app (bundle-attributed) |
| `app_state` / `focused_app` / `matched_skill` | one each (when set) | — |

Granular for operator-authored content (skills, apps), single rows for runtime-derived state.

## Honest scope limits

- **Historical mode reconstructs layer-3 skills only.** Identity, prefs, overlays, apps, focused-app, and app-state aren't recorded in events with enough detail to reconstruct; the tool surfaces a \`non_l3_layers_omitted\` warning directing the caller at live mode for the rest.
- **Mutation detection is layer-3-only.** Hashes only exist for L3 skills (#125). Other layers can drift between run-time and inspect-time without a hash check. Future PRs can extend hashing if needed.

## Test plan

- [x] `bun test test/unit/prompt.test.ts` — 70 pass (62 existing + 8 new for `composeSystemPromptTraced`)
- [x] `bun test test/integration/compose-effective-context.test.ts` — 8 pass (live mode, historical match/drift, bundle filter, conv-id resolution)
- [x] `bun run test:unit` — 2149 pass (was 2137; added 12)
- [x] `bun run verify:static` — clean (only pre-existing biome warnings on unrelated files)

## Refs

- `SKILL_MANAGEMENT_DESIGN.md` § Phase 1 (\`compose_effective_context\`)
- Closes #119
- Builds on #125 (contentHash telemetry — historical hash verification depends on it)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@
 
 ### Added
 
-- `skills.loaded` events now carry a `contentHash` (SHA-256 hex) per skill — telemetry foundation for the upcoming `compose_effective_context` debug tool (#119), which uses the hash to detect whether an on-disk skill body has been mutated between when the skill loaded and when an operator inspects the run. ~64 bytes per skill per turn.
+- `compose__effective_context` debug tool — single-call answer to "what's in the system prompt for this conversation, with provenance per layer." Live mode returns the full traced composition; historical mode (`run_id` set) reads the recorded `skills.loaded` event for that run and verifies each layer-3 skill's `contentHash` against current source, with `_versions/` snapshot recovery on drift. Bundle filter narrows to one app's contributions ([#119](https://github.com/NimbleBrainInc/nimblebrain/issues/119)).
+- `skills.loaded` events now carry a `contentHash` (SHA-256 hex) per skill — telemetry foundation for the `compose__effective_context` debug tool. ~64 bytes per skill per turn.
 - Extended-thinking config (`thinking: off | adaptive | enabled` + `thinkingBudgetTokens`) in runtime config and Settings → Model; defaults to `adaptive` for catalog-flagged reasoning models, `off` otherwise ([#109](https://github.com/NimbleBrainInc/nimblebrain/pull/109)).
 - Reasoning (extended-thinking) content is now captured end-to-end: `stream.ts` handles `reasoning-*` parts, the engine emits `reasoning.delta` SSE events, and the chat UI renders a collapsed "Thoughts" block above the assistant text. Empty turns with non-stop finishReason now emit a placeholder so replay surfaces truncations.
 - `nb__read_resource` system tool — the agent can now load `skill://` / `ui://` resources advertised by an installed bundle's MCP server ([#3](https://github.com/NimbleBrainInc/nimblebrain/pull/25)).
--- a/src/prompt/compose.ts
+++ b/src/prompt/compose.ts
@@ -1,7 +1,81 @@
 import type { ParticipantInfo } from "../conversation/types.ts";
+import { approxTokens } from "../skills/tokens.ts";
 import type { Skill } from "../skills/types.ts";
 
 const SEPARATOR = "\n\n---\n\n";
+
+/**
+ * A single section of the composed system prompt, captured with provenance.
+ *
+ * The traced compose pipeline (`composeSystemPromptTraced`) emits one
+ * `TracedLayer` per section that ends up in the prompt, in the order they
+ * appear. Joining `layers.map(l => l.text)` with `SEPARATOR` reconstructs
+ * the same string `composeSystemPrompt` returns — the trace is non-lossy.
+ *
+ * `subItems` is populated for sections that aggregate multiple operator-
+ * authored entries (apps, layer3 skills). It lets debug tools render per-
+ * item attribution, filter by bundle, and detect content drift on a
+ * per-skill basis without re-parsing the section text.
+ */
+export interface TracedLayer {
+  kind: TracedLayerKind;
+  /**
+   * Stable identifier. Filesystem path for file-backed layers; `nb:<slug>`
+   * for runtime-derived layers; `instructions://<scope>` for overlays.
+   */
+  id: string;
+  /** Human-readable origin (display string for the debug tool's row UI). */
+  source: string;
+  /** The text contribution this layer makes to the composed prompt. */
+  text: string;
+  /** Approximate tokens for `text`. */
+  tokens: number;
+  /**
+   * Bundle attribution, when applicable. For the apps section / focused-app
+   * section / layer3-skills under a bundles/<name>/ subdir. Used by the
+   * compose-effective-context tool's `bundle` filter.
+   */
+  bundle?: string;
+  /**
+   * Per-entry breakdown for sections that aggregate multiple operator-
+   * authored items. Empty / absent for atomic sections.
+   */
+  subItems?: TracedSubItem[];
+}
+
+export type TracedLayerKind =
+  | "default_identity"
+  | "core_skill"
+  | "user_context_skill"
+  | "user_prefs"
+  | "participants"
+  | "workspace_context"
+  | "org_overlay"
+  | "workspace_overlay"
+  | "layer3_skills"
+  | "apps"
+  | "app_state"
+  | "focused_app"
+  | "matched_skill";
+
+export interface TracedSubItem {
+  /** Item kind — finer-grained than the parent layer's kind. */
+  kind: "app" | "layer3_skill";
+  /** Stable identifier — filesystem path for skills; bundle name for apps. */
+  id: string;
+  /** Human-readable display. */
+  source: string;
+  /** Bundle attribution when known. Drives the `bundle` filter. */
+  bundle?: string;
+  /** Free-form metadata appropriate to the kind (skill scope, app trustScore, etc.). */
+  metadata?: Record<string, unknown>;
+}
+
+export interface ComposedPrompt {
+  text: string;
+  layers: TracedLayer[];
+  totalTokens: number;
+}
 
 /**
  * Strip newlines and control characters from single-line fields.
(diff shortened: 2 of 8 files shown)


mgoldsborough:
QA pass — pushed \`f6b3fbd\`.

**W1 (recovered path uncovered):** Fixed. New integration test \`recovers the loaded body from a _versions/ snapshot when the live file has drifted\` plants the snapshot, edits the live file, asserts \`hashStatus="recovered"\` + the snapshot path is reported. The headline historical-mode feature now has a regression test.

**W2 (bundle filter only L3):** Fixed via a new unit-test file \`test/unit/tools/platform/compose-bundle-filter.test.ts\`. Exported \`applyBundleFilter\` + \`ComposeResponse\` and added 5 tests covering: apps section per-subItem narrowing, focused_app \`ownBundleMatches\` branch (which the L3 integration test couldn't reach), layers without bundle attribution dropped, mixed subItems pared, totalTokens recomputed. The same per-subItem code path covers both apps and L3, but now both branches plus the layer-level branch are explicitly asserted.

**S1 (conv-id existsSync before validation):** Fixed. Gated up front with \`CONVERSATION_ID_RE.test(convId)\`. Existing test fixtures updated to use valid 16-hex-char ids — they had been using sentinel strings (\`conv_filter\`, \`conv_ws\`) that never matched the regex.

**S2 (no access check on historical reads):** Filed as #130 (\"compose__effective_context: add ConversationAccessContext check on historical reads\"). The same pattern exists in \`skills.ts::skills__loading_log\` and \`skills__active_for\` — fixing one without the other would be band-aid; the issue tracks the broader audit.

**S3 (deriveBundleFromSkillPath duplicated):** Fixed. Promoted to an export of \`prompt/compose.ts\` and removed the duplicate. Original justification was moot once the parent function went public.

**S4 (called twice per skill):** Fixed with a \`const bundle = …\` cache.

**S5 (POSIX-only path check):** Comment added documenting the deployed-targets assumption (Linux/macOS) and what a future Windows port would need.

**S6 (historical \`text\` misleading):** Fixed with a leading banner — \`[historical mode — layer 3 only. Use live mode for the full composition; see warnings[] for details.]\` — so a programmatic caller reading \`text\` sees the incompleteness up front instead of silently getting an L3 slice.

Tests: 2154 unit (was 2149; +5 bundle-filter unit). 14 compose-tool tests (was 8; +5 unit + 1 integration). \`verify:static\` green.

mgoldsborough:
QA round 2 — pushed \`fe44a33\`.

**Five new findings from this review, all real:**

**(1) Live mode dropped the workspace identity override.** Critical fidelity bug — the trace would tell an operator using \`workspace.identity = "You are LegalBot..."\` that \`DEFAULT_IDENTITY\` was in the prompt. Fixed by exporting \`makeIdentitySkill\` from runtime.ts and replicating the per-workspace override append in \`composeLive\`. Added an integration regression test that plants \`workspace.identity\`, calls the tool, and asserts the override appears as a \`core_skill\` row + \`default_identity\` does not.

**(2) \`hasProxiedTools: false\` hard-coded.** The reviewer was right — my "informational; safe default" comment was wrong. The apps section emits a 3-sentence block based on \`proxied.length > 0\`, common for Tier 2/3 workspaces. Fix: call \`surfaceTools(allTools, null, {})\` and read \`proxied.length\` exactly like \`runtime.chat()\` does.

**(3) Tool list not orgRole-filtered before L3 selection.** Same direction as #2 — non-admin users see more tools than they get, inflating L3 selection. Fix: apply \`isToolVisibleToRole(toolName, identity?.orgRole)\` filter before passing tool names to \`selectLayer3Skills\`.

**(4) \`hashStatus: "no-recorded-hash"\` declared but never produced.** With \`contentHash\` required (last QA round), this branch is unreachable. Dropped from the union — type now describes what's actually emitted.

**(5) Tautological lossless-trace test.** Reviewer is right: \`composeSystemPrompt\` is now a thin wrapper, so the test compared the same value to itself. Replaced with a structural-integrity check (every emitted layer has a known kind, non-empty id/source/text, non-negative tokens). The 62 existing string-variant tests are the real safety net for text reproduction now that they exercise the wrapper.

**Items 6-13 from this review were already addressed in \`f6b3fbd\`** (the previous QA pass — see PR comment trail). The reviewer appears to have run on an older commit. Specifically:
- 6. Recovered hash status test ✓ (\`historical mode > recovers the loaded body from a _versions/ snapshot...\`)
- 7. Bundle filter against apps section ✓ (\`test/unit/tools/platform/compose-bundle-filter.test.ts\` — apps subItem narrowing, ownBundleMatches, mixed subs, totalTokens recompute)
- 8. POSIX-only path comment ✓
- 9. \`deriveBundleFromSkillPath\` deduplicated ✓
- 10. Conv-id regex gate ✓
- 11. ConversationAccessContext on historical reads → tracked as #130 (broader audit)
- 12. Historical \`text\` banner ✓
- 13. \`deriveBundleFromSkillPath\` cached to a const ✓

Tests: 2154 unit pass, 85 tests across the 3 affected files green. \`verify:static\` clean.

mgoldsborough:
QA round 3 — pushed \`b927fe5\`.

**Two important notes for the reviewer up front:**

(a) Your items marked "unchanged from 2nd pass" — items 1, 2, 6, 7, 8 — and items 9, 10 from earlier passes — **are all already fixed.** They were addressed in commit \`fe44a33\` (round 2) and \`f6b3fbd\` (round 1). The current code at \`fe44a33\` has:

- Item 1 (workspace identity override): \`makeIdentitySkill\` imported (compose.ts:50), called (compose.ts:214) + integration regression test in \`compose-effective-context.test.ts\`
- Item 2 (\`hasProxiedTools\`): replaced with \`surfaceTools(allTools, null, {}).proxied.length > 0\` (compose.ts:241)
- Item 6 (orgRole filter): \`isToolVisibleToRole(t.name, identity?.orgRole)\` filter applied at compose.ts:239
- Item 7 (\`no-recorded-hash\`): removed from the type union (compose.ts:414, only 4 statuses now)
- Item 8 (tautological test): replaced with structural-integrity check in prompt.test.ts
- Item 9 (recovered hash status test): \`recovers the loaded body from a _versions/ snapshot...\` in compose-effective-context.test.ts
- Item 10 (bundle filter coverage): \`test/unit/tools/platform/compose-bundle-filter.test.ts\` covers apps subItem, ownBundleMatches, no-attribution drop, mixed subs, totalTokens recompute

The reviewer appears to have run against \`ff68dfd\` (the original PR open) rather than the latest. Force-pushing isn't possible since these are real commits — please pull \`b927fe5\` and re-review.

(b) Items 11-16 (suggestions from the 1st/2nd passes) are also all addressed in \`f6b3fbd\` or tracked as #130 (access-control audit) and #131 (gather-inputs refactor).

**Three genuinely-new findings from this round, all fixed:**

**3 (bundle filter self-inconsistent):** Real bug. After \`applyBundleFilter\`, \`layers\` and \`totalTokens\` were filtered but \`text\` wasn't — a caller using \`totalTokens\` to budget against \`text\` would be misled. Fixed by rebuilding \`text\` from the filtered layer texts. New unit test asserts the joined-text invariant after filtering.

**4 (\`conversation_id\` cosmetic in live mode):** The reviewer is right — the description implied convId drove workspace selection but live mode uses the current request's workspace. Fixed the description to be honest: \"live mode composes against the **calling request's workspace**; convId is echoed as a label and does NOT select the workspace.\" Looking up the conv's workspace would require a global resolver across all workspace dirs (conv ids don't encode workspace) — overkill for v1.

**5 (composeLive docstring inaccurate):** Real. The comment claimed "everything else is workspace-scoped and faithful" but \`getContextSkills()\` is global; the workspace-scoping comes from the identity-override append. Refreshed the docstring to spell out what's global, what's workspace-scoped via the override, what's skipped vs. \`runtime.chat()\`, and why.

**17 (contextAssembled fetched-and-discarded):** Added a comment explaining the narrow read. Historical mode v1's contract is "L3 only" — surfacing the rest of the event's data (per-source breakdown, exclusions) would imply we can reconstruct layers we can't. A future "rich historical mode" PR could expand this; not in scope here.

Tests: 2155 unit pass, 86 tests across the 3 affected files green. \`verify:static\` clean.

PR #132, opened 2026-04-29
fix: surface bundle subprocess stderr in runtime logs (#116)

### LINKED ISSUE #116: Surface bundle subprocess stderr in runtime logs
## Problem

When an MCP bundle subprocess crashes during `bun run dev` (or any runtime invocation), the developer sees only a generic `source.crashed` event in the terminal. The actual cause — the Python traceback, missing-module error, config validation failure, etc. — is invisible.

The bundle is spawned with `stderr: "pipe"` in `src/tools/mcp-source.ts:235`, but **nothing in NB ever reads `transport.stderr`**. Whatever the bundle writes to stderr goes to a pipe with no consumer.

## Real-world cost

A developer recently spent ~2 hours debugging an `mpak bundle run` failure for `@nimblebraininc/synapse-todo-board`. The bundle's Python subprocess was raising `ModuleNotFoundError: No module named 'rpds.rpds'` due to an interpreter version mismatch (mpak issue [NimbleBrainInc/mpak#90](https://github.com/NimbleBrainInc/mpak/issues/90)). The traceback was clear and actionable — but `bun run dev` showed only a generic crash with no payload, so we had to reproduce the failure outside the runtime to even see the error. Every minute of that debug loop was avoidable.

### 1. Drain stderr to the runtime log

In `src/tools/mcp-source.ts` after constructing `StdioClientTransport`, attach a reader on `transport.stderr`:

- Buffer chunks, split into lines.
- Forward each non-empty line via `log.debug(\"mcp\", ...)` (or a new namespace) prefixed with the source name.
- Maintain a bounded ring buffer (last ~50 lines) on the McpSource for use by criterion 2.

### 2. Include captured stderr tail in `run.error` events

Update the `onclose` handler on the stdio transport to include `stderrTail: this.lastStderr.join(\"\\n\")` in the `run.error` event payload. The console event sink (`src/adapters/console-events.ts`) renders the tail when present.

### 3. Make stderr surfacing default-on (not gated by `NB_DEBUG`)

Bundle stderr is signal, not noise — the Python ecosystem uses stderr for warnings, deprecation notices, and tracebacks, all of which are useful in development by default. Format dimmed and prefixed (`[bundle:synapse-todo-board] ...`) to be visually distinct from NB's own output. Keep `NB_DEBUG=mcp` for the additional dispatch tracing already documented in `src/cli/log.ts`.

### 4. (Optional) Startup health check

After `client.initialize()`, if the source crashes within ~2s of spawn, log `bundle exited during startup — last stderr:` with the buffered output before throwing. Catches the most common dev-loop failure (config error, import error, missing dep) and labels it correctly.

## Scope

All changes land in `src/tools/mcp-source.ts` and a small render addition in `src/adapters/console-events.ts`. No protocol changes, no MCP SDK changes. Estimated ~50 lines end-to-end.

## Out of scope

- Persisting bundle logs to disk (separate feature; the runtime work dir already has a `logs/` directory that could host this later).
- Web UI surfacing of `stderrTail` — a follow-up once the data is in the event payload.

mgoldsborough:
Closes #116.

## Summary

- Drain the MCP bundle subprocess's `transport.stderr` into a per-source 50-line ring buffer and a default-on dimmed `[bundle:<name>]` live print. Attached at construction so initialize-time crashes are captured.
- Add the previously missing stdio `onclose` handler so subprocess exits are detected eagerly instead of on the next tool call. Include `stderrTail` in the `source.crashed` payload across all transport modes.
- Render `stderrTail` in `ConsoleEventSink`, dimmed and indented under the error line, so the traceback is visible without digging through separate streams.
- Suppress `source.crashed` on graceful `stop()` via a `stopping` flag — `transport.close()` fires `onclose` synchronously, which would otherwise surface every deliberate teardown as a crash. This was a latent bug across remote/in-process; adding stdio made it loud.

## Why default-on (not `NB_DEBUG`-gated)

Bundle stderr is the bundle author's deliberate diagnostic output — Python uses stderr for tracebacks, warnings, deprecation notices, logs. Different concern than NB's own protocol tracing (`NB_DEBUG=mcp`). The cost asymmetry is severe: hiding signal causes the multi-hour debug loop the issue was filed about; showing dimmed prefixed lines is recoverable with shell redirection or by quieting the bundle. Visual prefix + dim formatting makes the channel tunable by eye without a config flag — same pattern as Docker / `foreman` / `overmind`.

## Architecture

Three concerns kept separate:
1. **Live drain** — bytes-to-lines, attached for the subprocess's lifetime.
2. **Tail capture** — bounded ring buffer (50 lines, 8 KB per line) feeding the crash event payload.
3. **Surfacing** — dev console (default-on) + `run.error.data.stderrTail` (additive field for downstream consumers like a future web UI).

Edge cases handled: partial lines (`print(end="")`), CRLF endings, runaway-line truncation, lossy UTF-8 decode, restart cycles (ring buffer reset at `start()`), graceful stops (suppressed via `stopping` flag).

Reader is attached **at transport construction**, not after `connect()`, because `StdioClientTransport.stderr` is a `PassThrough` exposed synchronously from the constructor — the SDK's own doc-comment calls this out specifically to avoid losing early child output. No race, no monkey-patching of `start()`.

Did NOT introduce a new `bundle.log` engine event for live lines. Out of scope per the issue, and live console output covers the dev-loop case. Web UI surfacing falls out cheaply later by reading `event.data.stderrTail`.

## Test plan

- [x] `bun run verify` passes (2141 unit + 213 web + 416 integration + 17 smoke, exit 0)
- [x] Smoke output shows live `[bundle:echo] ...` lines streaming through during real subprocess runs
- [ ] Manual: trigger a Python `ModuleNotFoundError` in a bundle and confirm the traceback appears under the crash line in `bun run dev`
- [ ] Manual: `bun run dev`, restart a bundle (stop + start) and confirm no spurious `source.crashed` events fire

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -139,6 +139,10 @@ Namespaces (`src/cli/log.ts`):
 
 Add a namespace by calling `log.debug("ns", "message")` (from `src/cli/log.ts`). Keep this table and the `log.ts` doc comment in sync.
 
+### Bundle subprocess stderr (default-on)
+
+Lines a bundle writes to stderr — Python tracebacks, warnings, application logs — are surfaced verbatim and prefixed `[bundle:<sourceName>]`, dimmed. **No flag required.** This is the bundle author's deliberate diagnostic output, separate from NB's own `NB_DEBUG=mcp` tracing; hiding it costs hours when a bundle crashes (issue #116). To quiet a chatty bundle, silence at the bundle level (logger config) or redirect at the shell (`bun run dev 2> >(grep -v '\[bundle:')`). The last 50 lines are also captured into the `source.crashed` event payload as `stderrTail`, so post-mortem consumers see the cause-of-death.
+
 ### Browser (`localStorage.nb_debug`)
 
 ```js
--- a/scripts/repro-issue-116.ts
+++ b/scripts/repro-issue-116.ts
@@ -1,0 +1,167 @@
+#!/usr/bin/env bun
+/**
+ * End-to-end reproducer for issue #116 (bundle subprocess stderr surfacing).
+ *
+ * Spawns a deliberately-broken Python MCP server as a stdio bundle, runs
+ * the production `McpSource` against it, and verifies:
+ *
+ *   1. The subprocess's stderr lines reach the developer (live `[bundle:…]`
+ *      print path).
+ *   2. The death is reported via exactly one `source.crashed` event.
+ *   3. That event's `stderrTail` payload contains the Python traceback.
+ *
+ * Run: `bun run scripts/repro-issue-116.ts`
+ *
+ * Exits non-zero on any contract failure so it can be wired into a smoke
+ * check.
+ */
+
+import { mkdirSync, writeFileSync, rmSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { McpSource, type McpTransportMode } from "../src/tools/mcp-source.ts";
+import type { EngineEvent, EventSink } from "../src/engine/types.ts";
+
+const RESET = "\x1b[0m";
+const GREEN = "\x1b[32m";
+const RED = "\x1b[31m";
+const DIM = "\x1b[2m";
+const BOLD = "\x1b[1m";
+
+function pass(msg: string) {
+  console.log(`${GREEN}✓${RESET} ${msg}`);
+}
+
+function fail(msg: string, detail?: string) {
+  console.log(`${RED}✗${RESET} ${msg}`);
+  if (detail) console.log(`${DIM}    ${detail}${RESET}`);
+  failures++;
+}
+
+let failures = 0;
+
+const bundleDir = join(tmpdir(), `nb-repro-issue-116-${Date.now()}`);
+mkdirSync(bundleDir, { recursive: true });
+
+// Python server that writes some stderr lines, then raises.
+// Mirrors the original bug shape — a real ModuleNotFoundError-style death.
+const SERVER_PY = `
+import sys
+print("[broken-bundle] starting up", file=sys.stderr)
+print("[broken-bundle] reading config", file=sys.stderr)
+print("[broken-bundle] about to import a missing module", file=sys.stderr)
+raise ModuleNotFoundError("No module named 'definitely_not_real'")
+`;
+
+writeFileSync(join(bundleDir, "server.py"), SERVER_PY);
+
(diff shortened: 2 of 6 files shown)


mgoldsborough:
Addressed QA review (commit 9a2b3be).

**Fixed:**
- **Warning 2 (duplicate emit):** Consolidated all four `onclose` handlers and `execute()`'s catch into a single `emitSourceCrashed` helper with `dead`/`stopping` guards. A subprocess death observed by both the transport's `onclose` and an in-flight `execute()`'s catch now fires exactly one `source.crashed`. Whichever path runs first wins; its payload is canonical.
- **Warning 3 (`cleanupOnStartFailure`):** Sets `stopping = true` so a failed-start no longer emits a parallel crash event for a source the listener never saw alive. The OAuth retry path explicitly clears the flag before the second connect, so successful retries re-arm crash detection.
- **Warning 1 (test coverage):** Added 13 unit tests in `test/unit/mcp-source-stderr.test.ts` covering: chunk-boundary line splitting, cross-chunk partial-line buffering, end-of-stream flush, CRLF stripping, ring-buffer FIFO eviction at 50, runaway-line truncation, dead-guard de-dup, payload shape with `stderrTail`, graceful-stop suppression, failed-start suppression, restart re-arms detection, ring-buffer reset across start cycles. Bounded to high-value cases — silent regressions (buffer cap, partial-line flush) and the new lifecycle invariants (dead/stopping guards).
- **Suggestion 1 (CLAUDE.md):** Added a "Bundle subprocess stderr (default-on)" subsection under Debug Logging documenting the `[bundle:<name>]` channel, why it's not gated, and how to quiet a chatty bundle.
- **Suggestion 3 (helper):** Done as part of warning 2 — `emitSourceCrashed` collapses ~50 LoC of duplicated `onclose` blocks. Net negative LOC despite adding the helper.

**Skipped, with reasoning:**
- **Suggestion 2 (`console.error` event-loop blocking):** Premature. Node `PassThrough` applies backpressure — a bundle writing faster than we drain blocks on its own `write()`. The system self-throttles. A bundle that genuinely streams 100 MB to stderr is itself the problem; the fix would belong at that bundle, not in our drain.
- **Suggestion 4 (empty-string `stderrTail`):** Cosmetic. Uniform shape simplifies downstream consumers (no `?.` chains).

**Verify:** `bun run verify` clean — 2154 unit (+13) / 213 web / 416 integration / 17 smoke.

PR #142, opened 2026-04-30
fix: Gemini schema, thinking-block replay, layered config persistence

mgoldsborough:
## Summary

Three related fixes that surfaced during the ipsdi (Gemini-only) and HQ (Anthropic) tenant incidents this week. Each commit stands alone and reviews independently.

### 1. `4313a4a` — LCD-narrow `set_model_config` schema for Gemini compatibility

`set_model_config`'s `thinking` and `thinkingBudgetTokens` properties used JSON-Schema union types with `null` in the enum (`type: ["string","null"]`, `enum: [...,null]`). Gemini rejects `enum` on non-string types, blowing up **every** tool call on Google-only tenants — including the call that would let an operator fix it. Self-locking bug.

Narrowed the schema to LCD-clean shapes (`type: "string"` with a string-only enum, `type: "number"`) and exposed the clear semantic via two new boolean inputs: `clearThinking` and `clearThinkingBudget`. Handler normalizes the booleans into the existing null-sentinel logic; downstream merge code is unchanged.

Also fixes a stale JSDoc in `runtime/types.ts` claiming the platform default for reasoning models is `adaptive` — has been `enabled` since the cost incident at `resolve-thinking.ts:108-119`.

### 2. `dbcdf06` — Preserve content block ordering on history replay

Anthropic validates the latest assistant message byte-for-byte and rejects any modification of thinking blocks. The conversation event reconstructor was splitting each `llm.response` into separate messages by category — one assistant message per (reasoning, tool-call, text) bucket — which hoisted reasoning blocks to the front and reordered tool-calls relative to text. With extended thinking enabled, multi-iteration runs 400'd on the second user message:

```
messages.N.content.M: thinking blocks in the latest assistant message cannot be modified
```

The reconstructor now emits **one** assistant message per `llm.response`, preserving the provider's original block order. Orphaned tool-calls (no matching `tool.done`) are still filtered. Placeholder/role-alternation logic for empty turns and abandoned runs unchanged.

The chat UI consumes its own projection from `src/bundles/conversations/src/jsonl-reader.ts`; nothing here affects display.

This commit also lifts the `providerMetadata`→`providerOptions` and tool-call-input-parse logic out of the engine and the reconstructor into a shared `src/model/inbound-fit.ts` module — single source of truth for provider response normalization, idempotent, used at both the in-memory iteration boundary (engine) and the disk-replay boundary (reconstructor).

New canary test asserts interleaved `[reasoning, tool-call, reasoning, tool-call]` returns in the same order with both signatures preserved.

### 3. `21cfea0` — Layered config: seed + overrides so `set_model_config` survives deploys

The init container overwrites `/data/nimblebrain.json` on every pod start to keep Helm values authoritative. That meant any change made via the UI or chat (`set_model_config`) silently reverted on the next rollout — operator pinned `thinking: off`, deploy ran, runtime came back on the seed default. Bad UX, hard to debug.

Split the file into two:

```
/data/nimblebrain.json            seed, Helm-managed, overwrite on deploy
/data/nimblebrain.overrides.json  user, written by set_model_config,
                                  PVC-resident, untouched by deploys
```

The loader (`cli/config.ts::loadConfig`) reads both and 1-level deep-merges override over seed. Override wins per-key. `set_model_config` writes only to the override file and reads the merged effective state (via `runtime.getRuntimeConfig()`) for cross-field validation. Existing clear-flag semantics work unchanged — they delete keys from the override file, which lets the seed value take over on the next runtime reload.

Startup log line surfaces divergence:
```
[config] Applied N runtime override(s) from <path>: <keys>
```

Bonus fix: `thinking` and `thinkingBudgetTokens` are now propagated from the file config into `RuntimeConfig` (they were silently dropped before). Pre-existing gap, surfaced by the new round-trip test.

New canary test simulates the deploy-replay scenario: operator pins config, runtime restarts, effective config still reflects the override. This is the test that would have caught the original HQ regression.

A companion docs change (`deployments/nimblebrain/CLAUDE.md`) describing the new architecture is on a separate branch in the deployments repo and will follow.

## Test plan

- [x] `bun run verify:static` — clean (1 unrelated pre-existing IPv4 lint warning)
- [x] `bun run verify:test-unit` — 2195 unit + 213 web pass, 0 fail
- [x] `bun run test:integration` — 457 pass, 6 skip, 0 fail
- [x] New canary tests for each commit:
  - `nb__set_model_config schema declares thinking as plain string + boolean clear flags`
  - `preserves original block ordering within an llm.response`
  - `set_model_config writes survive a runtime restart (layered seed + override)`
- [ ] Manual: deploy to ipsdi staging, run `set_model_config { thinking: "off" }` against Gemini, confirm no 400 and value persists
- [ ] Manual: deploy to HQ staging with `thinking: "enabled"`, run a multi-iteration tool-using conversation, confirm no Anthropic 400 on iteration 2+
- [ ] Manual: deploy to either tenant, run `set_model_config`, redeploy, confirm value still in effect

## Commit-by-commit review notes

The three commits are independent and can be reviewed in order. Phase 1 unblocks ipsdi standalone; Phases 2 and 3 fix HQ standalone; Phase 4 prevents the regression class going forward.

### DIFF
--- a/src/cli/config.ts
+++ b/src/cli/config.ts
@@ -18,8 +18,23 @@ export interface CliFlags {
   defaultWorkDir?: string;
 }
 
-/** Validate config file contents using JSON Schema. Throws on structural errors, warns on unknown keys. */
-function validateConfig(config: Record<string, unknown>, path: string): void {
+/**
+ * Validate config file contents using JSON Schema. Throws on structural
+ * errors. Warns on unknown keys when `warnUnknownKeys` is true (default).
+ *
+ * The override file is written by `set_model_config` and contains a known
+ * small surface — its valid keys (e.g., `thinking`, `thinkingBudgetTokens`)
+ * are not yet in the published JSON schema, so unknown-key warnings would
+ * fire on every boot for any tenant that's run `set_model_config`. The
+ * structural validation (type errors) still fires; only the key-name
+ * warning is suppressed for that file.
+ */
+function validateConfig(
+  config: Record<string, unknown>,
+  path: string,
+  opts: { warnUnknownKeys?: boolean } = {},
+): void {
+  const warnUnknownKeys = opts.warnUnknownKeys ?? true;
   const validate = getValidator();
   const valid = validate(config);
 
@@ -38,8 +53,10 @@ function validateConfig(config: Record<string, unknown>, path: string): void {
       }
     }
 
-    for (const key of warnings) {
-      console.error(`[config] Warning: unknown key "${key}" in ${path} (ignored)`);
+    if (warnUnknownKeys) {
+      for (const key of warnings) {
+        console.error(`[config] Warning: unknown key "${key}" in ${path} (ignored)`);
+      }
     }
 
     if (errors.length > 0) {
@@ -95,7 +112,11 @@ export function loadConfig(flags: CliFlags = {}): RuntimeConfig {
     try {
       const overrideRaw = readFileSync(configOverridePath, "utf-8");
       const override = JSON.parse(overrideRaw) as Record<string, unknown>;
-      validateConfig(override, configOverridePath);
+      // Suppress unknown-key warnings: the override file's vocabulary
+      // (thinking, thinkingBudgetTokens) is not yet in the published
+      // JSON schema, and warning every boot for every tenant that's
+      // run set_model_config is noise. Structural errors still throw.
+      validateConfig(override, configOverridePath, { warnUnknownKeys: false });
       const overrideKeys = Object.keys(override);
       if (overrideKeys.length > 0) {
         fileConfig = mergeConfigs(seedConfig, override) as Partial<RuntimeConfig> &
--- a/src/cli/config.ts
+++ b/src/cli/config.ts
@@ -1,6 +1,7 @@
 import { existsSync, readFileSync, writeFileSync } from "node:fs";
 import { dirname, join, resolve } from "node:path";
 import { getValidator } from "../config/index.ts";
+import { deriveOverridePath, mergeConfigs } from "../config/overrides.ts";
 import type { RuntimeConfig } from "../runtime/types.ts";
 
 const DEFAULT_CONFIG_FILE = "nimblebrain.json";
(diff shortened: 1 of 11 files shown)


mgoldsborough:
## QA adjudication — `9eb1d2c`

Verified each item against the code and acted as follows:

### Warnings

**1. `clearThinking + thinkingBudgetTokens` orphan budget — fixed.**
Real bug, reproducible via trace at `core-source.ts:430-442`. With `clearThinking: true` setting `input.thinking = null`, line 430 deletes both fields, then line 441 re-sets the budget from the patch. Disk gets `{thinkingBudgetTokens: 4096}` with no mode. Live runtime stays consistent because the handler passes both as null to `updateConfig` when `input.thinking === null` — surfaces only at next restart.

Fix: added a third mutex check alongside the two existing ones. Symmetric `clearThinkingBudget + thinking` is **not** needed — I traced all combinations; the existing `thinking="enabled" requires budget` validator covers the only divergent case. New regression test in `test/integration/core-source.test.ts` reproduces the bug and asserts the new error.

**2. Schema unknown-key warnings on every boot — mitigated.**
Confirmed the published schema at `schemas.nimblebrain.ai/v1/` doesn't include `thinking` or `thinkingBudgetTokens`. Pre-PR these warnings fired too (when `set_model_config` wrote to the seed) but they were rare because the init container kept wiping the seed. The new "Applied N runtime overrides" log line makes them newly visible.

Fix: added a `warnUnknownKeys` flag to `validateConfig`; the override-file validation passes `false`. Structural type errors still throw loudly. Proper fix is to update the JSON schema in `products/nimblebrain/schemas/` — separate PR.

**3. Migration for tenants whose `thinking` is in the seed — skipped, no real loss.**
Pre-PR `set_model_config` wrote to the seed via `getConfigPath`. The init container has been overwriting the seed on every deploy with the ConfigMap content (which has no `thinking` field) — meaning any pre-PR `thinking` value in the seed has *already* been wiped by the most recent rollout. There is no in-flight `thinking` value sitting in any tenant's seed file at the moment this PR ships. A copy-on-first-load migration would copy zero bytes. Mentioning here so the next reviewer doesn't ask the same question.

### Suggestions

**S1. Avoid mutating `input` — deferred, documented.**
Started a `const patch = { ...input }` refactor but reverted: the handler reads `input.X` in ~20 places after the normalize block, and a wholesale `input → patch` rename in this commit muddies the QA-fix diff. The mutation is safe today (no shared callers, per-request input). Added a code comment noting the assumption and when to revisit. Worth a follow-up commit on its own.

**S2. Placeholder drops unsigned reasoning — documented.**
Confirmed the behavior change. The new behavior is *more* correct: the AI SDK Anthropic provider silently drops unsigned blocks on replay with an `unsupported reasoning metadata` warning, so the reconstructed message previously promised content that wasn't actually sent. Filtering up-front is honest. Added a comment in `event-reconstructor.ts` explaining the change and why.

**S3. Concurrent `set_model_config` race — documented.**
Confirmed there's no lock between read-existing and write-new. Atomic write via temp+rename, but two parallel calls can lose updates. Admin-only and rare. Added a comment on the tool entry; if it becomes a real concern, gate writes on a per-path mutex (`async-mutex` etc.).

### Verification

- `bun run verify:test-unit` — 2196 unit + 213 web pass, 0 fail
- `bun run test:integration` — 458 pass, 6 skip, 0 fail (one new test added for Warning 1)

PR #143, opened 2026-04-30
fix(model): route bare gemini/openai ids to the right provider

mgoldsborough:
## Summary

Fixes a routing bug that 404'd any non-Anthropic model selected via the settings UI:

```
[api] AI_APICallError: model: gemini-3.1-pro-preview
       url: https://api.anthropic.com/v1/messages
       responseBody: {"type":"error","error":{"type":"not_found_error","message":"model: gemini-3.1-pro-preview"}}
```

The settings model-picker wrote `defaultModel: "gemini-3.1-pro-preview"` to disk (bare id, no `google:` prefix) because the dropdown option used `m.id` alone, ignoring the `provider` already in scope at `<optgroup>`. The runtime resolver at `src/model/registry.ts:resolveModelString` then defaulted bare ids to `anthropic:`, sending the Gemini model name to `api.anthropic.com`.

## Two-part fix

**1. UI — encode provider into option value (`web/src/pages/settings/ModelTab.tsx`)**
```tsx
const qualified = `${provider}:${m.id}`;
return <option key={qualified} value={qualified}>...</option>;
```
Stops the bug at the source. New saves write `google:gemini-3.1-pro-preview` etc.

**2. Runtime — catalog-driven fallback in `resolveModelString`**
Bare ids now look up the catalog for an owning provider before falling back to `anthropic:`. Rescues bare ids that older UI versions already wrote to disk — tenants don't need to re-save to fix routing.

```ts
export function resolveModelString(model: string): string {
  if (model.includes(":")) return model;
  const catalogProvider = findProviderForModelId(model);
  if (catalogProvider) return `${catalogProvider}:${model}`;
  return `anthropic:${model}`;  // backward compat for bespoke ids
}
```

Added `findProviderForModelId(modelId)` to `src/model/catalog.ts` — scans the catalog and returns the first provider that declares the id (catalog ids are unique across providers in practice).

## Test plan

- [x] `bun run verify:static` — clean
- [x] `bun run verify:test-unit` — 213 web pass, 0 fail
- [x] `bun test test/unit/model-registry.test.ts` — 13 pass, includes new regressions:
  - bare gemini id → google
  - bare openai id → openai
  - bare anthropic id → anthropic (existing behavior preserved)
  - bare unknown id → anthropic fallback (backward compat preserved)
- [ ] Manual: deploy to ipsdi/HQ, select Gemini model in settings, send a message, confirm routes to api.google.com (or whatever Gemini endpoint) without 404
- [ ] Manual: any tenant with a previously-saved bare gemini id should now work without re-saving

## Note on the UI dropdown after this lands

If a tenant has a bare id (like `gemini-3.1-pro-preview`) saved from before this PR, the dropdown will show "Select a model" as the placeholder because option values are now qualified — the bare id won't match any option. The model still works at runtime (catalog fallback), but the visual indicator is off until they re-select. Worth a small follow-up to normalize bare ids to qualified ones on load if the UX warrants it.


mgoldsborough:
## QA adjudication — `c2c2151`

### Warnings

**1. `findProviderForModelId` iteration-order dependency — fixed (with stronger mitigation).**
Took the runtime-check route over the test-only mitigation. Tests don't run in production; a duplicate that slips past CI silently changes routing. Better still: built a precomputed `Map<id, provider>` at module load — surfaces duplicates as a `console.warn`, makes the lookup O(1) instead of O(providers). First-seen wins (matches prior behavior); not fatal because duplicates are a data hygiene issue, not a correctness one. No throws at boot.

**2. Legacy bare ids show empty dropdown — fixed in this PR.**
Reviewer is right that this would have generated tickets. ~5 lines in the load `useEffect`: walk `availableModels` (already in hand from `get_config`), prepend the matching provider when the saved id is bare. Side benefit — when the user saves without changing anything, the qualified value gets written back, quietly migrating the persisted state.

### Suggestions

**1. Direct unit tests for `findProviderForModelId` — added.**
Five cases in `test/unit/catalog.test.ts`: known anthropic/google/openai ids, unknown id, empty string. Direct coverage so a future catalog refactor that breaks the helper fails on the actual broken thing instead of through `resolveModelString` indirection.

**2. Sync-time invariant in `sync-models.ts` — skipped, redundant.**
The runtime duplicate-id warning from #1 covers the same risk less invasively. If a duplicate ever lands, every boot warns about it, the team investigates. Adding a second gate at the sync boundary expands this PR's scope into the sync subsystem (separate concerns, separate tests). Defer until/unless the runtime warning actually fires.

### Verification

- `bun run verify:test-unit` — 213 web pass, 0 fail
- `bun test test/unit/catalog.test.ts test/unit/model-registry.test.ts` — 38 pass (5 new for `findProviderForModelId`)

mgoldsborough:
## Round-2 QA adjudication — `88f78a6`

### Item 1 — Resolver-only normalization is incomplete (downstream pipelines see bare ids)

**Agreed strongly, fixed.** Verified the reviewer's full trace — every line is real:

- `resolve-thinking.ts:103` — bare gemini → `getModelByString` → undefined → no thinking
- `resolve-max-output-tokens.ts:27` — undefined → static default instead of catalog limit
- `engine.ts:74` — `getProviderFromModel` returns `"anthropic"` for bare gemini → wrong-shape `providerOptions` on Google calls
- `usage-aggregator.ts:92` — undefined → **$0 cost reports for the entire session** (the bad one)
- `llm.response` events log the bare string

The reviewer's framing is the architecturally correct one: normalize at the request-entry boundary, not at the resolver boundary. One-line fix at `runtime.ts` after the alias-resolution block calls `resolveModelString` so the qualified form propagates to every downstream consumer. Updated PR description claim ("tenants don't need to re-save") is now genuinely true.

### Item 2 — Misleading test descriptions in `multi-model.test.ts:268-287`

**Agreed, deleted.** The block was duplicate AND mis-describing — `test/unit/model-registry.test.ts` is the authoritative location with semantically-correct cases (gemini→google, openai→openai, anthropic→anthropic, unknown→anthropic fallback). Replaced with a one-line breadcrumb pointing at the new home.

Also removed the now-unused `resolveModelString` import from the file.

### Item 3 — `findProviderForModelId` iteration-order dependency

**Already addressed in `c2c2151` (round-1 fixes).** That commit replaced the linear scan with a precomputed `Map<id, provider>` built at module load and warns on duplicates via `console.warn`. O(1) lookup, visible-but-not-fatal divergence detection. Reviewer may have been looking at `fb2f08b` only — the runtime check is exactly the option (b) they suggested.

I declined the additional sync-models invariant test (their option (a)) in the round-1 adjudication as redundant, and stand by it: the runtime warn fires on every boot; a sync-time test only fires on PRs to the catalog. Defense in depth fine, but the runtime check is sufficient for the failure mode.

### New regression test

`test/integration/runtime/model-qualification.test.ts` asserts:
1. Bare `gemini-3.1-pro-preview` arriving as `request.model` produces `run.start.data.model === "google:gemini-3.1-pro-preview"`.
2. Already-qualified ids pass through unchanged.

Locks in the rescue so a future refactor that drops the request-boundary `resolveModelString` call fails with a meaningful test signal instead of silent $0 cost reports.

### Verification

- `bun run verify:test-unit` — 213 web pass, 0 fail
- `bun run test:integration` — 460 pass, 6 skip, 0 fail (2 new in `model-qualification.test.ts`)

mgoldsborough:
## Round-3 QA adjudication — `fb8649d`

**Heads up: three of four warnings appear to be re-raises against `fb2f08b` (the original commit). Items already on the branch:**
- W1's *trace* (resolver-only catalog fallback) was addressed in `88f78a6` with a `resolveModelString` call at `runtime.ts:783-787` (request-entry boundary). The line numbers in your trace match the pre-fix state.
- W2 (misleading `multi-model.test.ts` block) was deleted in `88f78a6` with a breadcrumb pointing at `model-registry.test.ts`.
- W3 (`findProviderForModelId` iteration order) was addressed in `c2c2151` — precomputed `Map<id, provider>` at module load with `console.warn` on duplicates. O(1) lookup, visible-but-not-fatal.
- S1 (direct unit tests for `findProviderForModelId`) was added in `c2c2151` (5 cases in `test/unit/catalog.test.ts`).

So this round's actionable items are W1's *alternative fix-point proposal* (centralize at `getModelSlots()` instead of only at request entry) and W4 (CHANGELOG entry).

### W1 alternative — fix-point at `getModelSlots()` ✅ **Adopted, layered**

The reviewer's proposal is genuinely better architecture than my round-2 fix: it covers `get_config` (which feeds the dropdown), telemetry, and any future consumer that reads slots — all paths I'd have had to remember to qualify individually.

Adopted in addition to the request-entry call, not as a replacement. Belt-and-suspenders:

- `getModelSlots()` qualifies all three slots and workspace overrides → covers `engineConfig.model`, `get_config` (so the UI dropdown auto-matches), telemetry, briefing, etc.
- Request-entry `resolveModelString` from 88f78a6 still qualifies `request.model` (the per-request override path bypasses `getModelSlots`).
- Catalog fallback in `resolveModelString` itself (fb2f08b) stays as a defensive layer for any direct caller.

The two are orthogonal: slot reader handles "what's the configured default", request-entry handles "what did the API call ask for." Removing either creates a hole.

I considered removing the catalog fallback in `resolveModelString` (option (c) the reviewer noted) since all reads now come pre-qualified — but kept it as cheap defense in depth. If a future code path bypasses both qualification points, the utility self-corrects. Decision: keep.

### W4 — CHANGELOG entry ✅ **Added**

One line under `[Unreleased] / Fixed` describing the production-visible 404, the dual fix (UI-side prefix + runtime-side rescue), and that existing tenant configs continue routing without re-saving. Linked to PR #143.

### Test impact

The slot-reader change made `getRuntimeConfig().defaultModel` return the qualified form for tenants whose override file has bare ids. Updated `set_model_config writes survive a runtime restart` to expect that form (with a comment explaining: write-side stays bare, read-side qualifies via the slot reader).

New regression in `model-qualification.test.ts`: feeds bare ids for all three providers via `Runtime.start({models: {...}})`, asserts `getModelSlots()` returns each properly qualified.

### Verification

- `bun run verify:test-unit` — 213 web pass, 0 fail
- `bun run test:integration` — 461 pass, 6 skip, 0 fail (1 updated assertion + 1 new slot-reader test)

The reviewer noted a "defensible counter-framing" — that the metadata gap (cost, capabilities, limits) for bare gemini predates this PR and is arguably a separate bug. We chose not to ship narrow: the metadata gap is fixed in the same PR via the slot-reader normalization, because $0 cost reports for an entire session is exactly the kind of silent failure that hides for weeks. This PR closes it now.

PR #145, opened 2026-04-30
fix(inbound-fit): preserve Gemini thoughtSignature on tool-call replay

mgoldsborough:
## Summary

Gemini 2.5+/3.x signs tool calls with a `thoughtSignature` (its analogue of Anthropic's thinking-block `signature`). After PR #142 generalized the inbound-fit module, the `providerMetadata → providerOptions` rename was applied **only** to reasoning blocks — but Google attaches `thoughtSignature` to tool-call, text, file, AND reasoning parts. Tool-call signatures got dropped on round-trip, and Gemini 400'd:

```
Function call is missing a thought_signature in functionCall parts.
Additional data, function call \`default_api:nb__status\`, position 6.
```

Generalizes the rename to apply uniformly across content types. A type-discriminated rename was the original gap; uniform application is the defensive shape.

## What changed

- `src/model/inbound-fit.ts` — `normalizeForReplay` now lifts `providerMetadata → providerOptions` for every content type that carries metadata, not just reasoning. Plus the existing tool-call input parsing.
- Return type changed from `LanguageModelV3Content[]` (stream-side, no `providerOptions`) to a new exported `ReplayContent[]` (prompt-side `LanguageModelV3*Part` union, has `providerOptions`). Stream→prompt is what the function actually does; typing it that way removes the cast cascade an earlier draft needed (no `as unknown as` anywhere).
- Tool-result / source / tool-approval-request: explicitly skipped (filtered) instead of falling through unchanged. Tool-results in this codebase are built by the runtime as `role: "tool"` messages, not pulled from `doStream`; source and approval-request are stream-side concepts that never appear in assistant prompt content. Documented in code.
- New `test/unit/inbound-fit.test.ts` with 10 cases: tool-call input parsing, Anthropic signature on reasoning, **Gemini thoughtSignature on tool-call (the regression case)**, Gemini thoughtSignature on text, parts without metadata pass through, idempotency, no input mutation, both transformations on a single tool-call.
- `AGENTS.md` convention: prefer typed-safe paths over `as unknown as T`; cast escape hatches require a naming-the-mismatch comment. Plus a one-line meta-note that `AGENTS.md` edits should be terse and token-aware (audience is agents, not humans).

## Test plan

- [x] `bun run verify:static` — clean
- [x] `bun run verify:test-unit` — 213 web + unit pass, 0 fail
- [x] `bun run test:integration` — 461 pass, 6 skip, 0 fail
- [x] New regression test for the Gemini thoughtSignature path
- [ ] Manual: ipsdi tenant — Gemini chat with a tool call (e.g., \`whats your model?\` → triggers \`nb__status\`), confirm no \`Function call is missing a thought_signature\` error

## Note on the AGENTS.md change

This PR's refactor pass demonstrated the rule. An earlier draft used `as unknown as LanguageModelV3Content` to paper over a stream-vs-prompt-side type mismatch. The fix was to type the function as stream→prompt, not to widen with `unknown`. Codifying the lesson plus a pointer to the worked example so the next agent on the codebase finds the typed path first.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,7 @@
 # NimbleBrain
 
+> This file is read by agents. Keep edits terse, imperative, token-aware. No long-form prose; bullets with concrete triggers and examples.
+
 Self-hosted platform for MCP Apps and agent automations, built on Bun. Agentic loop + MCP bundle management + interactive UI host + cron-scheduled automations + skill-driven prompt composition + HTTP API + web client.
 
 ## Build & Verify
@@ -36,6 +38,7 @@ bun run build:bundles      # Rebuild every src/bundles/*/ui (vite single-file)
 - **Module system:** ESM only. All imports use `.ts` extensions.
 - **Linting:** Biome (not ESLint/Prettier). Run `bun run lint`.
 - **Type checking:** `bunx tsc --noEmit`. Strict mode enabled.
+- **Prefer typed-safe paths over `as unknown as T`.** When TS errors, find the input/output type matching runtime shape (e.g. stream-side vs prompt-side) before widening. Cast escape hatches require a comment naming the mismatch. Example: `src/model/inbound-fit.ts`.
 - **HTTP framework:** Hono for routing and middleware. Typed context via `AppEnv`/`AuthEnv`.
 - **Model types:** Use Vercel AI SDK V3 types (`LanguageModelV3`, `LanguageModelV3Message`, etc.) from `@ai-sdk/provider`. The engine calls `model.doStream()` directly.
 - **No classes for data** — plain interfaces + factory functions preferred.
--- a/src/conversation/event-reconstructor.ts
+++ b/src/conversation/event-reconstructor.ts
@@ -6,7 +6,7 @@
  * builder, and the web client.
  */
 
-import type { LanguageModelV3Content } from "@ai-sdk/provider";
+import type { LanguageModelV3Content, LanguageModelV3ReasoningPart } from "@ai-sdk/provider";
 import { estimateCost } from "../engine/cost.ts";
 import { normalizeForReplay } from "../model/inbound-fit.ts";
 import type { ConversationEvent, LlmResponseEvent, StoredMessage, ToolDoneEvent } from "./types.ts";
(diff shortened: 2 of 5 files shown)


mgoldsborough:
## Self-critique + QA round-2 follow-ups — \`66b341b\`

Six small items in one commit, no behavior change to the public function. Type-check survives every cast removal — proves they were already redundant.

### From self-critique (the gaps I called out earlier)

- **Stale call-site casts dropped.** The prior commit's type narrowing made these load-free; this commit removes them and lets tsc verify:
  - \`engine.ts:439\` — \`as LanguageModelV3Message\`
  - \`event-reconstructor.ts:282, 353\` — \`as StoredMessage\` (replaced with \`const x: StoredMessage = {...}\` so the shape still validates)
- **Test for the tool-result filter behavior.** Asserts that \`tool-result\`, \`source\`, and \`tool-approval-request\` parts are dropped from the output. Pins the intentional filter so a future "simplification" can't silently restore the wrong-shape pass-through.
- **\`ReplayContent\` unexported.** Zero external consumers (verified by grep); keeping it exported would create an unintended contract.

### From QA round 2

- **Q1 — useless \`continue\` on tool-result branch:** dropped (Biome info-level lint).
- **Q2 — narrowing predicate mistyped:** tightened to \`c is LanguageModelV3ReasoningPart\` in \`event-reconstructor.ts:343\`. Removes the \`as { providerOptions?: unknown }\` cast on the next line because the prompt-side type already declares the field. Reviewer's note that "the AGENTS.md rule the PR introduces would itself recommend the cleanup" is exactly right.
- **Q3 — test style:** \`expect(...).not.toHaveProperty("providerOptions")\` instead of \`'providerOptions' in out[0]!\`.

One thing I left as-is per the reviewer's own note: \`event-reconstructor.ts:236\` still has \`(c): c is LanguageModelV3Content & { type: "tool-call" }\`. Same predicate-vs-actual-type smell, but the input there is \`llmResp.content\` (which IS \`LanguageModelV3Content[]\`, the stream-side type stored on the event) — not \`replayContent\`. So the predicate matches the input type correctly. Different code path; not the same issue.

### Verification

- \`bun run check\` — clean (3 casts removed, no regressions)
- \`bun run verify:static\` — clean
- \`bun run verify:test-unit\` — 213 web pass + 11 inbound-fit unit (1 new for the filter)
- \`bun run test:integration\` — 461 pass, 6 skip, 0 fail

PR #152, opened 2026-04-30
fix(engine): promote tools discovered by nb__search into the next iteration

dom-mp:
## Summary
- promote exact tool names discovered through `nb__search(scope="tools")` into the callable tool set for the next iteration of the same engine run
- add regression coverage for the discovery-to-execution flow
## Problem
In workspaces with enough tools to trigger tiered surfacing, app tools can be placed in the proxied/discoverable bucket instead of the direct/callable bucket.
In that state, the model can:
1. call `nb__search`
2. discover an app tool name such as `newsapi__get_top_headlines`
but it could not actually call that discovered tool in the next step, because the discovered tool was never promoted into the callable tool list for the run.
This was especially visible with OpenAI-backed agent runs, where the model would report that the tool existed but was not directly callable in the current tool set.
## Root Cause
The proxied-tool workflow had a missing bridge between:
1. discovery by name through `nb__search(scope="tools")`
2. execution of the discovered tool in a subsequent iteration
The engine preserved the original direct tool list across the run, so discovered proxied tools never became callable.
## Fix
- track tools discovered from successful `nb__search(scope="tools")` results during an engine run
- parse exact discovered tool names from the search result text
- promote those discovered tools into the direct callable tool set for the next iteration
- keep promotion scoped to the current run only
- do not change runtime tool surfacing policy globally
## Why This Approach
This keeps NimbleBrain's tiered surfacing design intact and fixes the actual gap in the proxied-tool flow, instead of broadening tool exposure for a whole provider.
It avoids provider-specific behavior and avoids exposing all tools directly just to work around the issue.
## Testing
Added regression coverage for:
- promoting a discovered tool after `nb__search`
- making that tool callable on the next iteration
- preserving normal behavior when no discovery occurs
Focused checks run:
```bash
bun test test/unit/engine.test.ts test/unit/tool-surfacing.test.ts
bun run verify
```
## Reproduction
One concrete reproduction was:
- workspace contains enough tools to exceed DEFAULT_MAX_DIRECT_TOOLS
- @nimblebraininc/newsapi is installed and healthy
- the agent asks for technology headlines in the US
- the model discovers newsapi__get_top_headlines via nb__search
- before this fix, the discovered tool remained non-callable for the run
- after this fix, the discovered tool becomes callable on the next iteration

### DIFF
--- a/src/engine/engine.ts
+++ b/src/engine/engine.ts
@@ -225,6 +225,17 @@ function addCacheBreakpoint(messages: LanguageModelV3Message[]): LanguageModelV3
   return result;
 }
 
+const DISCOVERED_TOOL_NAME_RE = /^- \*\*([a-zA-Z0-9_-]+__[a-zA-Z0-9_-]+)\*\*:/gm;
+
+function extractDiscoveredToolNames(text: string): string[] {
+  const names = new Set<string>();
+  for (const match of text.matchAll(DISCOVERED_TOOL_NAME_RE)) {
+    const name = match[1];
+    if (name) names.add(name);
+  }
+  return [...names];
+}
+
 export class AgentEngine {
   constructor(
     private model: LanguageModelV3,
@@ -258,19 +269,13 @@ export class AgentEngine {
       if (t.annotations) toolAnnotations.set(t.name, t.annotations);
     }
 
-    // Translate ToolSchema[] to LanguageModelV3FunctionTool[] for the model call
-    const modelTools: LanguageModelV3FunctionTool[] = tools.map((t) => ({
-      type: "function" as const,
-      name: t.name,
-      description: t.description,
-      inputSchema: t.inputSchema as JSONSchema7,
-    }));
-
-    // Build tool schema lookup for input validation (once per run, not per iteration)
-    const toolSchemaMap = new Map<string, ToolSchema>();
-    for (const t of tools) {
-      toolSchemaMap.set(t.name, t);
+    const allToolSchemaMap = new Map<string, ToolSchema>();
+    for (const t of allRouterTools) {
+      allToolSchemaMap.set(t.name, t);
     }
+
+    const directTools = [...tools];
+    const directToolNames = new Set(directTools.map((t) => t.name));
 
     this.events.emit({
       type: "run.start",
@@ -330,6 +335,18 @@ export class AgentEngine {
 
     try {
       while (iteration < maxIter) {
+        const modelTools: LanguageModelV3FunctionTool[] = directTools.map((t) => ({
+          type: "function" as const,
+          name: t.name,
+          description: t.description,
+          inputSchema: t.inputSchema as JSONSchema7,
+        }));
+
+        const toolSchemaMap = new Map<string, ToolSchema>();
+        for (const t of directTools) {
+          toolSchemaMap.set(t.name, t);
+        }
+
         // 1. Apply context/prompt hooks and call LLM
         const windowed = config.hooks?.transformContext
           ? config.hooks.transformContext([...history])
@@ -459,6 +476,11 @@ export class AgentEngine {
             if (gatedCall === null) {
               return {
                 toolCall,
+                gatedCall: {
+                  id: toolCall.toolCallId,
+                  name: toolCall.toolName,
+                  input: parsedInput,
+                },
                 result: {
                   content: textContent("Tool call was denied by policy."),
                   isError: true,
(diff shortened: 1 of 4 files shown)


mgoldsborough:
`src/tools/system-tools.ts:147` currently returns every tool, including ones annotated `ai.nimblebrain/internal`. This PR makes those callable to the LLM, breaking the contract documented at `in-process-app.ts:24-28`.

This is a fix you'd want regardless of #152 — the search tool was already leaking internal tools into its results, the PR just makes the leak visible.

```
  // src/tools/system-tools.ts, around line 145-149                                                                                         
  const all = (await getRegistry().availableTools()).filter(
    (t) => !t.annotations?.["ai.nimblebrain/internal"],                                                                                     
  );                                                                                                                                        
  if (!q) return groupToolsBySource(all);                                                                                                   
  const matches = all.filter(                                                                                                               
    (t) => t.name.toLowerCase().includes(q) || t.description.toLowerCase().includes(q),
  );                                                                                                                                        
```
                                                             
One line. Fixes the engine concern as a side effect — internal tools never appear in bullets, so they can't be promoted.

mgoldsborough:
# Replace regex parsing with structured output 

The current regex (engine.ts:226) reverse-engineers structured intent out of human prose that nb__search already had structured (system-tools.ts:147). This is the change that does the most work for the least code:
                                                                                                                                            
  In system-tools.ts, return tool names alongside the prose:                                                                                
  
```
  const matches = all.filter(...);                                                                                                          
  if (matches.length === 0)                                       
    return { content: textContent(`No tools matched "${query}".`), isError: false };                                                        
                                                                                                                                            
  const lines = [`Found ${matches.length} tool(s) for "${query}":\n`];                                                                      
  for (const t of matches) lines.push(`- **${t.name}**: ${t.description}`);                                                                 
                                                                                                                                            
  return {                                                        
    content: textContent(lines.join("\n")),                                                                                                 
    structuredContent: { tools: matches.map((t) => ({ name: t.name })) },
    isError: false,                                                                                                                         
  };
```
                                                                                                                                            
In engine.ts, delete the regex (engine.ts:226-235) and the markdown-extraction call. Read names from the tool result's structuredContent instead:

```                                                                                              
  if (                                                            
    gatedCall.name === "nb__search" &&
    !result.isError &&
    gatedCall.input.scope === "tools"                                                                                                       
  ) {
    const structured = result.structuredContent as { tools?: { name: string }[] } | undefined;                                              
    for (const { name } of structured?.tools ?? []) {                                                                                       
      if (directToolNames.has(name)) continue;
      const schema = allToolSchemaMap.get(name);                                                                                            
      if (!schema) continue;                                      
      if (schema.annotations?.["ai.nimblebrain/internal"]) continue; // belt + suspenders w/ fix #1                                         
      directTools.push(schema);                                                                                                             
      directToolNames.add(name);                                                                                                            
    }                                                                                                                                       
  }
```

Why this matters:                                               
- Removes the format-coupling failure mode entirely.
- Aligns with the platform CLAUDE.md §1.3 ("Storage-symmetric shape" — structured data is the source of truth, prose is presentation).    
- Sets up Phase 2 (nb__use) trivially — that handler will read the same structuredContent the LLM saw.                                
- The test still works as written; just update the mock nb__search handler to also return structuredContent.

mgoldsborough (changes requested):
Comments below. Please review and take a look.

mgoldsborough:
@dom-mp Once this lands, the architectural follow-up is #158: replace the search-driven auto-promotion with explicit `nb__use` / `nb__release` tools  so the LLM controls its toolset directly, demotion becomes possible, and the engine stops special-casing nb__search by name. This PR is the right tactical fix to ship now; #158 is the next step and better long term archtecture.

dom-mp:
- `nb__search(scope="tools")` now filters out tools annotated with `ai.nimblebrain/internal`, so internal-only tools no longer appear in search results
- `nb__search` now returns discovered tool names in `structuredContent.tools` alongside the existing human-readable `content`
- removed the regex/prose parsing path from `engine.ts`
- the engine now promotes discovered tools from `result.structuredContent` instead of scraping names out of markdown
- added a defensive internal-tool check in the engine promotion path (`schema.annotations?.["ai.nimblebrain/internal"]`)
- updated the engine test so the mocked `nb__search` result returns `structuredContent` as well
Verified with:
```bash
bun test test/unit/engine.test.ts test/unit/system-tools.test.ts test/unit/tool-surfacing.test.ts
bun run verify
```

mgoldsborough (approved):

Issue #155, opened 2026-04-30
New chats do not appear in Conversations list until refresh or tab switch


dom-mp:
## Summary
When a user creates a brand new chat while the Conversations tab is open, the new conversation is not shown in the Conversations list immediately.
The conversation appears later if the user:
1. Refreshes the browser
2. Switches away from Conversations and then back
This looks like a live-update bug rather than intentional behavior.
## Steps To Reproduce
1. Open the app
2. Go to the Conversations tab
3. Start a brand new chat/conversation
4. Observe the Conversations list
## Expected Behavior
The newly created conversation should appear in the Conversations list immediately, without requiring a browser refresh or tab switch.
## Actual Behavior
The new conversation does not appear in the Conversations list right away.
It only shows up after:
- refreshing the browser, or
- navigating away and back to Conversations
## Impact
This makes the Conversations tab feel stale and inconsistent, especially when users are actively managing or reviewing conversations while chatting.

mgoldsborough:
Resolved on `main`. `runtime.chat` emits `data.changed{server:"conversations"}` on a new conversation (`emitConversationsChanged`), and the Conversations dashboard refreshes on it — the new chat row appears without a refresh or tab switch.

PR #161, opened 2026-05-02
fix(api): bump MCP session TTL default to 8h + log session-miss

mgoldsborough:
## Summary

The 30-minute idle TTL on MCP sessions was sweeping connector sessions mid-workday — every sweep returns 404 to the next request, which surfaces as a "timeout" on the client. This is the immediate fix.

- **TTL default 30 min → 8 h.** Sessions are evicted on idle, not absolute age, so an actively-used connection survives indefinitely (each request bumps `lastAccessedAt`).
- **`MCP_SESSION_TTL_MS` env var renamed to `MCP_SESSION_TTL_SECONDS`** *(breaking)*. Operator-facing knobs should use the unit operators think in — nobody sets a sub-second timeout, and `28800000` is hostile to read when the actual value is 8 hours. Seconds also matches every adjacent abstraction (K8s timeouts, ALB `idle_timeout.timeout_seconds`, ALB `stickiness.lb_cookie.duration_seconds`, Redis EXPIRE). Internal sweep math stays in ms; conversion happens once at the env-parse boundary.
- **`parsePositiveIntEnv` guard** rejects non-positive and non-integer values. The previous `parseInt(env, 10)` left two failure modes open: `8h` parsed to `8` (an 8-second TTL that evicted every session on the next sweep), and `foo` parsed to `NaN` (silently disabled eviction). Both surfaced as runaway capacity-cap 429s.
- **Structured logging at the three session-confusion paths** so this class of issue is diagnoseable without `NB_DEBUG`:
  - POST `/mcp` with unknown session id → `warn`
  - POST `/mcp` non-init without session id → `warn`
  - DELETE `/mcp` with unknown session id → `info`

Each log line carries `sessionId=<8-char prefix> workspace=<id> identity=<id> ip=<x-forwarded-for>` for cross-tenant correlation.

This is the unblock-today fix. The architectural work (pluggable `SessionRegistry`, Redis provider, Helm `RollingUpdate` + ALB stickiness for horizontal scale) is in NimbleBrainInc/nimblebrain#162.

## Breaking changes

`MCP_SESSION_TTL_MS` → `MCP_SESSION_TTL_SECONDS`. **Any existing deployment setting the old env var will silently fall through to the new 8h default.** Mention in release notes; update operator docs (handled in the docs PR).

## Test plan

- [x] `bun run verify` (unit + integration + smoke; 2274+ tests).
- [x] New unit tests cover `parsePositiveIntEnv` for 7 input shapes (unset/empty/valid/non-numeric/zero/negative/decimal). Each rejection case asserts the warning fires.
- [x] New integration tests for the three log branches assert both status codes AND structured log fragments (`sessionId=`, `workspace=`, `identity=`, `ip=`) via a `captureLogs()` spy. The earlier QA review caught that workspace context wasn't reaching the DELETE path; the test now asserts it does.
- [x] Manual smoke test against running dev server.
- [x] Bootstrap endpoint loads cleanly in Chrome.

### DIFF
--- a/src/api/mcp-server.ts
+++ b/src/api/mcp-server.ts
@@ -56,9 +56,26 @@ const MCP_SERVER_VERSION = process.env.NB_VERSION || mcpPkg.version;
  * day never expires mid-use. Sessions are evicted on idle, not absolute age,
  * so an actively-used connection survives indefinitely (each request bumps
  * `lastAccessedAt`). Override via `MCP_SESSION_TTL_MS` for tighter limits.
+ *
+ * `parsePositiveIntEnv` rejects `NaN` and non-positive values: a typo like
+ * `MCP_SESSION_TTL_MS=8h` would otherwise leave `SESSION_TTL_MS=NaN`, which
+ * silently disables eviction (NaN comparisons are always false) and the
+ * capacity cap eventually 429s every new client.
  */
-const MAX_MCP_SESSIONS = parseInt(process.env.MCP_MAX_SESSIONS ?? "100", 10);
-const SESSION_TTL_MS = parseInt(process.env.MCP_SESSION_TTL_MS ?? String(8 * 60 * 60 * 1000), 10);
+const MAX_MCP_SESSIONS = parsePositiveIntEnv("MCP_MAX_SESSIONS", 100);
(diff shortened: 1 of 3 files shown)


mgoldsborough:
fixes #155

PR #166, opened 2026-05-04
feat(connectors): unified Connectors UX — registries + permissions + operator setup

mgoldsborough:
This PR has grown well beyond its original Tracks A+B+C scope. The
final shape is a coherent unification of every install-and-manage
surface in NimbleBrain into a single primitive — **the connector** —
backed by a registry abstraction, per-tool permissions, and an in-
product operator-setup flow.

## The conceptual model

A **connector** is anything that exposes tools or UI surfaces to the
agent: local stdio bundles, local URL bundles, Synapse apps, remote
OAuth services. They flow through one list, one Configure surface,
one install/uninstall path. Differences (transport, auth, UI) are
implementation details surfaced as badges, not separate primitives.

### Registries

`src/registries/` — pluggable sources of installable connectors.

- `RegistryStore` — instance-level config at `<workDir>/registries.json`.
  Auto-seeds on first read with `curated` (locked) and `mpak` (default
  on, URL-overridable). Atomic writes.
- `ConnectorRegistry` interface + `DirectoryEntry` (uniform shape) +
  `InstallAction` discriminated union (`remote-oauth` | `mpak-bundle`
  | `direct-url`).
- `CuratedRegistry` wraps the in-process catalog. Computes
  `operatorConfigured` per-call via a `ListEntriesContext` resolver —
  workspace-aware seam without coupling registries to the runtime
  singleton.
- `MpakRegistry` is a stub returning a small hand-curated set of
  well-known mpak bundles. Real mpak.dev API client + bundle install
  are separate-PR follow-up.
- `DirectoryAggregator` builds the active registry list, isolates
  per-registry failures into `errors[]` so one bad source doesn't
  blank Browse.

### Permissions

`src/permissions/` — per-tool policy enforcement.

- `PermissionStore` — file-backed at
  `users/<userId>/permissions.json` and
  `workspaces/<wsId>/permissions.json`. Default-allow when no entry
  exists. Atomic writes.
- `ToolRegistry.execute()` consults the store before dispatching.
  `UserPoolSource` calls key on the calling principal; everything
  else keys on workspace. Disallowed tools return a structured
  `tool_permission_denied` error.

### Operator OAuth app setup

`workspace.json#oauthOperatorApps[catalogId]` carries the public
`client_id` plus an audit trail (configuredAt, configuredBy). The
matching `client_secret` lives in the workspace credential store at
`operatorSetup.clientSecretKey`. Independent of bundle lifecycle —
uninstall doesn't touch it. New `manage_connectors` actions:
`setup_operator` (admin-gated upsert) + `remove_operator_setup`
(refuses while installed).

### Settings → Workspace → Connectors (`/settings/workspace/connectors`)

Flat list. One row per installed bundle: icon, name, status nudge
only when something needs attention (reauth / failed). Whole row
links to Configure. Search appears at >5 items. Top-right `Browse`
button.

### Browse (`.../connectors/browse`)

Aggregated directory across all enabled registries. Curated +
mpak.dev entries render side-by-side with the same row template;
install dispatch follows the entry's `install` discriminator.
Per-row `Curated` / `mpak.dev` badge for attribution. Already-
installed entries demote to bottom of the list. Static-auth
entries:
- not configured + admin → `Set up` button → modal
- not configured + non-admin → `Operator setup required` pill
- configured + admin → `Install` + `Edit OAuth app` link
- configured + non-admin → `Install`

### OperatorSetupModal

Same component for first-setup and rotate. Portal link, two fields,
clientId pre-fillable (secret never echoes). Esc/click-outside closes.

### Configure (`.../connectors/:serverName`)

Centered, no card chrome. Top-right action bar: optional Docs link
+ `Uninstall` (text-style destructive). Header: icon + name +
metadata (type, scope, state, version, connected-as). `Reconnect`
inline when reauth needed. `ToolPermissionsTable` below: per-tool
Allow / Disallow toggles, optimistic with rollback, bulk `Allow
all` / `Disallow all`. MCP descriptions trimmed to first sentence
with full text in row's `title` tooltip.

### Settings → Organization → Registries

Admin-only. Each registry has an enable/disable toggle (curated is
locked-on); mpak has an editable URL field for private mpak
instances.

## Notable bug fixes along the way

- **Uninstall stale state**: `lifecycle.uninstall` cleared its own
  `instances` map and the legacy global `nimblebrain.json` but
  didn't touch `workspace.json#bundles[]`. Reinstall after uninstall
  failed silently then 404'd at OAuth initiate. Fixed at the connector-
  tools layer + self-heal in `handleInstall`.
- **OAuth callback path**: hardcoded `/settings/workspace/connections`
  (renamed to `/connectors`) caused a blank page after authorize.
  Callback now redirects to the workspace connectors tab.
- **list_tools source resolution**: was reading from `Connection.source`
  which is only populated on the user-flow OAuth path — boot-restored
  bundles have `source: null` even when running. Now reads from the
  workspace registry, which is the authoritative location.

## Personal Connectors UI parked

`/settings/personal/connectors` was empty for nearly every user (no
catalog entries land in user scope today; the user-scope services
require operator setup the v1 doesn't surface). Removed from the nav
+ routes. Backend pathways (`UserConnectorStore`, lifecycle user-
scope methods, `manage_connectors scope: \"user\"`) stay intact —
the abstraction is unbroken; only the UI surface goes. Routes
redirect to `workspace`.

## About tab

\"Installed Bundles\" table removed. Connectors page is now the
canonical surface; About just shows platform version + a link.

## Test coverage

- 4 new unit-test files (38 tests):
  - `permission-store.test.ts` (8) — default-allow, set/get round-
    trip, scope isolation, path-traversal defense
  - `registry-store.test.ts` (7) — seed defaults, persist across
    instances, locked semantics, auto-restore curated
  - `directory-aggregator.test.ts` (7) — aggregation, attribution,
    dedupe, context plumbing
  - `connector-tools.test.ts` (17) — setup_operator, remove_operator_
    setup, list_directory, static-auth install path

- Total: **2480 unit + 466 integration + 17 smoke** all passing.
- `bun run verify` clean.

## What's deferred

- **mpak.dev real API integration**: `MpakRegistry` is a stub.
  Surface is right; the API client + `kind: \"mpak-bundle\"` install
  flow are net-new work. Best as a separate PR.
- **Personal connectors UI**: parked. Backend ready when there's a
  reason to surface it.
- **Integration test for install → policy → enforce**: the constituent
  pieces have unit coverage; an end-to-end MCP-protocol test would
  add value but isn't a prerequisite for ship.

## Post-review additions (QA pass)

After QA review, three follow-up commits landed:

- **fix(connectors): protocol validation triple** — extends iconUrl's allowlist to operatorSetup.portalUrl, enforces http(s) on `manage_registries.set_url`, and validates the OAuth callback's returnUrl scheme. Defense-in-depth across operator-controlled URL surfaces.
- **chore(connectors): dead code + stale comments sweep** — removes `hasPersistedMemberOAuthTokens` (orphaned by user-scope rename), `getPendingAuthUrl` (no callers post-route-removal), refines comments referencing `/v1/connections/*` and the old `member` vocab, swaps PermissionStore's substring slicing for `dirname()`, canonicalizes MpakRegistry URLs via `new URL(...).origin`.
- **fix(connectors): test backstop + browse dedupe + revert resolveWithCode shape** — new `registry-permission-enforcement.test.ts` (5 tests) proves the `ToolRegistry.execute` permission gate wires correctly. ConnectorBrowsePage dedupes mount + post-modal-save fetches into one `fetchDirectory`. Reverts `resolveWithCode` to boolean since nothing consumes the metadata today; cleanly removes the `void matched` smell.

Final test count: **2483 unit + 466 integration + 17 smoke**. `bun run verify` clean.

## Filed as follow-up

- [#182 — OAuth startAuth: thread AbortSignal through redirect-probe fetches](https://github.com/NimbleBrainInc/nimblebrain/issues/182). 1–2 hours of care; not a security fix; tracked separately.


mgoldsborough:
**Rebased onto main + retargeted.** Step 1 (PR #175) is now in main, so this PR no longer carries those 7 commits.

Branch was rebased with \`git rebase --onto main 15ce2a0 feat/connections-tracks\` — replays the 12 net-new Tracks A/B/C commits onto current main, dropping the 7 Step 1 commits as duplicates.

One small follow-up commit on top: \`56d5842\` updates the \`POST /v1/mcp-auth/initiate\` 404 unit test to assert against the new \`bundle_not_found\` error code and extends \`StubLifecycle\` with the \`getInstance()\` method this PR's route handler now requires. Originally PR #175 tested only the legacy "no pending flow" 404 path; Track B added the bundle-existence check up front, so the test needed to follow.

Local verification on the rebased branch:
- \`bun run verify:static\` — clean (1 pre-existing url-validator warning)
- \`bun run test:unit\` — 2432 pass, 0 fail (1 pre-existing unhandled async error from JSZip parsing in some upstream test, doesn't fail any test — to be tracked separately)
- \`bun run test:integration\` — 456 pass, 12 skip, 0 fail
- \`bun run smoke\` — 17 pass, 0 fail

PR #168, opened 2026-05-04
feat: bundle upgrade flow with update UI

### LINKED ISSUE #35: Bundle upgrade flow: no  action, no version pinning, no reconcile
## Problem

NimbleBrain's bundle lifecycle supports `install`, `uninstall`, `start`, `stop` — but **no `upgrade`**. Once a registry-based bundle (`{name: "@nimblebraininc/foo"}`) is installed in a workspace, there's no mechanism that pulls a newer version from mpak. The workspace keeps running whatever was fetched on initial install, indefinitely.

Today the only way to get a newer version is:

```
nb__manage action=uninstall name=@nimblebraininc/foo
nb__manage action=install name=@nimblebraininc/foo
```

This works but has real downsides (see below). More importantly, it's not discoverable — users don't know "upgrade = uninstall + install" is the incantation.

## Why this matters

Concrete scenario from this week:

- We cut synapse-crm v0.2.0 and synapse-todo-board v0.2.0 to fix a user-visible tool bug (agents orphaning tasks into undefined columns) and migrate to a new tool contract (flat kwargs).
- For every tenant / local install currently running v0.1.0 of these bundles, **NimbleBrain doesn't know the newer version exists**. There's no "updates available" signal, no reconcile, no prompt.
- The tenant keeps running a known-bad version until someone manually intervenes with the uninstall/install dance.

Multiply by every bundle in the registry and every tenant running them. The answer to "how do I upgrade?" shouldn't require reading the lifecycle source.

## Downsides of uninstall+install as the upgrade path

1. **Not atomic**: between uninstall and install, the bundle is unavailable. Any agent turn in that window fails.
2. **Loses installed state**: if configure-style credentials or env allowlists live alongside the bundle entry, uninstall wipes them.
3. **Not discoverable**: neither the agent nor a human user has an obvious way to express "move to latest."
4. **No rollback**: if the new version is broken, getting back requires another uninstall + install pinned to an older version — and we don't support version pinning either (see below).
5. **No "check for updates"**: you can't ask NimbleBrain "is there a newer version than what I have?" without manually inspecting the mpak registry.

## Proposed

Three independent-ish pieces. Ship in any order; each is useful alone.

### 1. `nb__manage action=upgrade` action

```
nb__manage action=upgrade name=@nimblebraininc/foo
```

Behavior:
- Read the currently installed version from bundle metadata
- Query mpak for the latest version (or the latest matching a pin — see #2)
- If newer: fetch new bundle, swap in place, preserve workspace-level config (`allowedEnv`, credentials refs, etc.)
- If same or older: no-op with a clear message
- Atomic from the agent's perspective — old tools remain callable until the new bundle is ready, then switch

Implementation notes:
- `BundleLifecycleManager` gets an `upgrade(name, wsId, registry)` method that composes the existing `installNamed` + diff logic
- Keep `uninstall + install` working as a fallback; `upgrade` is additive

### 2. Version pinning in `workspace.json`

Today:
```json
{ "name": "@nimblebraininc/foo" }
```

Proposed (all valid, backward-compat):
```json
{ "name": "@nimblebraininc/foo" }                           // implicit latest (current behavior)
{ "name": "@nimblebraininc/foo", "version": "latest" }      // explicit latest
{ "name": "@nimblebraininc/foo", "version": "^0.2" }        // semver range
{ "name": "@nimblebraininc/foo", "version": "0.2.1" }       // exact pin
```

On install / upgrade, resolve the spec against mpak's published versions. Matches npm / pip conventions; no surprises for consumers used to those.

Opens the door to:
- `nb__manage action=upgrade` honors the pin (won't cross a major boundary unless you widen the range)
- Future `nb__manage action=rollback` to previous installed version

### 3. Optional: "updates available" surface

Not required but cheap to add once #1 is in:

- A read-only tool `nb__updates_available` (or part of existing `nb__search` output) that lists installed bundles whose mpak `latest` exceeds their installed version and respects any pin
- Surfaces in a dashboard widget or as a notification
- No automatic action — just visibility

Deliberately **no** automatic background reconcile / auto-upgrade. Upgrades should be user-triggered. Too easy to break a workspace with a silent auto-update at the wrong moment.

## Out of scope for this issue

- **Path-based bundles** (`{path: ...}`) — these are dev-mode references and intentionally have no upgrade concept. Changes are picked up on bundle-process restart. The upgrade flow here is "pull + restart the bundle server", handled at the git layer.
- **Remote-URL bundles** (`{url: ...}`) — could support versioned URLs but treating them as always-latest is probably correct.
- **Reconcile loop / auto-upgrade** — explicitly not proposed. Manual trigger only.

## Related

- Built on top of the existing `BundleLifecycleManager` in `src/bundles/lifecycle.ts`
- Informs the agent-driven install flow in `src/tools/system-tools.ts` (where `nb__manage` is wired)
- synapse-crm v0.2.0 and synapse-todo-board v0.2.0 (just cut) are the concrete bundles that today have no path to reach existing installs

## Severity

Medium-to-high. Every bundle update beyond v0.1.0 hits this gap. As the mpak ecosystem grows, the pain scales linearly.

Ovaculos:
## Summary

- Add `upgrade()` method to `BundleLifecycleManager` — best-effort hot-swap that stops the running source, re-installs from mpak, and starts the new version. Preserves workspace data and credentials.
- Extend `manage_app` tool with `"upgrade"` action and add new `check_updates` tool that queries the registry for available updates
- Add Update column to the About page — shows version diff per bundle, Update button gated to `org_admin` role, "updating" status badge during upgrade

Closes #35 (pieces 1 and 3 — upgrade action and updates-available surface. Piece 2, version pinning, deferred to follow-up PR)

## Review feedback addressed

- **JSDoc "atomic swap" → "best-effort hot-swap"** — docs now honestly describe the remove-then-start sequence. No atomicity claims.
- **Stale automations after upgrade** — `removeBundleAutomations()` called before `syncBundleAutomations()`, matching the uninstall pattern. Prevents orphaned schedules when a new manifest drops a previously-declared automation.
- **UI swallowing upgrade failures** — `result.isError` now checked explicitly before calling `fetchApps()`. Error message extracted and routed to `setUpgradeErrors` so the inline error renders.
- **`check_updates` clobbering local-installed bundles** — added `installSource` field (`"registry" | "local" | "remote"`) to `BundleInstance` and `AppInfo`. `check_updates` and the frontend filter on `installSource === "registry"` instead of the `bundleName.startsWith("@")` heuristic. Local dev bundles with scoped manifest names are no longer offered for update.
- **Duplicated `NB_WORK_DIR` resolution** — extracted to `resolveWorkDir()` function (not a module-level constant — constants capture env at import time, breaking integration tests that override `process.env.NB_WORK_DIR` at runtime).
- **trustScore stale after upgrade** — `fetchTrustScore()` called during upgrade to refresh from mpak registry.
- **Placement unregister safety** — unconditionally calls `unregister()` before conditionally re-registering. `unregister()` is a no-op filter when no entries exist.
- **No `protected` check on upgrade** — intentional. Protected bundles can't be uninstalled but should receive version updates (security patches). Added code comment explaining the decision.
- **Unit tests expanded** — 13 tests covering upgrade action guards, schema validation, `check_updates` filtering by `installSource`, upgrade no-op when already at latest, and unknown instance error path.

## Test plan

- [x] 13 unit tests for upgrade flow (action guards, schema, check_updates filtering, no-op, error paths)
- [x] 2356 unit tests pass, 455 integration tests pass
- [x] Lint and format clean
- [x] Manual testing: About page shows Update column only for registry bundles, upgrade works end-to-end
- [ ] Test with bundle that has a newer version available on mpak
- [ ] Test role gating: non-admin user should see version info but no Update button

### DIFF
--- a/src/bundles/lifecycle.ts
+++ b/src/bundles/lifecycle.ts
@@ -22,6 +22,11 @@ import type {
   RemoteTransportConfig,
 } from "./types.ts";
 
+/** Resolve work dir at call time — env var may be overridden in tests. */
+function resolveWorkDir(): string {
+  return process.env.NB_WORK_DIR ?? join(homedir(), ".nimblebrain");
+}
+
 // ---------------------------------------------------------------------------
 // BundleLifecycleManager — owns the state of all installed bundles and
 // provides the formal install / uninstall / start / stop / restart flows
@@ -120,7 +125,7 @@ export class BundleLifecycleManager {
     // Workspace-scoped data dir keeps two workspaces installing the same
     // bundle from stomping on each other's entity data. Matches the
     // seedInstance layout used at platform boot.
-    const nbWorkDir = process.env.NB_WORK_DIR ?? join(homedir(), ".nimblebrain");
+    const nbWorkDir = resolveWorkDir();
     const bundleDataDir = join(nbWorkDir, "workspaces", wsId, "data", deriveBundleDataDir(name));
 
     const { sourceName, manifest } = await startBundleSource(
@@ -139,6 +144,7 @@ export class BundleLifecycleManager {
     const isUpjack = manifest._meta?.["ai.nimblebrain/upjack"] != null;
     const instance = createInstance(sourceName, name, manifest, isUpjack, wsId);
     instance.configKey = name;
+    instance.installSource = "registry";
     this.transition(instance, "running");
 
     instance.trustScore = await fetchTrustScore(name, this.mpakHome);
@@ -199,6 +205,7 @@ export class BundleLifecycleManager {
     // Use manifest.name (scoped name) as bundleName, not the filesystem path.
     const instance = createInstance(sourceName, manifest.name, manifest, isUpjack, wsId);
     instance.configKey = bundlePath; // config entry uses the filesystem path
+    instance.installSource = "local";
     this.transition(instance, "running");
 
     instance.ui = extractUiMeta(manifest);
@@ -248,7 +255,7 @@ export class BundleLifecycleManager {
     // via `installRemote` from any workspace would share OAuth tokens across
     // workspaces under the default id — silent cross-tenant credential
     // leakage.
-    const nbWorkDir = process.env.NB_WORK_DIR ?? join(homedir(), ".nimblebrain");
+    const nbWorkDir = resolveWorkDir();
     const { sourceName, meta } = await startBundleSource(
       { url, serverName, transport: transportConfig, ui: ui ?? null },
       registry,
@@ -269,6 +276,7 @@ export class BundleLifecycleManager {
       protected: false,
       type: "plain",
       wsId,
+      installSource: "remote",
     };
     this.transition(instance, "running");
 
@@ -367,7 +375,7 @@ export class BundleLifecycleManager {
     // Credentials are config, not data — they should not persist across
     // uninstalls. Data directories are preserved (step 6).
     if (instance) {
-      const workDir = process.env.NB_WORK_DIR ?? join(homedir(), ".nimblebrain");
+      const workDir = resolveWorkDir();
       try {
         await clearAllWorkspaceCredentials(instance.wsId, instance.bundleName, workDir);
       } catch (err) {
@@ -383,6 +391,104 @@ export class BundleLifecycleManager {
       type: "bundle.uninstalled",
       data: { serverName, bundleName: nameOrPath, wsId },
     });
+  }
+
+  // ---- Upgrade -----------------------------------------------------------
+
+  /**
+   * Upgrade a named bundle to the latest version from the mpak registry.
+   *
+   * Best-effort hot-swap: tears down the old process, then starts the new
+   * version. Brief unavailability window during the swap (sub-second,
+   * same pattern as `configureBundle`). If the new version fails to start
+   * the bundle is left unregistered until a reinstall or restart.
+   *
+   * No `protected` guard — protected bundles can be upgraded (but not
+   * uninstalled) so they aren't frozen out of security updates.
+   *
+   * Preserves: workspace-scoped data dir, credentials, config entry.
+   * Emits: bundle.upgraded event on success.
+   */
+  async upgrade(
+    name: string,
+    wsId: string,
+    registry: ToolRegistry,
+  ): Promise<{ from: string; to: string; serverName: string }> {
+    const serverName = deriveServerName(name);
+    const instance = this.instances.get(`${serverName}|${wsId}`);
+    if (!instance) {
+      throw new Error(`No bundle instance found for "${name}" in workspace "${wsId}"`);
+    }
+
+    const mpak = getMpak(this.mpakHome);
+    const fromVersion = instance.version;
+
+    // Check if a newer version exists
+    const latest = await mpak.bundleCache.checkForUpdate(name, { force: true });
+    if (!latest) {
+      return { from: fromVersion, to: fromVersion, serverName };
+    }
+
+    // Fetch the new artifact into cache
+    await mpak.bundleCache.loadBundle(name, { force: true });
+
+    // Resolve workspace-scoped paths (same as installNamed)
+    const nbWorkDir = resolveWorkDir();
+    const bundleDataDir = join(nbWorkDir, "workspaces", wsId, "data", deriveBundleDataDir(name));
+
+    // Remove old source, then spawn new process. The registry throws on
+    // duplicate names so we can't hold both simultaneously. The window is
+    // sub-second (process spawn time) — same pattern as configureBundle.
+    if (registry.hasSource(serverName)) {
+      await registry.removeSource(serverName);
+    }
+
+    const { sourceName: newSourceName, manifest } = await startBundleSource(
+      { name },
+      registry,
+      this.eventSink,
+      this.configPath ? dirname(this.configPath) : undefined,
+      { dataDir: bundleDataDir, wsId, workDir: nbWorkDir },
+    );
+    if (!manifest) {
+      throw new Error(`No manifest found for ${name} after upgrade fetch`);
+    }
+
+    // Update instance metadata in place
+    const isUpjack = manifest._meta?.["ai.nimblebrain/upjack"] != null;
+    instance.version = manifest.version;
+    instance.description = manifest.description;
+    instance.type = isUpjack ? "upjack" : "plain";
+    instance.ui = extractUiMeta(manifest);
+    instance.briefing = extractBriefing(manifest);
+    instance.trustScore = await fetchTrustScore(name, this.mpakHome);
+    this.transition(instance, "running");
+
+    // Always unregister stale placements, then re-register if new manifest
+    // declares any. Without unconditional unregister, a version that drops
+    // all placements would leave stale nav entries.
+    this.placementRegistry?.unregister(serverName, wsId);
+    if (instance.ui?.placements) {
+      this.placementRegistry?.register(newSourceName, instance.ui.placements, wsId);
+    }
+
+    // Clean stale automations then sync from new manifest. Without the
+    // remove, schedules dropped between versions keep running with stale prompts.
+    await this.removeBundleAutomations(name, registry);
+    await this.syncBundleAutomations(manifest, name, registry);
+
+    this.eventSink.emit({
+      type: "bundle.upgraded",
+      data: {
+        serverName,
+        bundleName: name,
+        fromVersion,
+        toVersion: manifest.version,
+        wsId,
+      },
+    });
+
+    return { from: fromVersion, to: manifest.version, serverName };
   }
 
   // ---- Start / Stop / Restart -------------------------------------------
@@ -640,12 +746,16 @@ export class BundleLifecycleManager {
         ? join(dataDir, manifestMeta.upjackNamespace, "data")
         : undefined;
 
+    const installSource: BundleInstance["installSource"] =
+      "name" in ref ? "registry" : "url" in ref ? "remote" : "local";
+
     const instance: BundleInstance = {
       serverName,
       // Prefer the scoped manifest name over the config label (filesystem path)
       bundleName: manifestMeta?.manifestName ?? bundleName,
       // Config key for reliable uninstall — the original value from nimblebrain.json
       configKey: bundleName,
+      installSource,
       version: manifestMeta?.version ?? "unknown",
       description: manifestMeta?.description,
       state: "running",
--- a/src/tools/system-tools.ts
+++ b/src/tools/system-tools.ts
@@ -163,14 +163,15 @@ export async function createSystemTools(
     {
       name: "manage_app",
       description:
-        "Install, uninstall, or configure an app. 'configure' prompts for API keys/credentials securely via the terminal. Requires user approval.",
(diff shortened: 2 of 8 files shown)


mgoldsborough (commented):
Three critical issues from full QA review (warnings + suggestions in separate review summary). Happy to draft fixes for any of these.

src/bundles/lifecycle.ts
```diff
@@ -385,6 +385,95 @@ export class BundleLifecycleManager {
...
+  /**
+   * Upgrade a named bundle to the latest version from the mpak registry.
+   *
+   * Atomic swap: spawns the new version first, only tears down the old
+   * process after the new one is healthy. If the new version fails to
+   * start, the old process is left untouched.
+   *
+   * Preserves: workspace-scoped data dir, credentials, config entry.
+   * Emits: bundle.upgraded event on success.
+   */
```
mgoldsborough: **JSDoc claims atomic swap, but the code is not atomic.**

This docstring says: *"Atomic swap: spawns the new version first, only tears down the old process after the new one is healthy. If the new version fails to start, the old process is left untouched."*

The actual sequence at L428-440 calls `registry.removeSource(serverName)` **before** `startBundleSource`. If `loadBundle` or `startBundleSource` throws (network failure, manifest validation, spawn error), the user is left with no running source — the opposite of "left untouched." The PR description repeats the same false claim.

The inline comment at L427 ("Remove old source, then spawn new process") is correct; this JSDoc above it contradicts it.

**Fix:** pick one — either rewrite to truly atomic (start new under a temp serverName, swap on success), or update this docstring + the PR description to reflect the actual best-effort hot-swap.

src/bundles/lifecycle.ts
```diff
@@ -385,6 +385,95 @@ export class BundleLifecycleManager {
...
+    this.transition(instance, "running");
+
+    // Re-register placements with new metadata
+    if (this.placementRegistry && instance.ui?.placements) {
+      this.placementRegistry.unregister(serverName, wsId);
+      this.placementRegistry.register(newSourceName, instance.ui.placements, wsId);
+    }
+
+    // Sync automations from new manifest
+    await this.syncBundleAutomations(manifest, name, registry);
```
mgoldsborough, line 478: **Stale bundle-contributed automations after upgrade.**

`syncBundleAutomations` is idempotent on `automationId` (see L625 — `createAutomation` returns existing if the id matches) and never removes automations that existed for the prior version but were dropped in the new manifest.

If v1 declares schedules `["a", "b"]` and v2 declares only `["a"]`, schedule `b` keeps running after upgrade with a stale prompt, until the bundle is fully uninstalled.

**Fix:** call `removeBundleAutomations(name, registry)` before `syncBundleAutomations`, or compute a name-set diff and delete the obsolete ones.

web/src/pages/settings/AboutTab.tsx
```diff
@@ -72,6 +98,37 @@ export function AboutTab() {
...
+      setUpgrading((prev) => new Set(prev).add(bundleName));
+      setUpgradeErrors((prev) => {
+        const next = new Map(prev);
+        next.delete(bundleName);
+        return next;
+      });
+      try {
+        await callTool("nb", "manage_app", { action: "upgrade", name: bundleName });
+        await fetchApps();
+      } catch (err) {
```
mgoldsborough, line 129: **UI silently swallows upgrade failures.**

`callTool` returns the raw `ToolCallResult` and does **not** throw on `isError: true` — only `parseToolResult` throws (see `web/src/api/tool-result.ts:18`).

When the upgrade tool returns `{ isError: true, content: "Failed to upgrade …" }`, the `await` resolves normally, the `catch` is never hit, and `fetchApps()` runs. The user sees the same version with no error indication.

**Fix:** check `result.isError` (or call `parseToolResult` and let it throw) and route to `setUpgradeErrors` so the destructive-text span at L226 actually renders the failure reason.

src/tools/system-tools.ts
```diff
@@ -893,6 +898,119 @@ async function uninstallBundleFromWorkspaceViaCtx(
...
+      if (!wsId) {
+        return {
+          content: textContent("Workspace context required for update checking."),
+          isError: true,
+        };
+      }
+
+      const instances = lifecycle.getInstances().filter((i) => i.wsId === wsId);
+      // Only check named bundles (scoped names start with @)
+      const named = instances.filter((i) => i.bundleName.startsWith("@"));
```
mgoldsborough: **`check_updates` can offer to clobber local-installed bundles.**

`installLocal` sets `instance.bundleName = manifest.name` (`src/bundles/lifecycle.ts:200`), which can be a scoped name like `@org/foo`. Such an instance is a local dev bundle (its `configKey` is the disk path, not a scoped name), but this filter treats it as a registry bundle.

`check_updates` will fetch from mpak; if a registry version exists with the same scoped name, the UI offers an Update button that — when clicked — pulls the registry version and replaces the local dev bundle.

**Fix:** filter on `configKey?.startsWith("@")` (which is set per install path: scoped name for `installNamed`, filesystem path for `installLocal`, URL for `installRemote`), or add an explicit `installSource` field to `BundleInstance`.

mgoldsborough (commented):
Follow-up review covering four additional issues (one warning, three suggestions) flagged in the QA pass.

test/unit/bundle-upgrade.test.ts
```diff
@@ -0,0 +1,227 @@
...
+  it("upgrade action is listed in tool schema", async () => {
+    const registry = await makeRegistry();
+    const tools = await createSystemTools(() => registry);
+    const allTools = await tools.tools();
+    const manageApp = allTools.find((t) => t.name === "nb__manage_app");
+    expect(manageApp).toBeDefined();
+    const schema = manageApp!.inputSchema as { properties: { action: { enum: string[] } } };
+    expect(schema.properties.action.enum).toContain("upgrade");
+  });
+});
```
mgoldsborough, line 156: **No unit test for the upgrade success path.**

All three tests in this `describe` block cover error/dispatch surface (missing lifecycle, missing manageBundleCtx, schema enum). None exercise `BundleLifecycleManager.upgrade()` end-to-end:

- instance metadata updated (`version`, `description`, `type`, `ui`, `briefing`)
- `bundle.upgraded` event emitted with correct `fromVersion` / `toVersion`
- registry source replaced (old `serverName` removed, new one registered)
- placement registry re-registered

The PR description's test plan flags this as a manual test. Worth a deterministic unit test that fakes a v0.1.0 → v0.2.0 cache transition and asserts on instance state + emitted events. Would also catch the stale-trustScore and stale-automations bugs flagged elsewhere in this review.

src/bundles/lifecycle.ts
```diff
@@ -385,6 +385,95 @@ export class BundleLifecycleManager {
...
+    const latest = await mpak.bundleCache.checkForUpdate(name, { force: true });
+    if (!latest) {
+      return { from: fromVersion, to: fromVersion, serverName };
+    }
+
+    // Fetch the new artifact into cache
+    await mpak.bundleCache.loadBundle(name, { force: true });
+
+    // Resolve workspace-scoped paths (same as installNamed)
+    const nbWorkDir = process.env.NB_WORK_DIR ?? join(homedir(), ".nimblebrain");
```
mgoldsborough: **Duplicated `NB_WORK_DIR` resolution.**

`process.env.NB_WORK_DIR ?? join(homedir(), ".nimblebrain")` appears at L123, L251, L370, and now L424. Worth extracting a private helper (or a module-level constant) so the default path is defined in exactly one place. Drift here would be silent — different code paths could resolve different work dirs.

src/bundles/lifecycle.ts
```diff
@@ -385,6 +385,95 @@ export class BundleLifecycleManager {
...
+    name: string,
+    wsId: string,
+    registry: ToolRegistry,
+  ): Promise<{ from: string; to: string; serverName: string }> {
+    const serverName = deriveServerName(name);
+    const instance = this.instances.get(`${serverName}|${wsId}`);
+    if (!instance) {
+      throw new Error(`No bundle instance found for "${name}" in workspace "${wsId}"`);
+    }
+
```
mgoldsborough, line 422: **No `protected` check on upgrade.**

`uninstall` (L336) refuses on `instance.protected`. Upgrade has no equivalent guard, so a protected bundle can be hot-swapped to a new version even though it can't be uninstalled.

This is probably intentional — protected bundles shouldn't be frozen out of security updates — but worth an explicit decision and a one-line code comment so a future contributor doesn't "fix" it by adding the check (or vice versa).

mgoldsborough (changes requested):

src/bundles/lifecycle.ts
Ovaculos: Fixed. JSDoc now says "Best-effort hot-swap" and describes the actual remove-then-start sequence. No atomicity claims. PR description updated to match.

src/bundles/lifecycle.ts
Ovaculos: Fixed. `removeBundleAutomations(name, registry)` now called before `syncBundleAutomations()` in upgrade, matching the uninstall pattern. Stale schedules from dropped manifest entries are cleaned up.

web/src/pages/settings/AboutTab.tsx
Ovaculos: Fixed. `result.isError` checked explicitly after `callTool`. Error message extracted from content blocks and routed to `setUpgradeErrors` so the inline error span renders. Also fixed a TS2345 in the content filter/map — removed explicit type annotations and used `c.text ?? ""` + `|| "Upgrade failed"`.

src/tools/system-tools.ts
Ovaculos: Fixed. Added `installSource` field (`"registry" | "local" | "remote"`) to `BundleInstance` and `AppInfo`. Set during each install path (`installNamed` → `"registry"`, `installLocal` → `"local"`, `installRemote` → `"remote"`, `seedInstance` → derived from `BundleRef` shape). `check_updates` and frontend filter on `installSource === "registry"` instead of the `bundleName.startsWith("@")` heuristic.

test/unit/bundle-upgrade.test.ts
Ovaculos: Fixed. 13 unit tests now covering: upgrade action guards, schema validation, `check_updates` filtering by `installSource`, upgrade no-op when already at latest (verifies no events emitted, source untouched), and unknown instance error path.

src/bundles/lifecycle.ts
Ovaculos: Fixed. Extracted to `resolveWorkDir()` function. Intentionally a function, not a module-level constant — constants capture `process.env` at import time, which breaks integration tests that override `NB_WORK_DIR` at runtime. All 4 call sites now use `resolveWorkDir()`.

src/bundles/lifecycle.ts
Ovaculos: Intentional — protected bundles can't be uninstalled but should receive version updates (security patches). Added code comment in `upgrade()` explaining the decision so future contributors don't accidentally add/remove the guard.

mgoldsborough:
Superseded by #291, which revives this work on current `main`, resolves the rebase conflicts, and adds a durable fix for the shared-cache staleness incident (a bad cached bundle version self-heals on the next spawn).

Thanks @Ovaculos for the original implementation — the upgrade flow and update UI carry forward from here, and your co-authorship is preserved on #291. Closing in favor of #291.

PR #170, opened 2026-05-04
feat: upload and install custom .mcpb bundles from web UI

### LINKED ISSUE #169: feat: upload and install custom .mcpb bundles from the web UI
## Summary

Installing custom bundles not on the mpak registry requires manually unpacking and configuring them. There should be a way to upload `.mcpb` files directly through the web UI and have them validated and installed.

## Proposed

1. **REST upload endpoint** (`POST /v1/bundles/upload`): accepts multipart `.mcpb` file, saves to `{workDir}/workspaces/{wsId}/bundles/`, validates via `@nimblebrain/mpak-sdk`'s `validateMcpb()`, returns the saved path + manifest.

2. **Extended `manage_app` tool**: accepts optional `path` parameter for installing bundles by file path (in addition to existing `name` for registry bundles). Both paths share a single install function.

3. **`.mcpb` routing in `startBundleSource`**: archives are routed through the SDK's `prepareServer({ local })` instead of `buildLocalSource()` (which expects unpacked directories).

4. **Upload button on About tab**: "Installed Bundles" section gets an Upload button that chains the REST upload → MCP tool install flow.

## Dependencies

Requires NimbleBrainInc/mpak#93 (`validateMcpb()` export from the SDK).

Uploaded bundles are treated identically to registry and local directory bundles — no separate "sideloaded" distinction.

Ovaculos:
## Summary

- Adds `POST /v1/bundles/upload` REST endpoint for multipart `.mcpb` upload with SDK validation
- Extends `manage_app` tool to accept `path` for file-based bundle install (unified install function for both registry and uploaded bundles)
- Routes `.mcpb` archives through SDK's `prepareServer({ local })` in `startBundleSource`
- Adds Upload button to About tab's "Installed Bundles" section

## Dependencies

Requires [mpak#94](https://github.com/NimbleBrainInc/mpak/pull/94) — adds `validateMcpb()` to `@nimblebrain/mpak-sdk`.

### Using the updated SDK locally

Until mpak#94 is published to npm, symlink the local SDK build:

```bash
rm -rf node_modules/@nimblebrain/mpak-sdk
ln -s ../../mpak/packages/sdk-typescript node_modules/@nimblebrain/mpak-sdk
```

After mpak#94 is released and `@nimblebrain/mpak-sdk` is bumped in `package.json`, remove the symlink and run `bun install`.

## QA review changes (post 47496b1)

Ten follow-up commits address the [QA review](https://github.com/NimbleBrainInc/nimblebrain/pull/170) findings:

**Critical security**

- `fix(bundles): confine manage_app.path to workspace bundles dir` (bf65f8d) — `assertPathInWorkspaceBundlesDir` rejects any LLM-supplied `path` outside `<workDir>/workspaces/<wsId>/bundles/`. Wired into install, uninstall, and startup re-hydration. Realpaths both sides so symlinks pointing outside the dir are rejected.
- `fix(api): validate uploaded .mcpb in tempfile before commit to bundles dir` (96726b2) — Validate-then-commit flow. Upload writes to `os.tmpdir()`, runs `validateMcpb`, then `renameSync` into the bundles dir on success. Failure or thrown error unlinks the temp. Bundles dir now holds validated content only, regardless of crash timing.

**Other fixes**

- `fix(api): randomize uploaded .mcpb filename to prevent silent overwrite` (bfb85dc) — 64-bit random hex suffix before `.mcpb`. Two uploads of `bundle.mcpb` no longer clobber each other. Frontend display name comes from manifest, unaffected.
- `fix(api): enforce MAX_BUNDLE_SIZE post-buffer in handleBundleUpload` (e9117f9) — Authoritative `data.length` check after buffering. Route-level `bodyLimit` is advisory (Content-Length only); chunked transfer encoding bypasses it. Constant exported from `handlers.ts`, shared with route middleware.

**Refactors**

- `refactor(api): hoist node:fs and mpak-sdk imports in handleBundleUpload` (a84adc0) — Drops dynamic `await import` calls; handlers.ts loads on every API boot so deferring offered no benefit.
- `refactor(api): extract UploadedFileEntry helper, adopt in upload handlers` (45df46d) — `asUploadedFile` helper in `src/api/types.ts` replaces inline `as unknown as { arrayBuffer(): Promise<ArrayBuffer>; ... }` casts in three handlers (`handleChatMultipart`, `handleResourceUpload`, `handleBundleUpload`).

**Web UI**

- `feat(web): track upload phase explicitly in AboutTab` (3ffe22e) — `uploadPhase: "idle" | "validating" | "installing"` + `uploadError: string | null`. `isUploading` derived from phase. Replaces brittle string-equality checks against literal copy. `finally` resets phase on thrown errors.

**Tests**

- `test: import production filter in mcpb-upload unit test` (89373d9) — Extracts `bundleEntryMatchesTarget` to a named export in `system-tools.ts`, used by both install dedup and uninstall filter. Test imports the production function rather than re-implementing it.
- `test: cover .mcpb path guard and upload integration` (3bb76b4) — New `test/unit/bundle-path-guard.test.ts` (6 tests) and `test/integration/mcpb-bundle-upload.test.ts` (5 tests) covering valid upload, filename uniqueness, traversal stripping, non-`.mcpb` rejection, and invalid-archive rejection. **Integration tests will fail in CI until `@nimblebrain/mpak-sdk@>=0.7.0` is published and the dep is bumped past 0.6.0** — intentional, gates the SDK bump on the contract.

**Docs**

- `docs: changelog entry for .mcpb upload feature` (87e0307)

## Test plan

- [x] Upload valid `.mcpb` → validates, installs, appears in bundle list
- [x] Upload invalid file → error displayed inline
- [x] Install via `manage_app` with `path` → same behavior as `name` installs
- [x] `.mcpb` bundles survive server restart (persisted in `workspace.json`)
- [x] Typecheck passes (only pre-existing TS2688 ambient errors)
- [x] `bun run test:unit` (2356 pass)
- [x] `bun run test:web` (234 pass)
- [x] `bun run test:integration` locally (460 pass, 12 skip — `mcpb-bundle-upload.test.ts` passes via mpak-sdk symlink; will fail in CI until 0.7.0 published)
- [x] Path traversal in upload filename → stripped to basename inside bundles dir
- [x] LLM-supplied `manage_app({ path })` outside bundles dir → rejected
- [x] Two uploads of same source filename → distinct on-disk paths

Closes #169


mgoldsborough:
Lint is broken

src/api/handlers.ts
```diff
@@ -1348,3 +1348,88 @@ export async function handleResourceUpload(
...
+  }
+
+  // Save to workspace bundles directory
+  const { mkdirSync, writeFileSync } = await import("node:fs");
+  const { join: joinPath } = await import("node:path");
+
+  const bundlesDir = joinPath(runtime.getWorkspaceScopedDir(workspaceId), "bundles");
+  mkdirSync(bundlesDir, { recursive: true });
+
+  const safeName = sanitizeFilename(filename);
```
mgoldsborough: Path traversal: `sanitizeFilename` only strips control chars and quotes — it doesn't touch `/`, `\`, or `..`. A user-supplied filename like `../../etc/cron.daily/foo.mcpb` resolves outside `bundlesDir` and `writeFileSync` will happily write there. Use `path.basename(filename)` after the extension check, or reject any filename containing path separators.

src/api/handlers.ts
```diff
@@ -1348,3 +1348,88 @@ export async function handleResourceUpload(
...
+  const bundlePath = joinPath(bundlesDir, safeName);
+  writeFileSync(bundlePath, data, { mode: 0o600 });
+
+  // Validate via mpak SDK
+  const { validateMcpb } = await import("@nimblebrain/mpak-sdk");
```
mgoldsborough: `@nimblebrain/mpak-sdk@0.6.0` is the version pinned in package.json and it doesn't export `validateMcpb` — mpak#94 is still open. This will throw at runtime until that PR ships and the SDK pin is bumped here. Worth blocking the merge on mpak#94 + an SDK release rather than relying on the symlink workflow.

src/tools/system-tools.ts
```diff
@@ -190,24 +196,43 @@ export async function createSystemTools(
...
+          const target = name ?? path;
+          if (!target) {
+            return {
+              content: textContent("Either 'name' or 'path' is required for uninstall"),
+              isError: true,
+            };
+          }
           return await uninstallBundleFromWorkspaceViaCtx(
-            name,
+            target,
```
mgoldsborough: Uninstall by `path` doesn't actually work. `target` here is the absolute file path, which gets passed as `name` into `uninstallBundleFromWorkspaceViaCtx`. That helper does `deriveServerName(target)` → `foo-mcpb` (path basename), but the bundle was registered under the manifest-derived server name, so `lifecycle.getInstance` returns nothing and `uninstallBundleFromWorkspace` throws `No bundle "<basename>" found`.

The workspace.json filter further down (`ws.bundles.filter((b) => !("name" in b && b.name === name))`) also only matches the `name` variant, so even if the process were stopped the path entry would stay in workspace.json forever.

Needs the same `{ name?, path? }` threading you did for install.

src/bundles/startup.ts
```diff
@@ -317,6 +317,54 @@ export async function startBundleSource(
       },
       eventSink,
     );
+  } else if ("path" in ref && ref.path.endsWith(".mcpb")) {
+    // .mcpb archive — use the SDK to extract and prepare
+    const mpakHome = process.env.MPAK_HOME ?? join(homedir(), ".mpak");
+    const mpak = getMpak(mpakHome);
+    const nbWorkDir = opts?.workDir ?? process.env.NB_WORK_DIR ?? join(homedir(), ".nimblebrain");
+
+    const server = await mpak.prepareServer(
```
mgoldsborough: Unlike the named-bundle branch above, this skips `resolveUserConfig(...)` before calling `prepareServer`. Any uploaded bundle that declares `user_config` (API keys, etc.) will fail with a raw `MpakConfigError` here instead of getting the friendly `nb config set -w <wsId>` hint, and host env aliases won't be wired in.

src/bundles/startup.ts
```diff
@@ -317,6 +317,54 @@ export async function startBundleSource(
...
+      {
+        type: "stdio",
+        spawn: {
+          command: server.command,
+          args: server.args,
+          env: {
+            ...server.env,
+            ...filterEnvForBundle(process.env as Record<string, string>, undefined, ref.allowedEnv),
+            ...(ref.env ?? {}),
+            ...platformEnv,
```
mgoldsborough: This env block drifts from both the named-bundle branch (lines 305-316) and `buildLocalSource`:
- No `MPAK_WORKSPACE` / `UPJACK_ROOT` — Upjack-based `.mcpb` bundles will run against the wrong directory.
- No `internalEnv` injection (`NB_INTERNAL_TOKEN` / `NB_HOST_URL`) for `ref.protected` bundles.
- `prepareServer`'s `workspaceDir` is `opts?.dataDir ?? nbWorkDir` — for cold spawns this can be the global workdir instead of a per-bundle dir.

Probably want to share the env-building with the named branch rather than re-rolling it here.

src/bundles/startup.ts
```diff
@@ -317,6 +317,54 @@ export async function startBundleSource(
...
+    const mpakHome = process.env.MPAK_HOME ?? join(homedir(), ".mpak");
+    const mpak = getMpak(mpakHome);
+    const nbWorkDir = opts?.workDir ?? process.env.NB_WORK_DIR ?? join(homedir(), ".nimblebrain");
+
+    const server = await mpak.prepareServer(
+      { local: ref.path },
+      { workspaceDir: opts?.dataDir ?? nbWorkDir },
+    );
+
+    const serverName = deriveServerName(server.name);
```
mgoldsborough: Server-name collision: this registers under `deriveServerName(manifest.name)`, but `serverNameFromRef` in `workspace-ops.ts` derives from the path basename for `{path}` refs. So:
- The `registry.hasSource(serverName)` duplicate check upstream uses the wrong name and won't catch real collisions (re-uploading the same bundle from a different path).
- `resolveBundleDataDir(wsPath, ref.path)` produces an ugly, non-portable data dir; on restart, `buildProcessInventory` rebuilds inventory keyed by the path-derived name that doesn't match the registered one.

Either make `serverNameFromRef` / `bundleNameFromRef` `.mcpb`-aware (peek the manifest), or carry the manifest name through `installBundleInWorkspace` instead of re-deriving in two places.

mgoldsborough (changes requested):
Critical issues inline. Path traversal on upload and the broken uninstall-by-path are blockers; the `.mcpb` spawn branch also drifts from the named/local branches in ways that will bite real bundles. SDK dep also isn't released yet.

Ovaculos:
## ⛔ Blocked: needs mpak to have `validateMcpb()` (https://github.com/NimbleBrainInc/mpak/pull/94)

This PR introduces three runtime call sites for `validateMcpb` that don't exist in the currently-pinned `@nimblebrain/mpak-sdk@0.6.0`:

- `src/api/handlers.ts` — `handleBundleUpload` (validates the uploaded archive before save)
- `src/bundles/startup.ts` — `.mcpb` branch of `startBundleSource` (peeks manifest for credential resolution + canonical naming)
- `src/tools/system-tools.ts` — `uninstallBundleFromWorkspaceViaCtx` (peeks manifest to derive the registered server name)

CI surfaces the gap as three `TS2339: Property 'validateMcpb' does not exist` errors. Those errors are intentionally kept in place — they're the visible signal that the SDK bump is the prerequisite.

**Unblocks when:**
1. mpak#94 (the PR exporting `validateMcpb`) is merged
2. mpak cuts a release that includes it (target: `0.7.0`)
3. `package.json` here bumps `@nimblebrain/mpak-sdk` to that version
4. `bun install` repopulates `node_modules`

**Verification after bump:**
- `bun run check` clean (no TS2339 for `validateMcpb`)
- `bun run test:unit` — 2349/2349 pass
- Manual `.mcpb` upload in the web UI completes end-to-end (a malformed bundle should hit the friendly `validateMcpb` rejection, not the previous `is not a function` crash)

**Could a maintainer add a `blocked` (or equivalent) label?** Fork contributors don't have label permissions.

**Once unblocked, the deferred test work in this branch can land:** `test/unit/mcpb-upload.test.ts` currently covers Fix 1 (path traversal) and Fix 3b (workspace.json filter). Integration coverage for Fix 4/5/6 (the `validateMcpb`-dependent paths) was pulled because `bun:test`'s `mock.module` is process-global and was leaking SDK stubs into other test files. That coverage can return in `test/integration/` once real fixtures replace the mocks.

PR #192, opened 2026-05-08
feat(mcp-auth): brand the OAuth callback page

mgoldsborough:
## Summary

The OAuth callback page (served at `GET /v1/mcp-auth/callback` after a successful MCP connector authorization) was a bare unstyled HTML stub. This replaces it with a centered, brand-aligned interstitial that matches the platform's warm palette and typography.

- "You're in." headline in **Erode** (display serif), **NimbleBrain** wordmark with warm-orange diamond mark in **Satoshi** — same Fontshare load the web app uses
- Warm neutrals: `#faf9f7` / `#171717` light, `#0a0a09` / `#e5e5e5` dark
- `prefers-color-scheme` and `prefers-reduced-motion` respected
- 800ms redirect timing unchanged; meta-refresh fallback and dotted "go back →" link preserved
- `<title>` kept as "Authorization complete" — `mcp-auth-routes.test.ts` assertion is unchanged

Security guards (escapeHtml on `safeReturnUrl`, `<` injection guard on `safeJsReturnUrl`, http/https protocol allowlist) all preserved untouched.

## Test plan

- [x] `bun test test/unit/api/mcp-auth-routes.test.ts` — 13/13 pass
- [x] `bunx tsc --noEmit -p .` clean
- [x] Visually verified light + dark mode in Chrome DevTools
- [ ] Manual smoke: complete one real MCP OAuth flow (e.g. Granola) end-to-end and confirm the page renders + redirects to the connectors settings page

### DIFF
--- a/src/api/routes/mcp-auth.ts
+++ b/src/api/routes/mcp-auth.ts
@@ -267,13 +267,29 @@ export function mcpAuthRoutes(ctx: AppContext) {
     const safeJsReturnUrl = JSON.stringify(returnUrl).replace(/</g, "\\u003c");
     return c.html(
       `<!doctype html><html><head><meta charset="utf-8"><title>Authorization complete</title>
-        <meta http-equiv="refresh" content="1;url=${safeReturnUrl}"></head>
-        <body style="font-family:system-ui,sans-serif;padding:2rem;max-width:32rem;margin:0 auto">
-          <h3 style="margin:0 0 0.5rem">Authorization complete</h3>
-          <p style="color:#555">Returning to NimbleBrain…</p>
-          <p><a href="${safeReturnUrl}">Click here if you aren't redirected →</a></p>
-          <script>setTimeout(function(){location.replace(${safeJsReturnUrl});},800);</script>
-        </body></html>`,
+<meta name="viewport" content="width=device-width,initial-scale=1">
+<meta http-equiv="refresh" content="1;url=${safeReturnUrl}">
+<link rel="preconnect" href="https://api.fontshare.com" crossorigin>
+<link href="https://api.fontshare.com/v2/css?f[]=erode@500&f[]=satoshi@500,700&display=swap" rel="stylesheet">
+<style>
+html,body{margin:0;height:100%}
+body{font-family:'Satoshi',system-ui,-apple-system,BlinkMacSystemFont,sans-serif;background:#faf9f7;color:#171717;display:flex;flex-direction:column;align-items:center;justify-content:center;min-height:100vh;padding:1rem;box-sizing:border-box;-webkit-font-smoothing:antialiased}
+.h{font-family:'Erode',Georgia,serif;font-size:clamp(2.5rem,6.5vw,4.25rem);font-weight:500;letter-spacing:-0.02em;margin:0;animation:rise .35s ease-out both}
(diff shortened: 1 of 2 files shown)


mgoldsborough:
## QA review response

Adjudicated review against the codebase and pushed `576be93`.

**Accepted (fixed):**

- **CRITICAL — default CSP blocks all styling.** Verified: `securityHeaders()` middleware sets `Content-Security-Policy: default-src 'none'; frame-ancestors 'none'; base-uri 'none'` on every response, and `app.use("*", securityHeaders())` runs before `mcpAuthRoutes`. My visual verification used a `file://` preview that bypassed the middleware — that was the test gap. The route now sets its own CSP allowlisting exactly the inline `<style>` by sha256 (computed once at module load against the same constant the response embeds, so hash drift fails tests before shipping).
- **WARNING — drop Fontshare.** Done. 1s page life vs network round-trip is a poor bet, fonts are likely cached from prior web-app load anyway, and dropping it removes a third-party Referer leak on every successful OAuth callback for self-hosted operators. Tightens the CSP surface as a side benefit (no `style-src https://...`, no `font-src` allowlist).
- **WARNING — fallback link contrast.** Bumped light `.fb a` to `#404040` on `#faf9f7` (~9.7:1 AAA), dark `.fb a` to `#d4d4d4` on `#0a0a09` (~10.7:1 AAA). The "not redirecting?" preamble also passes AA now.
- **SUGGESTION — `lang="en"`.** Added.
- **SUGGESTION — delete inline `<script>`.** Done. It was already CSP-blocked in production (silent dead code), so removing it is purely good. Lets us drop `script-src` from the route CSP and delete the `safeJsReturnUrl` injection-guard machinery. Cost: redirect goes from 800ms (JS) to 1000ms (meta-refresh) — acceptable for a transition page.
- **REGRESSION TEST.** Added in `mcp-auth-routes.test.ts`. Wires the real `securityHeaders()` middleware in front of the route and asserts (a) the response CSP is the route's, not the platform default, (b) the CSP includes `'sha256-…'`, and (c) that hash equals `sha256(servedInlineStyle)`. This locks in the route's override and pins the hash to the served bytes — both regressions the QA reviewer flagged are now caught at unit-test time.

**Rejected:**

- **SUGGESTION — `<meta charset>` should precede `<title>`.** Pre-existing, not introduced by this PR, browser sniffs UTF-8 reliably without it. Skipped to keep the diff minimal.
- **SUGGESTION — "Connected to NimbleBrain." vs "You're in."** "You're in." was the explicitly chosen direction. Stylistic call belongs to the requester.

**Verification this round (closing the gap):**

- Booted the live route via `dev:worktree` + a preview wrapper that mounts `mcpAuthRoutes` behind the real `securityHeaders()` middleware. Curl confirms response header is `Content-Security-Policy: default-src 'none'; style-src 'sha256-AQ6h5i6xv8Pv5yqtFljwGur/rfLpYH98TPfk4Dq1LPU='; frame-ancestors 'none'; base-uri 'none'`.
- Loaded in Chrome through the live route. Light + dark both render styled, **zero CSP console violations**, computed styles match design intent (Erode→Georgia fallback, body bg `#faf9f7`/`#0a0a09`, `.fb a` `#404040`/`#d4d4d4`, SVG fill `#d4620a`).
- `bun run verify:static` clean. `bun run verify:test-unit` 238/238 pass.

Issue #200, opened 2026-05-10
Vendor outreach: re-enable Canva, Intercom, Vercel DCR connectors after host allowlisting


mgoldsborough:
## Background

The 14 DCR connectors added in #195 advertise compliant DCR per RFC 7591. We verified `registration_endpoint` was discoverable for all of them via `bun run check:catalog`. In production, **3 of 14** turn out to enforce a vendor-side host allowlist that the spec-compliant DCR registration alone doesn't expose:

| Vendor | Failure point | Vendor message |
|---|---|---|
| `com.canva/mcp` | `/authorize` | `Invalid redirect URI. It must be from an allowed host.` |
| `com.intercom/mcp` | `/register` | `Redirect URI ... is not in the allowlist, reach out to Intercom Customer Support if you believe we should support it` |
| `com.vercel/mcp` | `/register` | `The provided redirect URIs are not approved for use by this authorization server.` |

Probed against `https://hq.platform.nimblebrain.ai/v1/mcp-auth/callback`. The other 11 (Notion, Linear, Cloudflare, Mercury, Neon, PayPal, Sentry, Stripe, Webflow, Wix, Granola) work end-to-end.

These three are dropped from `catalog.yaml` until the vendor allowlists our hosts. Re-adding criteria:
- Two-step `bun run check:catalog` probe (after the script extension lands) returns OK.
- A representative tenant (`hq.platform.nimblebrain.ai`) successfully completes install end-to-end.

### Canva
- Channel: [canva.dev](https://canva.dev/) developer portal / partner program.
- Ask: allowlist host pattern `*.platform.nimblebrain.ai` for OAuth redirect URI on the public MCP server (`mcp.canva.com`). If wildcards aren't supported, list each tenant URL individually (`hq.platform.nimblebrain.ai`, `acme.platform.nimblebrain.ai`, `bayze.platform.nimblebrain.ai`, `demo.platform.nimblebrain.ai`, `ipsdi.platform.nimblebrain.ai`).
- Note: their authorize endpoint enforces this independently of DCR registration — both layers need allowlisting.

### Intercom
- Channel: Intercom Customer Support (per their error message).
- Ask: same as Canva — allowlist `*.platform.nimblebrain.ai` (or each tenant) on the OAuth registration endpoint at `mcp.intercom.com`.

### Vercel
- Channel: Vercel developer support / OAuth integration team.
[...]

PR #203, opened 2026-05-12
fix(engine): normalize MCP tool schemas at LLM boundary; add Dropbox

mgoldsborough:
## Summary

- New `src/engine/tool-schema-for-llm.ts` at the engine→LLM boundary translates MCP-spec JSON Schema into the narrower shape OpenAI (Chat + Responses) and Anthropic Messages both require for tool input schemas. Single transform, single boundary — `mcp-source.ts` stays a faithful pass-through.
- Adds `com.dropbox/mcp` to the connector catalog as `auth: static`. Dropbox advertises DCR but only honors it for a curated partner allowlist (Claude Desktop, Cursor, ChatGPT); operators bring their own OAuth app from the Dropbox App Console until #200 outreach lands.

## Why

Both providers reject identical shapes that the MCP spec allows: top-level `oneOf` / `anyOf` / `allOf` / `enum` / `not`, and object schemas without an explicit `properties` field. Dropbox's `list_folder` (root-level `oneOf`) and `get_usage_and_quota` (bare `{type: "object"}`) trip both gates today; Notion / Linear / Sentry hit similar shapes in their long-tail tools. Translating at the LLM boundary fixes every current and future case in one place, rather than per-vendor patching.

## Translation rules

- Root `oneOf` / `anyOf` → take the **first branch**. Any valid call already satisfies exactly one branch; synthesizing a union schema would let the model emit mix-and-match payloads that match no real branch — strictly worse failure mode.
- Root `allOf` → deep-merge all branches (semantically faithful: \"all hold simultaneously\").
- Root `enum` / `not` → strip (no flat-object representation).
- Always emit `type: \"object\"` + `properties: {}` at root.
- **Nested composition (inside properties) is preserved** — providers only restrict the top level, so we keep every drop of vendor info we can.

## Test plan

- [x] 12 unit tests in `test/unit/tool-schema-for-llm.test.ts` cover base shape, top-level oneOf/anyOf/allOf, enum/not stripping, nested-composition preservation, and non-mutation of input
- [x] Full unit suite: 2592 pass / 0 fail
- [x] `bun run verify:static` clean (format, lint, tsc, cycles, bundle-transport, codegen)
- [ ] Manual: install Dropbox connector locally with a Dropbox App Console-registered OAuth app, verify `list_folder` is usable on both `claude-haiku-4-5` and `gpt-5.5`

### DIFF
--- a/src/connectors/catalog.yaml
+++ b/src/connectors/catalog.yaml
@@ -94,6 +94,38 @@ servers:
         auth: dcr
         tags: [infra, devops]
 
+  # Dropbox advertises DCR but only honors it for a curated partner
+  # allowlist (Claude Desktop, Cursor, ChatGPT) — `/register` rejects
+  # everyone else with "Dynamic client registration is not supported."
+  # Wired here as `auth: static` so operators can bring their own
+  # OAuth app from the Dropbox App Console. Drop back to `auth: dcr`
+  # once #200 outreach lands.
+  - name: com.dropbox/mcp
+    title: Dropbox
+    description: Files, folders, and shared links
+    version: "1.0.0"
+    icons:
+      - src: https://static.nimblebrain.ai/icons/dropbox.png
+        sizes: ["any"]
+    remotes:
+      - type: streamable-http
+        url: https://mcp.dropbox.com/mcp
+    _meta:
+      ai.nimblebrain/connector:
+        defaultScope: workspace
+        auth: static
+        requiredScopes:
+          - files.metadata.read
+          - files.content.read
+          - files.content.write
+          - sharing.write
+          - account_info.read
+        operatorSetup:
+          portalUrl: https://www.dropbox.com/developers/apps
+          hint: Create a Scoped Access app in the Dropbox App Console, enable the listed scopes, add the NimbleBrain redirect URI, then paste App key + App secret here
+          clientSecretKey: dropbox.client_secret
+        tags: [storage, files]
+
   # com.intercom/mcp: dropped pending vendor host allowlisting.
   # Intercom rejects DCR /register itself with "Redirect URI ... is
   # not in the allowlist, reach out to Intercom Customer Support if
--- a/src/engine/engine.ts
+++ b/src/engine/engine.ts
@@ -22,6 +22,7 @@ import {
   textContent,
 } from "./content-helpers.ts";
 import { withRetry } from "./retry.ts";
+import { toolSchemaForLlm } from "./tool-schema-for-llm.ts";
 import type {
   EngineConfig,
   EngineResult,
@@ -330,7 +331,7 @@ export class AgentEngine {
           type: "function" as const,
           name: t.name,
           description: t.description,
-          inputSchema: t.inputSchema as JSONSchema7,
+          inputSchema: toolSchemaForLlm(t.inputSchema) as JSONSchema7,
         }));
 
         const toolSchemaMap = new Map<string, ToolSchema>();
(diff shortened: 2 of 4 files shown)


mgoldsborough:
Follow-up commit `ac038a0` incorporates the QA feedback:

**#1 — Strategy change: first-branch → union-merge for oneOf/anyOf.** The decisive observation in QA's review: `engine.ts` validates the LLM's emitted args against the *original* schema (`toolSchemaMap` holds pre-transform Tools, Ajv compiles `oneOf` correctly). With that safety net intact, first-branch was strictly lossy — alternative branches like Dropbox `list_folder`'s `shared_link` form were unreachable for no real safety win. Now: union-merge properties, intersect required for oneOf/anyOf, union required for allOf. An invalid mix-and-match payload from the model costs at most one wasted LLM turn before Ajv rejects and the model self-corrects.

**#2 — `properties` defensiveness.** Now treats `null`, arrays, and any non-object the same as missing → `{}`. Honors the "object schema" contract consistently rather than only catching `undefined`.

**#3 — Warn on non-object input.** `log.warn` with the tool name when an inputSchema isn't an object/null/undefined. `null`/`undefined` stay silent (legitimate "no input"); other types are surfaced as upstream bugs. Tool name threaded through from `engine.ts`.

**#4 — Verification date.** Docstring now carries `Verified against OpenAI (Chat Completions + Responses) and Anthropic Messages as of 2026-05-12` with an explicit "revisit on API version bumps" note.

Property-key collisions across branches use last-wins (predictable spread idiom). Vendor schemas with collisions are malformed already, and the MCP server's own validator still gates the actual call.

Tests grow from 12 → 21 cases covering union-merge semantics, required arithmetic, collisions, defensive coercion across all malformed property shapes, warning behavior, and preservation of nested composition. 2601 unit tests passing; `verify:static` clean.

mgoldsborough:
Follow-up commit \`277f8f7\` reverts oneOf/anyOf to **first-branch** strategy after empirical regression.

**What happened:** Tested the union-merge build live against Dropbox \`list_folder\` (\`oneOf: [{path}, {cursor}]\`). Hit a runaway loop — model emitted \`{path: "", cursor: ""}\` (reasonable given the merged schema gave no signal of mutual exclusivity), Ajv-against-original rejected with \`must NOT have additional properties; must match exactly one schema in oneOf\`, and the model retried with cosmetic whitespace variations 10+ times to the iteration cap. Captured the trace in workspace conversation logs.

**Architectural lesson:** the "Ajv backs the call → cheap recovery" argument assumed validation errors are recoverable for the model. JSON Schema oneOf errors are not — they tell you the call failed but give zero signal about which field is the "extra" one or which branch was closest. Post-hoc validation is a safety net only when its output is actionable. For oneOf, it isn't.

**New strategy:**
- \`oneOf\` / \`anyOf\` → **first branch wins**. LLM sees a schema it can satisfy in one call. Alt branches unreachable; vendors order branches by canonical use, so the loss is real but bounded (Dropbox \`path\` is the 99% case).
- \`allOf\` → union-merge (semantically faithful — all branches hold simultaneously, no info lost).
- \`enum\` / \`not\` at root → stripped.
- When oneOf/anyOf has 2+ branches, \`log.warn\` fires naming the properties that exist only in dropped branches. Operators see exactly which vendor functionality is unreachable — addresses QA's earlier collision-warning suggestion at the right granularity (lost branches > collisions).

Docstring rewritten to capture the union-merge failure mode so future contributors don't re-derive the wrong answer. Permanent fix is upstream: vendors should flatten root composition since no LLM provider supports it.

21 tests; 2601 unit tests passing; \`verify:static\` clean.

PR #204, opened 2026-05-12
feat(oauth): signed state envelope for multi-tenant callback bouncer

mgoldsborough:
## Summary

Adds a signed-state envelope protocol that lets a single OAuth `redirect_uri` be reused across multiple tenants in a multi-tenant deployment. A separate router service verifies the signature and routes the callback back to the originating tenant.

This PR is the platform-side protocol module only; the router and its deployment are separate, downstream concerns.

## Why

In a multi-tenant deployment where each tenant has its own hostname, every OAuth Client at every vendor needs every tenant's redirect URI registered upfront. That scales as O(tenants × vendors) and creates a stale-URI hygiene problem on tenant offboarding. A signed-envelope `state` lets one registered URI per vendor serve all tenants.

Single-instance self-hosters are unaffected — this is library code with no consumers yet.

## What's in this PR

- `src/oauth/envelope.ts` — signing + verification of the state envelope, including HKDF-derived per-tenant keys.
- `test/unit/oauth-envelope.test.ts` — 23 tests covering round-trip, forgery rejection, malformed input, expiration, tid handling, and HKDF determinism.

## Wire format

```
state = "v1." base64url(payload_json) "." base64url(mac)
payload = { tid, inner, iat, exp }
tenant_key = HKDF-SHA256(master_key, salt=tid, info="oauth-bouncer/v1", L=32)
mac = HMAC-SHA256(tenant_key, "v1." base64url(payload_json))
```

`inner` carries the existing CSRF state token unchanged — the envelope is transparent to the cookie-bound check in `mcp-auth.ts`.

## Security model

| Layer | Stops |
|---|---|
| HKDF per-tenant keys | Cross-tenant lateral movement on pod compromise |
| HMAC verification | Forged routing claims |
| `tid` regex pin to DNS-label grammar | Path traversal / off-origin redirect via forged tid |
| `inner` ride-through to existing cookie hash check | Leaked-state replay against another user session |
| TTL on `exp` (15 min default) | Long-term replay |
| Constant-time MAC comparison | Side-channel key recovery |

## What this PR does NOT do

- Does NOT modify `src/api/routes/mcp-auth.ts` — the consumer wiring lands in a follow-up so this PR stays reviewable in isolation.
- Does NOT add config schema fields — same reason.
- Does NOT change behavior for single-instance self-hosters — library code, no consumers yet.

## Test plan

- [x] Unit tests pass (`bun test test/unit/oauth-envelope.test.ts` → 23/23)
- [x] Round-trip, forgery, expiration, and HKDF properties verified
- [x] Constant-time MAC comparison covered

### DIFF
--- a/src/oauth/envelope.ts
+++ b/src/oauth/envelope.ts
@@ -1,0 +1,345 @@
+import { createHmac, hkdfSync, timingSafeEqual } from "node:crypto";
+
+/**
+ * Signed state envelope for the OAuth callback bouncer.
+ *
+ * The platform supports two deployment modes for vendor OAuth callbacks:
+ *
+ * 1. **Direct mode** (default; single-tenant self-hosts):
+ *    `redirect_uri` registered at the vendor is the tenant's own host.
+ *    `state` is opaque; no envelope.
+ *
+ * 2. **Bouncer mode** (multi-tenant deployments):
+ *    A single `redirect_uri` is registered at every vendor — pointing
+ *    at a stateless router that fans callbacks out to the originating
+ *    tenant by name. The wire `state` is wrapped in this envelope so
+ *    the router can authenticate the routing decision without holding
+ *    session state.
+ *
+ * Wire format (URL-safe, compact, version-tagged):
+ *
+ *     v1.<base64url(payload)>.<base64url(mac)>
+ *
+ *     payload  = { tid, inner, iat, exp }   (canonical JSON)
+ *     mac      = HMAC-SHA256(tenantKey, "v1.<base64url(payload)>")
+ *     tenantKey = HKDF-SHA256(masterKey, salt=tid, info="oauth-bouncer/v1", L=32)
+ *
+ * `inner` is the existing CSRF state token (already bound to
+ * `nb_oauth_state` cookie by the callback route). The envelope does not
+ * replace that binding — it carries it through opaquely. Both checks
+ * must pass on callback: router-side HMAC and tenant-side cookie hash.
+ *
+ * Trust boundary:
+ *   - `masterKey` lives ONLY on the bouncer.
+ *   - Tenant pods receive a pre-derived `tenantKey` at deploy time.
+ *   - Compromise of a tenant pod leaks one derived key — forged states
+ *     for that tid route back to that same tid (no lateral movement).
+ */
+
+export const ENVELOPE_VERSION = "v1";
+const VERSION_PREFIX = `${ENVELOPE_VERSION}.`;
+const HKDF_INFO = Buffer.from("oauth-bouncer/v1");
+
+export const DEFAULT_TTL_SECONDS = 15 * 60;
+const CLOCK_SKEW_TOLERANCE_SECONDS = 60;
+
+/**
+ * Maximum lifetime of an envelope at verify time, regardless of what `exp`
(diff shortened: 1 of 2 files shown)


mgoldsborough:
Addressed all QA findings. Force-pushed.

**Critical fixes:**
- TS strict (`noUncheckedIndexedAccess`): introduced `splitWire()` helper that narrows the three wire segments to defined `string`s, eliminating the four index-narrowing errors. Used in both `peekTid` and `parseAndVerify` so length+version checks live in one place.
- Biome format: ran `bun run format`, amended.

**Polish fixes:**
- **Tightened `ALLOWED_TID_PATTERN`** to true RFC 1123 §2.1 DNS-label grammar (`^[a-z]([a-z0-9-]{0,61}[a-z0-9])?$`). Trailing hyphens no longer admitted. Test suite no longer has the `if (bad === "trailing-") continue` skip.
- **Added `MAX_ENVELOPE_LIFETIME_SECONDS = 30 * 60`** verifier-side cap. `parseAndVerify` rejects any envelope claiming `exp - iat > 30min`, regardless of MAC validity. Bounds the blast radius of a derived-key leak to 30 minutes instead of forever. New test exercises this with a hand-crafted 24h-lifetime forgery.
- **`deriveTenantKey` now asserts** `masterKey.length >= 32` at the boundary with a clear error message. Boot-time misconfig fails loud. New tests cover too-short and 32-byte cases.
- **Dropped the tautological `payload.tid !== peeked` check** in `verifyEnvelopeAsRouter`. The MAC verification already guarantees the payload bytes (and therefore the tid) match what was signed. Comment in source notes why we don't recheck.
- **Added 4 boundary tests:** inner-length at 512 (accept) and 513 (reject), clock-skew at 60s (accept) and 61s (reject).

**Verify:**
- `bun run verify:static` clean (format:check + lint + check + check:cycles + codegen)
- `bun test test/unit/oauth-envelope.test.ts` → 30 pass, 0 fail (up from 23)

Ready for re-review.

mgoldsborough:
Round-2 polish addressed. Force-pushed.

**Finding 1 — asymmetric `inner` bounds.** Accepted. `signEnvelope` now rejects `inner` longer than `MAX_INNER_LENGTH` (512) before producing wire bytes. The cap is hoisted into a named constant (`MAX_INNER_LENGTH`) and `validatePayloadShape` uses the same constant — no more magic 512 on the verify side. New test confirms sign-side rejection; the prior verify-side boundary test is retained as a belt-and-suspenders check.

**Finding 2 — restore the post-parse tid check.** Accepted, with an honest comment. This is the exact opposite of what the previous round told me to drop, and both reviewers have a defensible point — `payload.tid === peeked` is tautological today, but the property is one parser-swap away from silently breaking. For crypto-adjacent code the cost (one comparison) is negligible and the benefit (immunity to a future parsing-path change that would turn routing into an attacker-controlled primitive) is worth keeping. Comment in source notes the redundancy explicitly so future readers don't think the check is load-bearing today.

**Finding 3 — `require("node:crypto")` in tests.** Accepted. `createHmac` is now in the top import; the two inline `require(...) as typeof import(...)` casts are gone.

**Verify:**
- `bun run verify:static` clean (format:check + lint + check + check:cycles + codegen)
- `bun test test/unit/oauth-envelope.test.ts` → 31 pass, 0 fail (up from 30)

Ready for re-review.

PR #205, opened 2026-05-12
feat(oauth): consume state envelope on mcp-auth routes for bouncer mode

mgoldsborough:
## Summary

Wires the envelope module from #204 into `mcp-auth.ts` so the platform can opt into bouncer mode via env vars. When configured, the redirect URI sent to every vendor's OAuth Client points at a separate router service (deployed separately) instead of this tenant's own host; `state` is wrapped in a signed envelope so the router can route the callback back. Cookie binding and the single-use flow registry are unchanged.

Single-instance self-hosters who don't set the env vars keep today's behavior — no behavioral change for the default case.

## Why

Multi-tenant deployments need a single `redirect_uri` per vendor rather than one per tenant — registering N×M URIs at vendor consoles and cleaning them up on offboarding doesn't scale. The envelope landed in #204; this PR makes it actually used.

## What's in this PR

- `src/oauth/bouncer-config.ts` — process-wide config read from two bouncer-specific env vars plus the general-purpose `NB_TENANT_ID` primitive. Lazy, cached, all-or-nothing — partial configuration is a deploy-time mistake and fails closed.
- `src/api/routes/mcp-auth.ts` — three changes:
  - `mcpAuthCallbackUrl()` returns the bouncer URL when configured.
  - `/initiate` wraps the SDK-generated `state` in a signed envelope and replaces the URL's `state` query param before returning the auth URL. The cookie still keys on the inner state, so the existing CSRF binding works unchanged.
  - `/callback` unwraps the wire state into the inner before the existing cookie-hash check and flow-registry lookup. Non-wrapped state in bouncer mode is refused.
- `test/unit/oauth-bouncer-config.test.ts` — 12 tests covering presence, partial-config rejection, `NB_TENANT_ID` precondition, value validation, and caching.
- Six new tests in `test/unit/api/mcp-auth-routes.test.ts` for wrap-on-initiate, unwrap-on-callback, wrong-key forgery rejection, wrong-tid forgery rejection, non-wrapped-state rejection, and the callback URL switch.

## Configuration

Bouncer mode is opt-in via these env vars:

| Variable | Scope | Source |
|---|---|---|
| `NB_OAUTH_BOUNCER_CALLBACK_URL` | Cluster-wide | Helm `oauthBouncer.callbackUrl` — set once in `base-values.yaml` |
| `NB_OAUTH_BOUNCER_TENANT_KEY` | Per-tenant secret | ExternalSecret → k8s Secret. Provisioned by `make derive-tenant-key` at onboarding |
| `NB_TENANT_ID` | Per-tenant non-secret | Helm `tenant.id` — already required for every tenant |

`NB_TENANT_ID` is a general-purpose primitive (not OAuth-specific). Setting it alone does not enable bouncer mode. The two `NB_OAUTH_BOUNCER_*` vars are what trip the mode; `NB_TENANT_ID` is a precondition when they're set.

Why three rather than two earlier-revisions had: tying the envelope tid and the HKDF salt of the derived tenant key both to the same `tenant.id` Helm value eliminates a class of `bad_mac` errors from operator drift between deploy-time derivation and runtime signing.

## Local development

Direct mode is preserved. Run locally with no `NB_OAUTH_BOUNCER_*` vars set:

- `mcpAuthCallbackUrl()` returns `${NB_API_URL}/v1/mcp-auth/callback` as today
- `/initiate` sends opaque state to the vendor (no envelope)
- `/callback` accepts opaque state and runs existing cookie + flow-registry checks

For testing OAuth flows against localhost, register both URIs in the vendor console:

```
http://localhost:27247/v1/mcp-auth/callback          (local dev)
https://connect.<domain>/v1/mcp-auth/callback        (staging/prod bouncer)
```

Most vendors (Google, Dropbox, etc.) support multiple registered redirect URIs.

## Security properties

| Layer | Stops |
|---|---|
| All env vars validated at boot | Partial / malformed configuration silently weakening the system |
| `state` envelope (from #204) | Forged routing claims at the router |
| Tenant-side `expectedTid` match | Envelope minted for a different tenant routed to this one |
| `nb_oauth_state` cookie hash check | Leaked-state replay against another user session — unchanged |
| Single-use flow registry | Replay against the same user session — unchanged |
| `Authorization session invalid` on all envelope failures | Envelope-error oracle for attackers probing the wire format |

## What this PR does NOT do

- Does NOT modify any JSON config schema. Bouncer config is env-var only — appropriate because the tenant key is secret material that should arrive via `secretKeyRef`.
- Does NOT add deployments / Helm wiring. The companion infra PR (#9) updates the agent chart to inject these env vars from `tenant.id` and the `oauthBouncer` values block.

## Test plan

- [x] `bun run verify:static` clean (format + lint + check + codegen + cycle checks)
- [x] `bun test test/unit/oauth-envelope.test.ts test/unit/oauth-bouncer-config.test.ts test/unit/api/mcp-auth-routes.test.ts` → 63 pass, 0 fail
- [x] Existing mcp-auth route tests unchanged (14/14 still pass — direct mode unaffected)
- [ ] End-to-end test against a deployed bouncer (depends on deployments-repo PR)

### DIFF
--- a/src/api/routes/mcp-auth.ts
+++ b/src/api/routes/mcp-auth.ts
@@ -2,6 +2,13 @@ import { createHash, timingSafeEqual } from "node:crypto";
 import { Hono } from "hono";
 import { WORKSPACE_PRINCIPAL_ID } from "../../bundles/connection.ts";
 import { log } from "../../cli/log.ts";
+import { getBouncerMode } from "../../oauth/bouncer-config.ts";
+import {
+  ENVELOPE_VERSION,
+  EnvelopeError,
+  signEnvelope,
+  verifyEnvelopeAsTenant,
+} from "../../oauth/envelope.ts";
 import { resolveWithCode } from "../../tools/oauth-flow-registry.ts";
 import { requireAuth } from "../middleware/auth.ts";
 import { requireWorkspace } from "../middleware/workspace.ts";
@@ -75,11 +82,25 @@ const SUCCESS_PAGE_CSP = `default-src 'none'; style-src 'sha256-${SUCCESS_PAGE_S
  * they leave to set up the OAuth app.
  */
 export function mcpAuthCallbackUrl(): string {
+  // In bouncer mode every vendor's OAuth Client is registered against
+  // the bouncer's single URL, not this tenant's host. The bouncer
+  // verifies the signed state envelope on the callback leg and 302s
+  // back to this tenant.
+  const bouncer = getBouncerMode();
+  if (bouncer) return bouncer.callbackUrl;
+
   const apiBase = process.env.NB_API_URL ?? "http://localhost:27247";
   return `${apiBase.replace(/\/+$/, "")}/v1/mcp-auth/callback`;
 }
 
 export function mcpAuthRoutes(ctx: AppContext) {
+  // Eagerly validate bouncer config (if any) so a misconfigured
+  // deployment fails at server startup with a precise error, rather
+  // than serving traffic until the first user clicks "connect" and
+  // hitting a generic 500. Idempotent: returns the cached value on
+  // subsequent calls in the route handlers below.
+  getBouncerMode();
+
   const app = new Hono<AppEnv>();
 
   // ── POST /v1/mcp-auth/initiate ────────────────────────────────────
@@ -178,6 +199,34 @@ export function mcpAuthRoutes(ctx: AppContext) {
         );
       }
 
+      // In bouncer mode, wrap the SDK-generated state in a signed
+      // envelope so the bouncer can route the callback back to this
+      // tenant. The inner state is what's bound to the cookie and what
+      // `oauth-flow-registry` is keyed on — both unchanged. The vendor
+      // sees only the wrapped value.
+      const bouncer = getBouncerMode();
+      if (bouncer) {
+        try {
+          const wrapped = signEnvelope({
+            tid: bouncer.tid,
+            inner: state,
+            tenantKey: bouncer.tenantKey,
+          });
+          urlObj.searchParams.set("state", wrapped);
+          authorizationUrl = urlObj.toString();
+        } catch (err) {
+          // signEnvelope rejects inputs that violate the envelope's
+          // own contract (oversize inner, invalid tid). Both are
+          // pre-validated above (tid at config load, inner from the
+          // SDK), so reaching here implies a regression elsewhere.
+          // Match the error posture of the startAuth catch above:
+          // log the cause, surface a generic 500.
+          const msg = err instanceof Error ? err.message : String(err);
+          log.warn(`[mcp-auth] envelope wrap failed for ${serverName} in ${wsId}: ${msg}`);
+          return apiError(500, "internal_error", "Failed to wrap OAuth state.");
+        }
+      }
+
       const stateHash = sha256Hex(state);
 
       // Cookie scoped to /v1/mcp-auth/callback so it's only sent on the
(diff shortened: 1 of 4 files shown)


mgoldsborough:
First-principles review pass: dropped the bouncer-specific tid env var and reused the general-purpose `NB_TENANT_ID` instead. Force-pushed.

**Why the change.** Before this revision, the platform pod read three OAuth-specific env vars including `NB_OAUTH_BOUNCER_TID`. That invented a tenant identity surface used only for the bouncer, even though the chart already had `.Values.tenant.id` for the same value (used for namespace labels, audit context). Two problems:

1. **Drift risk.** The HKDF derivation script at onboarding time uses one tid; the envelope signing at runtime uses another env var. If those drift through a typo or copy-paste error, bouncer mode silently fails with `bad_mac` from the verifier — a real footgun, not a hypothetical one.

2. **Accidental complexity.** Tenant identity is a generic primitive useful beyond OAuth (structured logs, telemetry, error reporting). Inventing it bouncer-specifically locks the surface to one feature.

**The fix.** Introduce `NB_TENANT_ID` as a first-class env var sourced from `.Values.tenant.id`. The bouncer config reads tid from there. Same `tenant.id` value feeds both the HKDF salt at deploy time and the envelope signing at runtime — drift is structurally impossible.

**What this means for callers.**
- `NB_TENANT_ID` is general-purpose; setting it alone does not enable bouncer mode.
- The two `NB_OAUTH_BOUNCER_*` vars trip bouncer mode; `NB_TENANT_ID` becomes a precondition when they're set.
- Existing tenants get `NB_TENANT_ID` injected automatically via the chart change in companion PR #9; no per-tenant Helm value edits required.

Updated test (12/12 in bouncer-config, 20/20 in mcp-auth-routes — added one for "NB_TENANT_ID alone doesn't trip bouncer mode"). PR description rewritten to reflect the change.

mgoldsborough:
Round-2 feedback addressed. Force-pushed.

**Finding 1 — eager validation.** Accept. Added a `getBouncerMode()` call at the top of `mcpAuthRoutes(ctx)`. Hono mounts routes at app construction, so this runs once at server startup. Misconfigured deployments now crashloop with a precise error instead of serving traffic until the first OAuth attempt hits a generic 500. Updated the module doc to match the new behavior (was "fail at boot" vs lazy reality — now actually eager).

**Finding 2 — `signEnvelope` can throw uncaught.** Accept. Wrapped in try/catch matching the existing `startAuth` error posture at lines 163-175: log the cause server-side, return `apiError(500, "internal_error", "Failed to wrap OAuth state.")`. Both paths the reviewer flagged are unreachable in practice (tid pre-validated at config load, SDK won't produce empty/oversize state), but the catch defends a future regression class consistent with the rest of the route.

**Finding 3 — envelope-failure cookie not expired.** Skipped on the reviewer's own "happy to leave" guidance. The behavior matches every other error branch in this route (only the success path clears the cookie). Adding clear-on-envelope-fail without changing the other paths would be the new inconsistency.

**Finding 4 — callback URL path not validated.** Accept with hard error. The `nb_oauth_state` cookie is `Path=/v1/mcp-auth/callback` in the route — a bouncer URL ending elsewhere is incoherent by construction and produces a debugging nightmare (every flow fails with "session mismatch", no signal pointing at the URL). Added a check that requires the URL pathname to end with `/v1/mcp-auth/callback`, surfaced via a clear error message naming the expected path and why. The expected path is a constant in `bouncer-config.ts`, hardcoded to match the route.

**Finding 5 — CHANGELOG.** Skipped per repo convention (per-release, not per-PR). Worth flagging to whoever cuts the release that #204 + #205 + the infra PR should land together under one "Added: OAuth callback bouncer mode" bullet.

**Verify:**
- `bun test test/unit/oauth-bouncer-config.test.ts` → 14 pass (up from 12, added 2 path-validation tests)
- `bun test test/unit/api/mcp-auth-routes.test.ts` → 20 pass (unchanged)

Ready for re-review.

PR #206, opened 2026-05-12
fix(oauth): tighten envelope inner-length, key, and tid checks

mgoldsborough:
## Summary

Four small follow-ups to the envelope module from #204, all driven by review feedback. Originally three; expanded after a round-2 suggestion flagged a symmetric defense gap on the derived tenant key.

### 1. `inner` length is measured in UTF-8 bytes

`String.prototype.length` counts UTF-16 code units. A 257-emoji inner string is 514 UTF-16 units but ~1028 UTF-8 bytes, and the post-serialization payload-cap check catches the oversize wire either way. The asymmetry is harmless in practice (CSRF tokens are ASCII), but the intent of `MAX_INNER_LENGTH` is "limit bytes on wire" — switching to `Buffer.byteLength(opts.inner, "utf8")` makes both checks consistent in units. Both `signEnvelope` and `validatePayloadShape` now agree.

### 2. `deriveTenantKey` rejects placeholder master keys

`Buffer.alloc(32)` is a real deployment foot-gun: an operator who wires up the env var with an unset secret or a copy-pasted placeholder gets a 32-byte buffer of zeros. The length check passes, the system runs, and every "signed" state is a deterministic function of payload. The new check rejects all-zeros and all-0xff master keys with a clear error. Zero false positives — CSPRNG output never matches these patterns.

### 3. `bouncer-config` rejects placeholder tenant keys

Symmetric defense at the derived layer. The pre-derived `NB_OAUTH_BOUNCER_TENANT_KEY` is loaded at config time and previously only got a length check. Same operator failure mode as the master (base64 of `Buffer.alloc(32)` passing length validation), smaller blast radius (one tenant vs. all). `isUniformByte` is exported from `envelope.ts` so the predicate has one source of truth; if we ever extend it to catch other patterns, both layers update together.

### 4. Comment on the unreachable defensive tid check

The `peeked !== payload.tid` check in `verifyEnvelopeAsRouter` is unreachable today because V8's `JSON.parse` is deterministic over identical bytes. The check was retained as defense-in-depth for a future parsing-path change. Added a comment explaining that no direct unit test exists for this branch, since constructing a divergence requires the same future change the guard exists to survive — and that removing the check should be paired with a single-pass parser instead.

No code change to the check itself; just the comment.

## Test plan

- [x] `bun run verify:static` clean
- [x] `bun test test/unit/oauth-envelope.test.ts` → 36 pass (was 31)
- [x] `bun test test/unit/oauth-bouncer-config.test.ts` → 16 pass (was 14)

New tests:
- Inner exactly at the 512-byte boundary with multibyte chars (accept)
- Inner over the 512-byte boundary but under the 512 UTF-16-unit boundary (reject — the pre-fix behavior would have accepted)
- All-zeros master rejected
- All-0xff master rejected
- 31 zero bytes + one real byte accepted (the entropy floor is "not the obvious placeholder," not "looks random")
- All-zeros tenant key rejected at config-load
- All-0xff tenant key rejected at config-load

### DIFF
--- a/src/oauth/bouncer-config.ts
+++ b/src/oauth/bouncer-config.ts
@@ -42,7 +42,7 @@
  * attempt. Subsequent calls hit the cache and don't re-validate.
  */
 
-import { ALLOWED_TID_PATTERN } from "./envelope.ts";
+import { ALLOWED_TID_PATTERN, isUniformByte } from "./envelope.ts";
 
 export interface BouncerMode {
   /** Full URL registered at every vendor's OAuth Client. */
@@ -170,6 +170,15 @@ function readAndValidate(env: NodeJS.ProcessEnv): BouncerMode | null {
   if (tenantKey.length !== REQUIRED_KEY_BYTES) {
     throw new Error(
       `[oauth bouncer] ${TENANT_KEY_ENV} must decode to ${REQUIRED_KEY_BYTES} bytes (got ${tenantKey.length})`,
+    );
+  }
+  // Symmetric defense with the master-key check in `deriveTenantKey`.
+  // The blast radius is smaller here (one tenant, not all), but the
+  // operator failure mode is identical: a base64-encoded placeholder
+  // buffer lands in 1Password and passes the length check.
+  if (isUniformByte(tenantKey, 0) || isUniformByte(tenantKey, 0xff)) {
+    throw new Error(
+      `[oauth bouncer] ${TENANT_KEY_ENV} is a placeholder pattern (all 0x00 or all 0xff); ensure HKDF derivation ran during onboarding`,
     );
   }
 
--- a/src/oauth/envelope.ts
+++ b/src/oauth/envelope.ts
@@ -64,7 +64,7 @@ const MAX_PAYLOAD_BYTES = 1024;
  * inputs upfront rather than producing wire bytes that every peer
  * refuses to verify.
  */
-const MAX_INNER_LENGTH = 512;
+export const MAX_INNER_LENGTH = 512;
 
 /**
  * Tenant identifiers are interpolated into hostnames downstream by the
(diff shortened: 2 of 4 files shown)


mgoldsborough:
Expanded scope: applied the same placeholder-pattern check to the pre-derived tenant key at config-load time, mirroring what we already had for the master key in `deriveTenantKey`. Force-pushed.

**Why expand the PR rather than open a follow-up:** the reviewer's note was "out of scope per PR title, but adjacent." The cost was ~5 lines + two tests, and the defense is the same defense one layer deeper. Splitting it across two PRs is more process than the change itself. PR title updated to drop "master-key" (now just "key") since both key boundaries are covered.

**Implementation:** `isUniformByte` is now exported from `envelope.ts` so master and tenant-key checks share one predicate. If the placeholder set ever grows (other obvious patterns, common test vectors, etc.), both layers update together.

Verify:
- `bun test test/unit/oauth-envelope.test.ts` → 36 pass
- `bun test test/unit/oauth-bouncer-config.test.ts` → 16 pass (added 2 placeholder-pattern tests)

mgoldsborough:
Round-3 feedback addressed.

**Finding 1 — magic number duplicated from envelope.ts.** Accept. Exported `MAX_INNER_LENGTH` from `envelope.ts` and imported into the test. Replaced the local `MAX_INNER_LENGTH_FOR_TEST` constant and the pre-existing `512` / `513` literals from earlier boundary tests with references to the imported constant. The multibyte tests now compute the emoji count from `MAX_INNER_LENGTH / 4` so they remain on-boundary if the cap ever moves. Confirmed via grep that no stray `512`/`513` literals remain in the test file.

**Finding 2 — `isUniformByte` only catches `0x00`/`0xff`.** Skip per the reviewer's own framing ("not in scope per the PR's own framing"). The current check targets the obvious `Buffer.alloc(32)` foot-gun. Extending to catch arbitrary uniform patterns (or sequential patterns, or common test vectors) is a different design decision worth its own discussion. The export from #206 means the source of truth is in one place if we ever decide to widen the predicate.

Verify: 36 pass, 0 fail. `bun run verify:static` clean. Force-pushed.

I think we're at the point of diminishing returns now. The remaining concerns are either contained (the predicate could grow, but the export makes it easy) or genuinely out of scope. Happy to do another round if you spot something, but this feels like a reasonable shipping point.

PR #207, opened 2026-05-13
fix(oauth): drive bouncer mode off URL presence; warn on stray key

mgoldsborough:
## Summary

The platform's bouncer config was strict in both directions — URL set without key OR key set without URL — and that two-way strictness turned out to be wrong in one direction. This change keeps the strict check where strictness matters and relaxes it where it doesn't.

## Why

Surfaced during the OAuth bouncer rollout (deployments PR #12). The agent chart uses `envFrom: secretRef` on `agent-secrets`, which pulls **every** key in the Secret into env regardless of whether the chart's `oauthBouncer.enabled` block renders the bouncer-specific env vars. So rolling a tenant back from bouncer mode (set `oauthBouncer.enabled: false`, redeploy) **doesn't** remove `NB_OAUTH_BOUNCER_TENANT_KEY` from env — it stays as long as the ExternalSecret still lists the field. The platform's old strict check then crashed the pod with `Partial configuration. Missing: NB_OAUTH_BOUNCER_CALLBACK_URL`, turning routine rollback into an incident. The deployments PR documented a two-step rollback workaround (edit the ExternalSecret first, then flip the Helm value), but that's friction operators forget.

## What changes

`NB_OAUTH_BOUNCER_CALLBACK_URL` becomes the authoritative mode signal:

| `URL` | `KEY` | Before | After |
|---|---|---|---|
| set | set | bouncer mode | bouncer mode (unchanged) |
| set | unset | fatal (partial config) | fatal: operator clearly meant to enable bouncer (unchanged) |
| unset | set | **fatal (partial config) ← rollback footgun** | **direct mode + one-time boot warn** |
| unset | unset | direct mode | direct mode (unchanged) |

The stray-key warn fires once at boot (via the eager `getBouncerMode()` call in `mcpAuthRoutes`) and names the cleanup step: drop the entry from the ExternalSecret and kubectl apply. The key is harmless when unused — no code reads it once URL is gone — so direct-mode operation continues.

## Test plan

- [x] 16/16 unit tests in `oauth-bouncer-config.test.ts` (incl. 3 new "URL is the mode signal" cases)
- [x] 20/20 unchanged tests in `mcp-auth-routes.test.ts`
- [x] `bun run verify:static` clean
- [ ] After deploy: confirm the stray-key warn fires on a tenant rolled back from bouncer mode

## Follow-ups

- Deployments-side docs (`hq.yaml` comment + `ONBOARD_TENANT.md`) currently describe the two-step rollback. Once this lands and ships to production, update the docs to describe the one-step toggle as the primary path. Separate PR; depends on this image rolling out to relevant tenants.

### DIFF
--- a/src/oauth/bouncer-config.ts
+++ b/src/oauth/bouncer-config.ts
@@ -32,16 +32,28 @@
  *     runtime, so plumbing both endpoints from one source eliminates
  *     a class of `bad_mac` mismatches from typos.
  *
- * Bouncer mode requires all three. Partial configuration of the two
- * `NB_OAUTH_BOUNCER_*` vars, a missing `NB_TENANT_ID` while bouncer
- * mode is requested, or any malformed value is a deploy-time error
- * that fails closed at the first call to `getBouncerMode()`. The
- * mcp-auth route module calls this at module load via `mcpAuthRoutes`,
- * so a misconfigured deployment crashes immediately at server startup
- * rather than serving traffic and failing on the first user OAuth
- * attempt. Subsequent calls hit the cache and don't re-validate.
+ * `NB_OAUTH_BOUNCER_CALLBACK_URL` is the mode signal — its presence,
+ * specifically, enables bouncer mode. When it is set, the key and the
+ * tenant id must also be set (and valid); a missing or malformed value
+ * in that state is a deploy-time error that fails closed at the first
+ * call to `getBouncerMode()`. The mcp-auth route module calls this at
+ * module load via `mcpAuthRoutes`, so a misconfigured deployment
+ * crashes immediately at server startup rather than serving traffic
+ * and failing on the first user OAuth attempt.
+ *
+ * The asymmetric case: `NB_OAUTH_BOUNCER_TENANT_KEY` set without the
+ * URL. This is the shape of a rolled-back tenant where the chart's
+ * `envFrom: secretRef` keeps injecting the key from `agent-secrets`
+ * even though `oauthBouncer.enabled: false`. Treat it as direct mode
+ * and log a single boot-time warning suggesting the cleanup step
+ * (drop the entry from the ExternalSecret). Crashing here would brick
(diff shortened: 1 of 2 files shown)


mgoldsborough:
QA round-1 feedback addressed (force-push).

**Critical 1 — tests don't assert log.warn was called.** Accepted. The reviewer is right that the tests previously proved "returns null" but not "warns the operator" — the named new behavior of the PR. Used the `spyOn(log, "warn")` pattern from `tool-schema-for-llm.test.ts` and added:

- **Positive assertions** on the two stray-key tests (URL unset + KEY set, with and without TID): warn called exactly once, message contains both the env-var-state narrative and `external-secret.yaml` cleanup guidance.
- **Negative assertions** on the clean-direct-mode tests: `expect(warnSpy).not.toHaveBeenCalled()` to lock in "warning only on stray key."
- **Negative on the fatal-direction tests** (URL set without key, missing tid): confirms the fatal path is mutually exclusive with the warn path — we don't accidentally fire both signals on the same misconfig.
- **New test** pinning the "warn fires only once across repeated getBouncerMode() calls" property explicitly. The behavior comes from the cache, but the test asserts it directly.

The spy uses `mockImplementation(() => {})` so the yellow warn text no longer pollutes test output either.

**Suggestion 2 — cache vs side-channel comment.** Accepted. Added a one-line note on `getBouncerMode()` explaining that "warn once at boot" is a side-effect of the cache, not a separate `hasWarned` flag. Pairs with the new cache-once test so a future maintainer who changes the cache logic sees both the comment AND the test.

**Suggestion 1 — vocabulary alignment.** Deferred per the reviewer's own framing. When `ONBOARD_TENANT.md` is updated post-merge to describe the one-step rollback, the warn message can be re-checked for terminology convergence.

Verify:
- `bun test test/unit/oauth-bouncer-config.test.ts` → 17 pass (was 16; +1 cache-once test)
- 32 expect() calls (was 19; +13 from warn assertions)

Ready for re-review.

PR #209, opened 2026-05-13
feat(composio): Gmail and Outlook via Composio aggregator (stopgap)

mgoldsborough:
## Summary

Adds a third connector auth kind — `auth: composio` — alongside `dcr` and `static`. Composio is a remote OAuth aggregator we use as a stopgap for toolkits where NimbleBrain doesn't yet hold restricted-scope verification (Gmail's `gmail.*` scopes, Outlook's `Mail.*`, etc.). Composio holds the vendor's tokens; the platform persists only an opaque `connectedAccountId` per workspace and routes tool calls through Composio's MCP endpoint.

Ships Gmail and Outlook as the first two Composio-backed connectors. Adding more is one catalog entry + one env var (no code change).

## Why a third auth kind

`dcr` and `static` both end with vendor OAuth tokens at the workspace credential store and the MCP server pulling them at runtime. Composio's model inverts the direction — tokens never leave Composio, tool calls flow through them. Trying to bend the existing kinds around this would have hidden Composio in `static` by leaving the secret in workspace.json or proxying through a custom MCPB bundle. Both worse than a discriminator.

## Architecture

- **`src/composio/sdk.ts`** — single SDK adapter. Owns every `@composio/core` import: session create with `sessionPreset: "direct_tools"` (so the toolkit's real tools appear on MCP, not Composio's meta-router), connected-account list/initiate/delete, eager `validateComposioConfig` (open-redirect mitigation on `COMPOSIO_API_BASE_URL`, multi-tenant `NB_TENANT_ID` enforcement when the bouncer is configured), 10-second timeout wrapper, canonical `composioUserId(wsId)` formula consumed by both routes and the install path.
- **`/v1/composio-auth/{initiate,callback,proxy}`** routes mirroring `mcp-auth.ts` in shape. `initiate` adopts an existing ACTIVE connected account when one exists (skips the redirect dance entirely) or kicks off a fresh flow with `allowMultiple: true`. `callback` verifies sha256-bound (nonce, connectorId, wsId) cookie + writes `connection.json` + transitions lifecycle state. `proxy` is the white-label forwarder for the vendor-side redirect URI.
- **`connection.json`** at `workspaces/<ws>/credentials/composio/<connectorId>/connection.json` — the user-presence signal for composio bundles. Parallel to `mcp-oauth/<server>/tokens.json` for native OAuth. 0o600 / 0o700 / atomic temp+rename — same security primitives as `workspace-credentials.ts`.
- **`BundleRef.composio.connectorId`** marker — carries the catalog id forward so boot-time state derivation and the catalog-by-URL→by-id fallback in `list_installed` can probe the right credential namespace and resolve the right metadata.
- **Curated `composio.tools` allowlist per catalog entry** — non-negotiable for huge toolkits. Gmail publishes 61 tools, Outlook 282. Without curation `nb__search` dumps every match's full description into context (~220k tokens for Outlook); gpt-5.5 hits its input cap and 400s, Sonnet burns 400k+ tokens per turn. Working sets are 15 tools each.

## Security

- API key held as a `\${COMPOSIO_API_KEY}` template on `BundleRef.transport.auth.value`, resolved at start time in `createRemoteTransport`. Secret never lands in workspace.json. Mirrors existing native `oauthClient.clientSecret: { ref: \"credential\", key }` pattern.
- Cookie-bound state nonce on `/callback`: sha256(nonce.connectorId.wsId), timing-safe compare, HttpOnly + SameSite=Lax + Path-scoped + Secure. Same posture as `mcp-auth.ts`.
- Eager startup validation throws on:
  - Non-http(s) `COMPOSIO_API_BASE_URL` (open-redirect mitigation on `/proxy`)
  - Bouncer mode active without `NB_TENANT_ID` (multi-tenant `user_id` collision risk)
- 10s timeout on every SDK call. Hanging Composio API can't block install/connect/disconnect indefinitely.
- `@composio/core` pinned to exact `0.9.1` (pre-1.0; minor upgrades may break).

## Deployment

Operators need (per environment):
- 1Password items: `composio/api-key`, `composio/auth-config-gmail`, `composio/auth-config-outlook` (and one per future toolkit)
- Composio dashboard: BYO Google / Microsoft OAuth client credentials wired into a Composio Auth Config per toolkit
- Egress to `backend.composio.dev` and `connect.composio.dev`

Operator runbook lives in the private ops repo, with the full step-by-step for adding new tenants and new toolkits.

## Test plan

- [x] `bun run verify` — 2,718 unit + 246 web + 485 integration + 17 smoke, 0 failures
- [x] 54 new unit tests: SDK config validation, multi-tenant safety, timeouts, allowlist behavior, session shape, callback nonce verification, proxy redirect, connection.json lifecycle
- [x] End-to-end in local dev: install Gmail → consent at Google → tool calls land → uninstall → reinstall Outlook → same flow
- [x] Reconnect after disconnect: deletes Composio account + local connection.json; next click forces a fresh OAuth dance
- [ ] Manual operator setup in staging (1Password items + dashboard auth configs) — runbook ready
- [ ] First production tenant cutover — gated on operator setup

## Follow-ups tracked

- **F1.** Per-tenant Composio sub-accounts (S4 long-term — currently single platform-wide API key; user_id namespacing isolates connected accounts but not enumeration)
- **F2.** Direct REST replacing @composio/core (S3 long-term — cuts ~11MB from image, removes pre-1.0 dep)
- **F3.** White-label proxy DNS automation (cosmetic — hides composio.dev flash in address bar)
- **F4.** `nb composio sync-tools <connector-id>` CLI to fetch and pick from a toolkit's tool list
- **F5.** Defensive `nb__search` truncation (would prevent recurrence of the 282-tools-into-context failure for any future big toolkit)
- **F6.** Per-toolkit usage telemetry to inform allowlist curation


mgoldsborough:
QA review addressed in 76c8cb1.

**Critical (all 3):**
- **C1** — Deleted the eager `recordConnectionStateChange(..., "running")` in the install path. `seedInstance` already returns the correct state from the `connection.json` probe; my override was masking the auth-required signal on a fresh install where the user cancelled the OAuth tab. Updated the inline comment to document why we *don't* force state there.
- **C2** — Added 8 route-level tests for `/v1/composio-auth/initiate` mirroring the `mcp-auth-routes.test.ts` pattern: happy-path cookie shape, adopt-existing short-circuit (asserts `alreadyConnected: true` + lifecycle transition + no fresh nonce cookie), missing API key (500), missing per-toolkit env (500), wrong auth kind (400), bad connectorId (400), unknown connector (404), non-JSON body (400). SDK mocked via `mock.module("@composio/core")` at the file head.
- **C3** — `stubCtx` now exposes a capturing `getLifecycle()` with `__lifecycleCalls.recordConnectionStateChange.lastCall`. The callback success test asserts the exact `(serverName, wsId, principalId, state)` tuple, so a refactor that drops the transition (currently masked by the route's `try/catch`) would fail loudly.

**Suggestions (all 5):**
- **S1+S2** — `remote-transport.ts`: dropped the double-pass; the for-loop already did what the explicit `resolveEnvTemplate` calls in the auth branches did. Documented the broad-scope intent in the docblock — `${ENV_VAR}` resolves on every outgoing header, not just auth values, so future connectors can use placeholders in arbitrary headers. New `remote-transport-env-substitution.test.ts` (7 tests) locks: substitution shape, unset → empty, case sensitivity (only `[A-Z_][A-Z0-9_]*` matches), literal `$` / `{...}` passthrough, broad scope.
- **S3** — Removed stale `serverIdEnv` from the fixture.
- **S4** — Tightened to `isValidConnectorId()` with explicit `..` and `//` substring rejection. Existing "rejects malformed cid" test rewritten to bind a valid cookie so it tests cid validation specifically (it had been passing for the wrong reason). New "rejects cid containing `//`" test added.
- **S5** — Cache comment on `_resetComposioConfigForTest` documents the production "set once at startup" contract — operators must restart to change `COMPOSIO_API_KEY` / `COMPOSIO_API_BASE_URL` / `NB_TENANT_ID`.

`bun run verify`: 2,725 unit + 246 web + 485 integration + 17 smoke, 0 failures.

Three design points where I considered three variations before picking:
1. **CID tightening** — class-regex + substring checks (picked) vs. negative-lookahead single regex vs. structural reverse-DNS regex. Picked the substring check for readability + defense-in-depth without locking us into a brittle id format.
2. **Lifecycle test stub** — full capturing mock vs. shape-capture last-call (picked) vs. boolean called/not. Picked shape-capture to match the reviewer's explicit ask ("assert state='running' and principalId='_workspace'").
3. **Env substitution scope** — single pass broad + tested (picked) vs. narrow to auth values vs. caller-level resolution refactor. Picked broad-scope-with-tests because the substitution logic is small and self-contained — keeping it at the transport boundary is a reasonable single-responsibility line, and the broad scope is useful for future header use cases.

mgoldsborough:
QA round 2 addressed in a31cb10.

**Critical (1/1):**
- **C1** — Extracted `cleanupComposioBundle({ workDir, wsId, connectorId })` as the canonical teardown recipe in `src/composio/sdk.ts`. Both `disconnect` and `uninstall` now call it; uninstall additionally rmSyncs the now-empty `composio/<connectorSlug>/` directory to match the mcp-oauth posture. Five tests pin the recipe: full success, upstream-fails-local-succeeds (idempotency), no-API-key (local-only, vs throw), nothing-to-clean (no-op), never-throws (best-effort contract).

**Suggestions (4/4):**
- **S1** — disconnect return shape is `{ access: upstreamDeleted }` only. Composio doesn't differentiate access/refresh; copying the same value into both fields was a lie.
- **S2** — secret-leak guard uses `v.replace(apiKey, "${COMPOSIO_API_KEY}")` when the API key appears anywhere in a header value (preserves prefixes like `Bearer <key>`). The exact-equal version only caught the unrealistic "header value === api key" case.
- **S3** — added a comment at the `composioClient` call site explaining that `validateComposioConfig()` is a cache read after the first (startup) call. Same architectural shape, clearer for the next reader.
- **S4** — new `ERROR_PAGE_CSP = default-src 'none'; frame-ancestors 'none'; base-uri 'none'` applied to both the OAuth-error branch and the session-mismatch branch on `/v1/composio-auth/callback`. Matches success-page posture; no inline styles needed here so tighter than the success policy.

**Two design points where I considered three variations:**

1. **Where the cleanup helper lives** — duplicate logic in two places (rejected: two-place truth), private method on `BundleLifecycleManager` (rejected: harder to unit-test), **exported function in `src/composio/sdk.ts`** (picked: the cleanup orchestrates upstream + local state, belongs in the composio adapter, fully testable via the existing SDK-mock pattern).

2. **Secret-leak guard breadth** — `v.includes(apiKey)` replacing the entire value (rejected: breaks `Bearer ` prefix), **`v.replace(apiKey, ${ENV})`** (picked: preserves prefix via the existing `resolveEnvTemplate` round-trip), drop the comment and keep equality (rejected: leaves a real defensive gap).

`bun run verify`: 2,743 unit + 246 web + 485 integration + 17 smoke, 0 failures.

mgoldsborough:
QA round 3 addressed in 985bb47.

**Critical (2/2):**

- **C1 (correctness)** — Reconnect after disconnect now restores the MCP source. New `BundleLifecycleManager.ensureSourceRegistered(serverName, wsId, workDir)` method: no-op when a source is already registered, otherwise reconstructs from the persisted `BundleRef` via `startBundleSource`. Called from both `/v1/composio-auth/initiate`'s adopt branch and `/v1/composio-auth/callback`'s success path before the existing `recordConnectionStateChange("running")`. Without this, post-disconnect-reconnect bundles reported `state: running` but had no registry entry — tool calls failed with "source not started" until platform restart.

- **C2 (test coverage)** — New `test/unit/connector-tools-composio-install.test.ts` (8 tests) covers the previously-uncovered ~80 LOC composio install branch in `handleInstallRemoteOAuth`. Asserts ref shape, transport.auth template, header scrubbing (with the secret repeated twice — validates `replaceAll` over `replace`), startBundleSource invocation (via the errResult shape that only fires when it throws), all three error branches, and the upstream SDK-failure path. The load-bearing assertion is `JSON.stringify(installed).includes("secret-api-key-DO-NOT-LEAK") === false` — proves the literal API key never reaches workspace.json.

**Suggestions (2/2):**

- **S1** — `v.replaceAll(apiKey, "\${COMPOSIO_API_KEY}")` instead of `replace`. Defense-in-depth against a future shape where the secret appears more than once in one header value.

- **S2** — Dropped `pendingConnectedAccountId` from the /initiate response shape, web client type, and tests. Was unused; dead code.

**S3 (PR-body F5 wording)** — Noted in this PR: `nb__search` defensive truncation IS the structural safety net for "future big toolkit ships without curation." It stays a follow-up (tracked in the original PR body) rather than blocking the current scope.

**Three-variation analysis on the design points where one was non-trivial:**

1. **C1 architectural fix** — Add `ensureSourceRegistered` helper called from reconnect paths (✅ picked: localizes the fix, keeps disconnect uniform across auth kinds). Considered: composio-specific "don't tear down on disconnect" (per-auth-kind branching, wrong UI signal for disconnected state) and "auto-heal in `recordConnectionStateChange`" (magical, couples concerns).

2. **C2 test placement** — New file `connector-tools-composio-install.test.ts` (✅ picked: isolates module-level mocks). Considered: `describe` block in the existing `connector-tools.test.ts` (cross-contamination via shared module cache), refactor `startBundleSource` to dependency injection (out-of-scope refactor).

3. **C2 startBundleSource mocking** — Initially mocked `bundles/startup.ts` at the module level; bun's `mock.module` leaked to other test files (`buildPlatformEnv` tests failed). **Resolution**: don't mock; let `startBundleSource` fail on the fake `composio.test` URL; assert state was persisted *before* the failure (which it is) and verify the call was reached via the errResult shape (`"Failed to start Composio MCP source"`). Achieves coverage without cross-file leakage.

`bun run verify`: 2,751 unit + 246 web + 485 integration + 17 smoke. **Zero failures.**

mgoldsborough:
QA round 4 addressed in e1a4717.

**Critical (3/3, all direct violations of the saved OSS no-leak rule):**

- **C1** — `"hq"` and `"  bayze  "` in `composioUserId` test fixtures replaced with `"tenant-a"` / `"  tenant-b  "`. The trim and prefix behaviors test the same code path with generic strings.
- **C2** — `https://hq.platform.nimblebrain.ai` (4 sites) replaced with `https://platform.example.test`. The internal multi-tenant subdomain pattern was disclosing our URL convention.
- **C3** — Both `nimblebrain-ops/research/COMPOSIO_SETUP.md` links dropped from CHANGELOG. They 404'd for OSS readers and disclosed the private-repo structure. When the operator runbook lands at `docs.nimblebrain.ai`, that link can replace the bare text — separate follow-up.

**Suggestions (3/3):**

- **S1** — Composio install now `errResult`s when `entry.defaultScope !== "workspace"`. The connection.json layout is workspace-scoped; a future `defaultScope: user` composio entry would silently bypass `disconnect`'s `isWorkspaceScope`-gated cleanup branch and orphan the upstream Composio account. Codifies the current architecture invariant.
- **S2** — `/proxy` reads from `validateComposioConfig().baseUrl` (validated + cached). Closes the gap where post-startup env mutation could bypass the http(s) check. Proxy tests now set `COMPOSIO_API_KEY` so the validated config honors the base-URL override.
- **S3** — The `NB_TENANT_ID` error message no longer references "the agent chart" / `.Values.tenant.id`. Generic "Set NB_TENANT_ID via the deployment env" — accurate for any orchestration substrate.

**Sweep:** ran `git diff main..HEAD` and grepped every PR-touched file for `bayze|acme|ipsdi|.platform.nimblebrain.ai|nimblebrain-ops|.Values.`. Only my own (now-fixed) introductions matched; the `acme` references in `workspace-credentials.test.ts` and `lifecycle.ts` are pre-existing fixtures (a generic placeholder convention, not my changes), and `static.nimblebrain.ai/icons/...` is the public asset CDN already used by every catalog entry. No remaining OSS leaks introduced by this PR.

**Two design points where the three-variation analysis was non-obvious:**

1. **S1 — workspace-scope enforcement.** Three options: drop the disconnect-side `isWorkspaceScope` gate (doesn't fix the underlying path-shape bug), enforce workspace at install (✅ picked — codifies invariant, single line), or refactor `connection.json` to scope-aware paths (out-of-scope refactor without current demand). Install-time enforcement is the least-disruptive fix that makes the latent footgun impossible.

2. **S2 — proxy base-URL source.** Three options: re-validate protocol on each request (extra work every hit), read from cached config (✅ picked — leverages existing validation, single line), or drop the override entirely (overreach). The cache-read carries the validate guarantee without re-running validation logic.

`bun run verify`: 2,751 unit + 246 web + 485 integration + 17 smoke. **Zero failures.**

mgoldsborough:
QA round 5 addressed in 70fb3b9.

**C1 — dedup is dead code for composio (real bug).**

Confirmed. `action.url` is the catalog placeholder (`https://backend.composio.dev/v3/mcp`); `ref.url = composioWiring?.url ?? action.url` is the per-install Composio session URL. The dedup at `connector-tools.ts:1182` was matching against `action.url`, so the orphan-recovery self-heal branch never fired for composio entries, and a duplicate-install click fell through to fresh-install and appended a second BundleRef plus orphaned the upstream Composio account.

But while testing the fix I hit a deeper issue: `createComposioSession` was being called *before* the dedup check, so even with corrected dedup, a re-click on an installed composio connector still burned a fresh upstream Composio session. The QA proposal (narrow dedup fix) doesn't address that.

**Three variations for the dedup itself:**

1. Branch on `action.auth === "composio"` and dedup on `b.composio?.connectorId === entry.id` for composio, `b.url === action.url` otherwise. Surgical but introduces a code path with two cases doing the same job.
2. **Uniform `b.serverName === serverName`.** serverName is the canonical lifecycle key (registry source name + workspace.json index), derived from entry.id and stable across installs. For non-composio bundles `action.url === ref.url` so this is strictly equivalent to the old URL match. ✅
3. New persisted `b.entryId` field. Cleanest type story but new persisted field for one regression.

Picked #2 + legacy URL fallback for records persisted before #195 (no `serverName` on disk — exercised by an existing test).

**Three variations for the session-create ordering:**

1. Narrow: fix dedup, leave session create where it is. Each duplicate click still burns a Composio session.
2. **Restructure so dedup runs before session create / static-credential read.** Cheap entry-shape and env validation up front; dedup next; expensive remote work only on fresh install. ✅
3. Make `composioWiring` a lazy thunk passed into the dedup branches. Adds indirection for no net win.

Picked #2. `buildComposioWiring()` and `loadStaticOAuthClient()` are now inline factories called from the dedup-miss path. `buildRef()` is also a function so the dedup branches can return without constructing a ref they'd discard. New test `(f)` asserts that hitting the self-heal branch invokes the composio SDK zero times.

**S1 — web type drift.** Added `tools?: string[]` to `ConnectorCatalogEntry.composio` and `DirectoryEntry.install.composio` in `web/src/api/client.ts`. The web shell doesn't read tools today; this is purity / parity with the server type so the next reader doesn't assume drift = absent on the wire.

**S2 — catalogById in a loop.** Added `ConnectorDirectory.catalogByIdMap()` sibling to `catalogByUrl()`. Replaced two per-bundle `await directory.catalogById(...)` calls in `handleListInstalled`'s installed-bundle loop with a single O(1) Map lookup. Catalog is ~30 entries so it's measurable-but-small; symmetry with the existing pattern is the real win.

`bun run verify`: 2,752 unit + 246 web + 479 integration + 17 smoke. **Zero failures.** New test `(f)` covers the previously-dead self-heal-for-composio path and asserts no second Composio session is created.

PR body updated to drop the `nimblebrain-ops/research/COMPOSIO_SETUP.md` private-repo path.

mgoldsborough:
S1 + S4 addressed in d811cb7. Skipped S2 (best-effort cleanup, QA agrees minor) and S3 (slugify collisions, near-zero practical risk on reverse-DNS catalog ids).

**S1 — install eager-start UX (`connector-tools.ts`).** Composio install eagerly calls `startBundleSource` so the tool list is populated before the user clicks Connect. If that eager start fails, the install itself has succeeded — workspace.json is written, seedInstance has run, and a Connect click runs the same `ensureSourceRegistered` path as a normal reconnect and recovers. Previously this returned `errResult("Failed to start Composio MCP source")` — the message said "install failed" while the bundle was in fact installed. Now: log + return success with a `warning` field on structuredContent and a human-readable note ("Installed ... — click Connect to retry"). Test (d) rewritten against the new contract.

**S4 — adopt-existing in `/initiate` (`composio-auth.ts`).** When `findActiveComposioConnection` matched an existing Composio account, the previous code wrote connection.json first and then tried `ensureSourceRegistered`; if source-register threw, it logged a warning and returned `alreadyConnected: true` anyway. The SPA showed the success page while the connector was actually `not_authenticated`. Reordered: `ensureSourceRegistered` runs first. If it throws, return 502 `composio_adopt_source_start_failed` and skip both `saveComposioConnection` and `recordConnectionStateChange`. No half-written state to reconcile, and the next retry runs a clean adopt-existing path with an honest error to the user. New test (b2) covers the failure path.

Both fixes share the same shape: align the on-disk state with what we tell the user. The previous code optimistically wrote first and hoped the side-effects would succeed — when they didn't, the UX lied. The fix is to make persistence transactionally follow the volatile step (source start), so a failure leaves nothing to undo.

**Skipped suggestions:**

- S2 (cleanupComposioBundle last-error overwrites first-error) — best-effort cleanup, only affects log output, not worth the error-accumulation plumbing.
- S3 (slugify collisions in dedup) — catalog IDs are reverse-DNS namespaced; a collision would require two entries like `com.foo/bar` and `com-foo/bar` to exist in the same catalog, which the curation process won't allow. A "hypothetically vulnerable" comment buys nothing.

`bun run verify`: 2,753 unit + 246 web + 479 integration + 17 smoke. **Zero failures.**

PR #221, opened 2026-05-14
fix(iframe-bridge): pin postMessage target origin + fix inline tool-result listener (#99, #220)

### LINKED ISSUE #99: Iframe bridge: tighten postMessage targetOrigin (currently *) across all Synapse bundles
## Background

Every Synapse bundle's iframe-hosted UI (the inline settings HTML at \`ui://<bundle>/settings\`, the main app UI, etc.) sends and receives JSON-RPC over the platform's iframe bridge using the well-known pattern:

\`\`\`js
window.parent.postMessage({ jsonrpc: "2.0", ... }, "*");

window.addEventListener("message", (e) => {
  // No e.origin / e.source check today
});
\`\`\`

Reference: \`synapse-collateral\`'s inline settings HTML, \`synapse-todo-board\`'s settings panel (#6), and other Synapse bundles all use this shape.

## The concern

- **Outbound (\`postMessage(..., "*")\`)**: any framing parent — including a malicious page that embeds the platform — receives the JSON-RPC payload. For tools that move sensitive data (saved instructions, contact info, document content), this is a small but real exfiltration surface.
- **Inbound (no \`e.origin\` / \`e.source\` validation)**: any window that holds a reference to the iframe can post messages that the bundle treats as legitimate JSON-RPC responses. Could be used to forge tool results.

This came up in QA review of #6 (synapse-todo-board custom instructions). Flagging as a platform-layer concern rather than fixing per-bundle, because:

- Every Synapse bundle uses the same shape today.
- Tightening per-bundle is whack-a-mole — any new bundle author reaches for the existing examples and inherits the gap.
- The right primitive is uniform: ship the sanctioned target origin to the iframe at handshake time and provide a helper bundle authors use everywhere.

## Proposed shape

1. **Synapse SDK exposes a sanctioned origin** at handshake. The platform sends it (along with theme tokens, etc.) when the iframe initializes. The SDK exposes it as \`hostContext.origin\` or similar.
2. **Bundle authors use a helper** instead of raw \`window.parent.postMessage\`. The helper passes the host origin as \`targetOrigin\`.
3. **Inbound listener validation**: SDK helper attaches a single message listener that checks \`e.source === window.parent\` and \`e.origin === hostContext.origin\` before dispatching to the bundle's handlers.
4. **Inline-HTML bundles** (the ones not using the React SDK) get a small helper script the platform can serve at \`/iframe-bridge.js\` — same pattern, reachable from inline HTML.

## Migration

- Existing bundles work as-is during the transition (loose mode).
- Once the SDK helper ships, update the canonical examples (\`synapse-hello\`, the inline-settings template) to use it.
- New bundles default to the helper; old bundles can migrate at their own pace.

## Out of scope

- Authentication / signing of messages — different concern.
- Cross-iframe communication beyond bridge JSON-RPC — different concern.

## Repos affected when this lands

- \`packages/synapse\` (SDK)
- \`products/nimblebrain/code\` (host side: send origin at handshake)
- All \`synapse-apps/*\` (migrate to helper at their own pace)

### LINKED ISSUE #220: home-briefing-inline listens for wrong tool-result method — never renders inline data
## Summary

The core resource `home-briefing-inline` (rendered by `InlineAppView` for inline tool-result widgets) registers a listener for the JSON-RPC method `synapse/tool-result`, but the host bridge dispatches `ui/notifications/tool-result`. The listener never fires, so `renderInline()` never runs in the inline path — the widget only renders if/when its own `callTool` round-trip happens to populate the DOM.

## Root Cause

`web/src/bridge/bridge.ts:582` — `bridge.sendToolResult(result)` emits:

```js
{ jsonrpc: "2.0", method: "ui/notifications/tool-result", params: { content, structuredContent } }
```

`src/tools/core-resources/scripts/home-briefing-inline.ts:74` (post #99 refactor; same mismatch existed prior with raw `addEventListener`):

```js
window.NBBridge.on("synapse/tool-result", function(msg) {
  var data = msg.params && msg.params.result;
  ...
});
```

The `synapse/tool-result` method is not defined anywhere in the host or the Synapse SDK — the SDK exposes `tool-result` as an SDK *event* mapped from the JSON-RPC method `ui/notifications/tool-result` (per `node_modules/@nimblebrain/synapse/README.md:195`).

## Impact

Inline tool-result widgets that depend on the host-pushed result payload (via `InlineAppView` → `bridge.sendToolResult`) never receive it through this listener. The bug pre-dates #99 and was preserved verbatim by the refactor; no behavior regression, but the script's data-injection path has been broken since it was introduced.

## Fix Options

1. **Listen for the correct method directly** — replace `synapse/tool-result` with `ui/notifications/tool-result` and read `msg.params.structuredContent` instead of `msg.params.result`.
2. **Use the Synapse SDK already wired up by `BRIDGE_HELPER`** — `synapse.on("tool-result", handler)` is the supported surface and survives future protocol-version drift.

Option 2 is preferred — it routes through the canonical SDK that already handles validation, schema versioning, and `hostContext` propagation.

## Steps to Reproduce

1. Trigger a tool call whose result UI renders via `ui://nb/home-briefing-inline`.
2. Observe that `renderInline()` never executes — only the placeholder/initial content shows.
3. (Verify the wiring) Log inside the listener — it never fires.

## Expected vs Actual

- **Expected:** inline widget renders briefing data the moment the host pushes the tool result.
- **Actual:** listener never fires; widget shows nothing data-driven from the inline path.

## Related

- #99 — postMessage targetOrigin hardening; touched this script but preserved the pre-existing method-name mismatch.

Ovaculos:
## Summary

Two related fixes in the iframe bridge layer.

- **#99 (platform-side):** bundle iframes can now safely post to the host without leaking JSON-RPC payloads to a hostile framing parent, and they can validate inbound messages from the host.
- **#220:** the inline home-briefing widget was listening for the undefined method `synapse/tool-result` reading the wrong field; rewired through the Synapse SDK so it actually receives host-pushed results.

## #99 — postMessage hardening

- **Outbound (iframe → parent)** is pinned to the host's real origin via a new `NBBridge` helper that captures `hostContext.origin` from the ext-apps handshake and uses it as `targetOrigin`.
- **Inbound (parent → iframe)** is validated against `event.source === window.parent` and the pinned host origin.
- **Host → iframe** stays `"*"` (srcdoc iframes have the opaque `null` origin; `postMessage` does not accept the literal `"null"` string). Tightening this requires the sandbox-proxy work tracked in `iframe.ts` and is explicitly out of scope here.

| Layer | Change |
|---|---|
| `web/src/bridge/bridge.ts` | Advertise `hostContext.origin = window.location.origin` in the ext-apps `ui/initialize` response. |
| `src/api/iframe-bridge-script.ts` (new) | `NBBridge` helper script: `send` / `on` / `off` / `getHostOrigin`. Handshake pins host origin only when claimed origin matches `event.origin`. Diagnostic `console.warn` on suspicious rejections; routine drops stay silent. |
| `src/api/routes/iframe-bridge.ts` (new) | `GET /iframe-bridge.js` — serves the helper. `Content-Type: application/javascript`, `Cache-Control: public, max-age=300`, `X-Content-Type-Options: nosniff`. Served unauthenticated (iframes can't reliably forward auth on a static script fetch). |
| `web/src/bridge/bridge.ts` + `web/src/hooks/useDataSync.ts` | Inline comments documenting why host → iframe `targetOrigin` stays `"*"` (null-origin constraint) so future contributors don't "fix" it by accident. |

## #220 — inline home-briefing tool-result listener

`home-briefing-inline.ts` listened for `synapse/tool-result` (a method the host never emits) and read `params.result` (the wrong field). The host's `bridge.sendToolResult` emits `ui/notifications/tool-result` with the payload on `params.structuredContent`. As a result the inline renderer never fired from the host-push path.

Rewired to the Synapse SDK's `App.on("tool-result", handler)` surface, which maps the JSON-RPC method to the short event name and delivers `{ content, structuredContent, raw }`. `autoResize: true` takes over the size notification, so the script no longer needs NBBridge or BRIDGE_HELPER setup — both come out and the script shrinks.

## What's out of scope

Per the #99 issue plan, these belong to other repos:

- `@nimblebrain/synapse` SDK helper consuming `hostContext.origin` — ships in the SDK's own release.
- Per-bundle migrations under `synapse-apps/*` — each bundle migrates to the SDK helper at its own pace.

## Tests

- `test/unit/api/iframe-bridge-route.test.ts` — route headers (incl. `nosniff`); helper behavior: queue-then-flush, pretender protection, source-mismatch drop, post-handshake origin-mismatch drop, `on` / `off` semantics, `console.warn` on suspicious rejections.
- `test/unit/core-resources/home-briefing-inline.test.ts` — static-shape pin: no stale `synapse/tool-result` literal, no `params.result` mis-read, `Synapse.connect` + `app.on("tool-result")` both present.
- `web/src/__tests__/bridge/bridge-target-origin.test.ts` — handshake response includes `hostContext.origin`; host → iframe target stays `"*"` (regression pin).
- `web/src/hooks/useDataSync.test.ts` — `data.changed` fan-out target origin stays `"*"` (regression pin).

`bun run verify:static`, `bun run test:unit` (2817 / 0), `bun run test:integration` (475 / 0) all green.

## Manual test plan

- [ ] `bun run build:bundles && bun run dev`
- [ ] `curl http://localhost:27247/iframe-bridge.js` → 200, JS body, `X-Content-Type-Options: nosniff`.
- [ ] Iframe Promise probe → `hostContext.origin` returns the platform URL.
- [ ] Trigger a tool-result-rendering flow → inline briefing widget populates from the host push instead of staying on the loading placeholder.

## Related

- Closes #99 (platform-side portion; SDK + bundle migrations follow separately in their own repos).
- Closes #220.


Ovaculos:
Sorry about the new issue for no reason. I misunderstood while working on this and thought it was an out-of-scope thing, but it's all fine in the same pull

PR #248, opened 2026-05-19
refactor(tools): shared output types for automations + drift guard

mgoldsborough:
## Summary

Follow-up to #245. Closes the architectural gap that drove five rounds of review on that PR: every consumer (CLI, integration tests, web client) redeclared handler response shapes inline with \`as { ... }\`, and the type system had no way to enforce that those redeclarations matched the handler. Each return-shape change drifted at least one surface (regex, schema description, CLI cast, integration test) — and shipped each time until reviewers caught it.

The repo already had the right architecture for *inputs*: TypeBox schemas in \`src/tools/platform/schemas/\`, derived static types via \`Static<typeof X>\`, codegen to web. This PR completes the other half — same pattern for *outputs*.

**Stacked on**: #245. Merge that one first.

### Named output types (\`schemas/automations.ts\`)

\`AutomationsListOutput\`, \`AutomationsStatusOutput\`, \`AutomationsRunOutput\` (discriminated union), \`AutomationsRunsOutput\`, \`AutomationsCancelOutput\` plus their building blocks (\`AutomationSummary\`, \`AutomationStatusDetail\`, \`AutomationRunRecord\`, etc.). Type-only exports — no TypeBox runtime schema for outputs, because the handler's TypeScript return type already constrains them.

### Compile-time drift guard (\`output-types-drift-guard.ts\`)

The schema types must be self-contained (no cross-tree imports of canonical \`Automation\` / \`AutomationRun\`) because the web codegen pins \`rootDir\` to \`schemas/\`. To prevent the canonical types and the schema mirrors from drifting, a small file in \`src/bundles/automations/src/\` uses type-level \`assignable<A, B>()\` constraints to bridge them. \`bun run check\` validates it as part of CI; any structural disagreement is a build failure.

Confirmed empirically: renaming a field on the schema side fails compile in the drift guard AND in the actual CLI consumer that dereferences the field. Both caught before merge.

### Typed handler returns

\`handleList\`, \`handleStatus\`, \`handleRuns\`, \`handleRun\`, \`handleCancel\` now declare named \`AutomationsXxxOutput\` returns instead of bare \`: object\`. This is the load-bearing change — TypeScript checks every return expression against the contract.

### Consumers migrated

- **CLI** (\`src/cli/commands/automation.ts\`): three inline \`as { ... }\` casts replaced with imports of the named types. The CLI bug from #245's round 4 (\`nb automation run\` crashing on the dispatched envelope) literally cannot happen anymore — handler and consumer share the same type.
- **Integration tests** (\`test/integration/automation-e2e.test.ts\`): \`expectSyncRun\` helper retyped against the imported union.
- **Unit tests** (\`test/unit/bundles/automations/server.test.ts\`): handleRun tests narrow via \`\"run\" in result\` / \`\"status\" in result\` rather than \`as { run }\`.
- **McpSource test driver** (\`test/unit/mcp-source-tasks.test.ts\`): stream-driver \`emit\` now takes a \`TaskStreamMessage\` union where the error variant is \`McpError\` (the real SDK class). Plain-object error mocks like \`{ message: \"...\" }\` — the kind that masked round-2's no-op regex — now fail to compile.

### Coding standard

New section §2.1 in \`src/tools/platform/AGENTS.md\` formalizes:
- Handler output types are named, shared, and the handler's return type.
- Inline \`as { ... }\` is an anti-pattern; consumers import the named type.
- Tests narrow discriminated unions via \`\"key\" in result\`, not \`as\`.
- SDK boundary tests must construct real production types (real \`McpError\`, not \`{ message }\` plain objects).
- Known gap documented: tests aren't typechecked by CI today (~1000 existing errors block the cleanup); the convention becomes a compile gate once that's fixed.

Plus four new anti-pattern table entries covering the specific drift classes #245 hit.

## What this prevents

The CLI bug from #245's round 4 literally cannot happen now. Future shape changes to \`handleRun\` force every consumer to handle every branch — or fail to compile. Type system enforces what review discipline kept missing.

## Testing

- [x] \`bun run verify:static\` clean (format, lint, typecheck, cycles, codegen — including the new drift guard)
- [x] 2971 unit tests pass
- [x] 5 integration tests pass
- [x] Empirical drift check: renamed a schema field → 4 compile errors across guard + CLI consumer; restored → clean
- [ ] Run \`nb automation run\` against staging (manual; not reachable from this env)

## Follow-up

- Extend the same pattern to other handler returns currently typed as \`object\` (skills, conversations, files, manage_tools, etc.). Each closes a consumer-drift class.
- Add \`test/**\` to the typecheck scope. ~1000 existing test type errors need cleanup first; once done, the test-side type tightening in this PR becomes load-bearing too.

### DIFF
--- a/src/bundles/automations/src/output-types-drift-guard.ts
+++ b/src/bundles/automations/src/output-types-drift-guard.ts
@@ -1,0 +1,107 @@
+/**
+ * Compile-time drift guard for automation output types.
+ *
+ * The output types in `src/tools/platform/schemas/automations.ts`
+ * (`AutomationSummary`, `AutomationStatusDetail`, `AutomationRunRecord`,
+ * etc.) are STRUCTURAL MIRRORS of the canonical `Automation` /
+ * `AutomationRun` / `ScheduleSpec` / `TokenBudget` types in `./types.ts`.
+ * They're duplicated because the schemas tree is self-contained for the
+ * web codegen (`scripts/codegen-web-platform-schemas.ts` pins `rootDir`
+ * to `schemas/`; cross-tree imports break the boundary).
+ *
+ * This module bridges the two sides at compile time. Each `assignable<A,
+ * B>()` declaration is a constraint that fails to type-check if A isn't
+ * assignable to B. `bun run check` (via `tsconfig.json`) validates this
+ * file as part of the standard CI gate — any drift between the canonical
+ * types and the schema types surfaces as a build failure here, not as a
+ * silent disagreement at consumer call sites.
+ *
+ * When you change `Automation` or `AutomationRun` (or their nested
+ * `ScheduleSpec` / `TokenBudget`) and this file fails to compile, the
+ * matching schema type in `schemas/automations.ts` needs an update.
+ * Treat it like a database migration — shape moves and consumers move
+ * together, not separately.
+ *
+ * This file has no runtime exports. The `void` calls below are the
+ * only emitted runtime work and exist solely to anchor the type-level
+ * constraints to a declaration TypeScript checks.
+ */
+
+import type {
+  AutomationRunRecord,
+  AutomationScheduleSpec,
+  AutomationStatusDetail,
+  AutomationSummary,
+  AutomationTokenBudget,
+} from "../../../tools/platform/schemas/automations.ts";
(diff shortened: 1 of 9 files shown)


mgoldsborough:
Re-targeted to main as #249 (auto-closed here when the stacked base branch was deleted post-#245-merge). Same diff, rebased onto current main.

Issue #254, opened 2026-05-20
Mid-turn conversation switch bleeds streaming response into destination chat


Ovaculos:
## Summary
When a user sends a message in one conversation and switches to another conversation before the response arrives, the streaming response continues to mutate the destination conversation's local state. The destination chat ends up displaying assistant deltas (and sometimes the source user message) from a different conversation. The user sees a turn that was never actually sent in the destination conversation.

The server-side persistence remains correct — the source conversation's file gets the full assistant turn and the auto-generated title, exactly as expected. The bleed is purely client-side UI state corruption.

## Steps To Reproduce
1. Open the app with the Conversations tab visible (sidebar + chat panel).
2. Start a new chat (or open an existing one — call it A).
3. Type a prompt in A and send it.
4. Before A's response finishes streaming, click a different conversation (B) in the list.
5. Watch B's chat panel.

## Expected Behavior
B's chat panel shows only B's persisted history. The in-flight streaming from A continues in the background against A's own state — navigating back to A shows the response still arriving. A's full turn — user message + assistant response + generated title — lands on disk in A's conversation file via `runtime.chat`, which completes on the server regardless of client navigation.

## Actual Behavior
B's chat panel renders B's history followed by A's streaming assistant deltas, attached to whatever B's last message happened to be. The assistant turn appears "ghosted" into B. Returning to A and back to B does not clear it until a full page reload, because the in-flight fetch keeps running and `flushToMessage` keeps rewriting `messages[messages.length - 1]` regardless of which conversation `useChat`'s state now points to.

## Root Cause
`useChat` (web/src/hooks/useChat.ts) is mounted by `ChatProvider` near the App root and persists across conversation switches. `sendMessage` starts a fetch via `streamChat` / `streamChatMultipart` and registers an `onEvent` callback whose `setMessages` calls always target the *current* state — there's no per-stream conversation guard.

When `loadConversation(B)` runs:
- It calls `setMessages(B's messages)`, replacing state.
- It does *not* isolate the in-flight stream from the new state.

The fetch's reader keeps delivering events. `text.delta` → `flushToMessage` → `setMessages(prev => updated[last] = {role:"assistant", content: cumulativeText, ...})`. Since `prev` is now B's array, B's last message gets overwritten.

## Impact
[...]

Issue #253, opened 2026-05-20
Auto-generated conversation titles often contain assistant response content instead of a short summary


Ovaculos:
## Summary
The `generateTitle` flow that names new conversations frequently returns the start of the assistant's response (truncated to ~30 tokens) instead of a 3–6 word summary. Affected conversations show titles like `"Here's something for you:\n\n---\n\nA man walks into a library..."` or `"# The Very Important Meeting of the Socks..."` in the Conversations list. Short, direct assistant responses tend to title correctly; creative or long responses (jokes, poems, stories, refusals) almost always misbehave.

This is a pre-existing bug. It was largely invisible until the #155 fix because new conversations did not appear in the list without a manual refresh.

## Steps To Reproduce
1. Start a new chat.
2. Send a prompt that draws a long or creative assistant response — e.g. `Write something funny`, `Write a little poem`, `Say something scary`.
3. Wait for the response to finish.
4. Check the Conversations list entry for that chat.

## Expected Behavior
The list entry shows a 3–6 word title summarizing the conversation (the system prompt asks for exactly that). Example: `Library Paranoia Joke`, `Moonlight Haiku`, `Socks in a Dryer`.

## Actual Behavior
The list entry shows the first ~30 tokens of the assistant's response, often with leading markdown (`# heading`, `---` divider, etc.) and cut off mid-sentence by the UI's 80-character truncation. Inspecting the conversation JSONL on disk confirms the `metadata.title` event was persisted with the assistant content as its `title` field — the generator returned bad text, the rest of the pipeline stored and displayed it faithfully.

## Root Cause
`generateTitle` (`src/conversation/auto-title.ts`) feeds the model this prompt:

```
[system]  "Generate a 3-6 word title for this conversation. Return only the title, nothing else."
[user]    "User: <userMessage>\nAssistant: <assistantResponse>"
```

with `maxOutputTokens: 30`.

The user-role message holds the entire transcript as a single string. Haiku pattern-matches this shape as a conversation it's expected to continue and predicts the next assistant token — i.e. extends `<assistantResponse>` — instead of obeying the system instruction. The longer or more creative the assistant content, the more the transcript-continuation signal outweighs the system instruction, so creative prompts (poems, jokes, stories) almost always fail while terse Q&A often succeeds.

## Impact
[...]

PR #255, opened 2026-05-20
fix(conversations): broadcast new chat to list live (#155)

### LINKED ISSUE #155: New chats do not appear in Conversations list until refresh or tab switch

Ovaculos:
## Summary

Fixes #155 — new conversations now appear in the Conversations list without a manual refresh.

Root cause: the runtime emitted `data.changed` onto the per-request chat sink, which never reaches the `/v1/events` SSE channel `useDataSync` consumes. The fix routes the emit through the runtime's default sink. Two broadcasts per new conversation — once right after the user message is persisted (label = first-message preview) and once after auto-title generation settles (label flips to the generated title).

## Follow-up issues filed during this work

Two adjacent bugs surfaced while wiring the live-update path. Both are pre-existing and out of scope for #155:

- #253 — Auto-generated titles often contain assistant response content instead of a short summary. Pre-existing prompt-shape bug in `generateTitle`; previously invisible because new conversations did not appear in the list without a refresh.
- #254 — Mid-turn conversation switch bleeds the streaming response into the destination chat. Pre-existing client-side state contamination in `useChat`.

## Commits

- `refactor(conversations): switch ConversationIndex to pull-on-demand` — replaces `fs.watch + 500ms debounce` with mtime-based reconciliation per call. Removes the timing-race class that the live broadcast would otherwise hit. `get(id)` keeps an `id → filePath` map so `handleGet` / `handleFork` / `handleUpdate` stat one file instead of walking.
- `fix(conversations): broadcast new chat to list live` — the routing fix itself, plus hardening that keeps the broadcast path correct end-to-end: `FaultIsolatedSink` so a logger throw can't abort the SSE-broadcast wrap, `AbortSignal.timeout` so a hung fast-model call can't leave the title-block `.finally` pending, dead chat-stream `data.changed` branch removed from `handlers.ts`.

## Known scope notes

- `data.changed` is `scope: "global"` in `SSE_ROUTES`. Pre-existing; a new conversation in workspace A causes every workspace-B iframe on `/v1/events` to refetch (each gets its own dir's data, so no leak — wasted bytes only). Worth a follow-up when the payload grows `wsId`.
- No broadcast on conversation resume. Deliberate — the conversation is already in the list. A "bump to top on resume" UX is a separate feature.

## Test plan

- [x] `bun run verify` — unit / web 249 / integration 482 / smoke 17, 0 fail, 0 errors
- [x] `ConversationIndex` reflects new + deleted + modified files immediately (no debounce)
- [x] Pre-turn + post-title broadcasts reach the default (SSE-bound) sink for new conversations; no broadcast on resume
- [x] Manual: new chat appears in list instantly with its first-message preview, then label flips to the generated title

### DIFF
--- a/src/api/handlers.ts
+++ b/src/api/handlers.ts
@@ -163,8 +163,7 @@ export async function handleChatStream(
           event.type === "tool.preparing.done" ||
           event.type === "tool.start" ||
           event.type === "tool.done" ||
-          event.type === "llm.done" ||
-          event.type === "data.changed"
+          event.type === "llm.done"
         ) {
           if (event.type === "chat.start") {
             broadcastUserMessageOnce();
--- a/src/bundles/conversations/src/index-cache.ts
+++ b/src/bundles/conversations/src/index-cache.ts
@@ -1,14 +1,14 @@
 /**
- * In-memory index of conversation metadata for fast listing, searching, and filtering.
+ * Pull-on-demand conversation index.
  *
- * Built on startup by scanning all JSONL file headers. Invalidated by fs.watch()
- * on the conversations directory (debounced 500ms).
- *
- * Types are defined locally — no imports from the runtime codebase.
+ * Each call reconciles in-memory state with the directory: stat every
+ * conversation file, re-read headers only for new or mtime-changed files,
+ * drop entries for files that have disappeared. There is no `fs.watch`, no
+ * debounce, and no broadcast-vs-debounce race — `list` and `get` always
+ * return disk-truth.
  */
 
-import { type FSWatcher, watch } from "node:fs";
-import { basename, join } from "node:path";
+import { statSync } from "node:fs";
 import { listConversationFiles, readConversationHeader } from "./jsonl-reader.ts";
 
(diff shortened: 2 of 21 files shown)


mgoldsborough:
Superseded by #302

PR #260, opened 2026-05-20
fix(tools): rank natural-language tool search

### LINKED ISSUE #33: nb__search matches literal prefixes, not natural-language tool queries
## Observation

From conversation `conv_30f049cdb75d464f` on ws_mat: an agent needed to find the todo-board tools and went through three searches before hitting any:

| Query | Result |
|---|---|
| `"todo task create"` | 0 results |
| `"todo board"` | 0 results |
| `"synapse-todo"` | 26 results |

Every attempt after the first was a wasted tool-call round-trip. The app name is `synapse-todo-board`; the tools are `create_board_task`, `batch_archive`, etc. Natural-language terms ("todo", "task", "create") appear in tool names and descriptions but `nb__search` didn't surface them.

## Behavior

`nb__search` appears to match on exact token or simple substring against tool names only, not against tool descriptions. Hyphenated app slugs (`synapse-todo-board`) aren't tokenized on `-`, so "todo" doesn't match.

## Why this matters

- **Visible cost**: each failed search is a tool-call round-trip — LLM output tokens + latency + context growth. Three searches ≈ 300+ output tokens on the example above.
- **Invisible cost**: agents that can't find a tool in 2-3 tries typically give up or fall back to a worse approach. Users see "the agent doesn't know how to do X" when really the agent didn't know *which tool to call*.

## Suggested

1. **Tokenize on `-` and `_`**: `synapse-todo-board` matches on `todo`, `board`, `synapse`.
2. **Match against descriptions, not just names**: the tool description for `create_board_task` starts with "Create a task on a specific board" — the query `"create task"` should hit it.
3. **Rank by query-term coverage**: a match on all query terms beats a match on one.
4. **Fuzzy / edit-distance fallback** (optional): handle `"todo boards"` → matches `"todo-board"`.

A simple BM25 or tf-idf over `name + description` per tool would cover (2) and (3) with ~20 LOC and no new deps.

## Severity

Medium. Every agent-driven workflow involving tool discovery pays this cost. Affects UX more than correctness.

## Source

conv_30f049cdb75d464f, ws_mat workspace, lines 9-22 in the JSONL log.

dom-mp:
## Summary
Fixes #33.
`nb__search` previously used literal substring matching over the full query, so natural-language queries like `todo task create` failed unless that exact phrase appeared contiguously in a tool name or description.
This PR adds deterministic tokenized ranking for tool discovery:
- Tokenizes tool source/name/description on separators like `-`, `_`, and spaces.
- Matches multi-term queries across source names, tool names, and descriptions.
- Ranks fuller query-term coverage above partial matches.
- Preserves empty-query browse behavior.
- Reuses the same ranked search for invalid-tool-name suggestions.
This allows queries like `todo task create` and `todo board` to discover `synapse-todo-board__create_board_task`.
## Testing
- `bun test test/unit/system-tools.test.ts`
- `bun run check`
- `bun run format:check`
- `bun run lint`

### DIFF
--- a/src/tools/registry.ts
+++ b/src/tools/registry.ts
@@ -3,6 +3,7 @@ import { textContent } from "../engine/content-helpers.ts";
 import type { ToolCall, ToolResult, ToolRouter, ToolSchema } from "../engine/types.ts";
 import type { PermissionStore } from "../permissions/permission-store.ts";
 import type { McpSource } from "./mcp-source.ts";
+import { rankToolSearchResults } from "./search-ranking.ts";
 import type { Tool, ToolSource } from "./types.ts";
 
 /**
@@ -233,12 +234,10 @@ export class ToolRegistry implements ToolRouter {
     return source.execute(localName, call.input, signal);
   }
 
-  /** Search all tools by keyword (substring match on name + description). */
+  /** Search all tools by natural-language terms over name + description. */
   private async searchTools(query: string): Promise<Array<{ name: string; description: string }>> {
-    const q = query.toLowerCase();
     const all = await this.availableTools();
-    return all
-      .filter((t) => t.name.toLowerCase().includes(q) || t.description.toLowerCase().includes(q))
+    return rankToolSearchResults(all, query)
       .slice(0, 5)
       .map((t) => ({ name: t.name, description: t.description }));
   }
--- a/src/tools/search-ranking.ts
+++ b/src/tools/search-ranking.ts
@@ -15,6 +15,12 @@ function tokenize(value: string): string[] {
     .filter(Boolean);
 }
 
+// Best-effort plural folding: a token longer than 3 chars ending in "s" also
+// matches its singular ("boards" → "board"). Applied symmetrically to both
+// query and corpus tokens, so an over-stripped junk variant ("status" →
+// "statu") only matches if some other real token stems to the same string —
+// which doesn't occur in the tool corpus. Intentionally naive; a real stemmer
+// isn't worth a dependency for discovery ranking.
 function tokenVariants(token: string): string[] {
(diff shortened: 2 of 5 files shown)


mgoldsborough:
Thanks for picking this up! The tokenization + multi-term matching is the right shape for the bug, and the new tests cover the three motivating queries from the issue cleanly. Two small things I'd want to address before merging — both are narrow fixes, not asking for a rework.

**1. Single-word prefix queries regress.** The old substring filter matched `"greet"` against tool `"greeting"`; the new ranker tokenizes both into atomic terms (`{"greeting"} ∌ "greet"`), so `matchedTerms === 0` and the tool gets dropped at `search-ranking.ts:72`. The `name.includes(normalizedQuery)` bonus a few lines up is actually dead — it raises the score, but the `if (matchedTerms === 0) continue` gate drops the tool before the score is ever read. This regresses common short queries: `"auth"`→`authenticate_user`, `"config"`→`configure`, `"doc"`→`documents`. One-line fix:

```ts
if (matchedTerms === 0 && !name.includes(normalizedQuery) && !description.includes(normalizedQuery)) continue;
```

Worth a regression test (`query: "greet"` finds `"greeting"`).

**2. The system-tools handler is uncapped.** `registry.ts:236` already does `.slice(0, 5)` on the suggestion path, but `system-tools.ts:163` returns every match. Multi-term queries that the old filter would have returned 0 for can now return 50+ tools, each with full description. Suggest capping at 25 and adjusting the message: `Found N tool(s) (showing top 25)`.

Tests, typecheck, and lint all pass locally. Nice clean module separation — having `search-ranking.ts` as a pure function makes both fixes one-liners.

PR #289, opened 2026-05-26
fix: soft-delete users instead of hard-deleting the identity

mgoldsborough:
## Why

A user deleted via Settings → Users → trash (the \`manage_users delete\` tool) was **hard-deleted from WorkOS** while the platform's local state was left untouched. Two bad consequences, both seen in production:

1. **Deleted users still appeared in the UI.** The user list and workspace member lists read the local \`UserStore\` / \`workspace.json\`, which the delete never touched. WorkOS forgot them; the platform didn't.
2. **Re-adding a deleted user broke all their access.** WorkOS mints a brand-new immutable user ID for the re-created account. Every prior workspace membership was keyed to the old ID, so the returning user could log in but saw none of their workspaces.

Root cause: \`handleDelete\` called only \`provider.deleteUser()\` → \`workos.userManagement.deleteUser()\`. No cascade to the local profile, workspace memberships, or personal workspace; and the WorkOS hard-delete is irreversible.

## What changed

\`delete\` is now a **soft delete** — deactivate, don't destroy:

- **`UserStore.softDelete` / `restore`** — stamp/clear a \`deletedAt\` tombstone (idempotent; \`restore\` drops the key entirely from the persisted JSON). The hard \`delete\` stays for migrations / the OIDC directory-owner path.
- **Access gate** — \`WorkosIdentityProvider.resolveUser\` and \`OidcIdentityProvider.verifyRequest\` deny any request whose local profile carries \`deletedAt\`. Covers both the web shell and \`/mcp\` (both route through \`verifyRequest\`). The WorkOS gate sits next to the existing "no org membership = deny" check and invalidates the identity cache so revocation is immediate.
- **`manage_users`** — \`delete\` → soft delete (no longer calls \`provider.deleteUser\`); new \`restore\` action; \`list\` surfaces \`deletedAt\` for deactivated users. Last-owner guard now ignores already-deactivated owners.
- **UsersTab** — deactivated users stay listed with a "Deactivated" badge and a restore action; the delete button becomes a restore button; confirm copy updated (it's reversible now).

The provider identity is **preserved** across deactivate/restore, so the returning-user-loses-their-workspaces failure mode can't happen.

## Tests

- \`test/unit/identity/user-store.test.ts\` — softDelete/restore (tombstone, retention, idempotency, key-drop, not-found).
- \`test/unit/tools/user-tools.test.ts\` — delete is soft + retains record, does **not** call \`provider.deleteUser\`, restore round-trip, list surfaces \`deletedAt\`.
- \`test/unit/identity/workos-soft-delete.test.ts\` — admit active user → soft-delete → \`verifyRequest\` denies → restore → admitted again.

\`bun run verify:static\`, \`test:unit\` (3119), \`test:web\` (368), and \`test:integration\` (618) all green.

## Note

This does not auto-clean records orphaned by the *old* hard-delete behavior; that was handled as a one-off operational fix. New deletes are reversible from here on.


mgoldsborough:
## Review adjudication (commit 6b2a6cb)

Thanks for the thorough pass. Resolution of each finding:

**Critical #1 — asymmetric last-owner guard — fixed.** Extracted `activeOwnerCount(users)` and used it on **both** the update (demote) and delete (deactivate) paths, so they can't drift again. Each guard also now fires only when the *target* is an active owner, so it's the precise invariant ("never leave zero active owners") rather than a count-only heuristic. Regression test added: two owners, one deactivated → demoting the sole active owner is still blocked.

**Critical #2 — deactivated users shown as normal workspace members — fixed.** `handleListMembers` now surfaces `deletedAt`, and `WorkspaceMembersTab` renders the same "Deactivated" badge + row opacity as the org Users tab. Members are **not** removed from `workspace.json` (retaining membership is what makes restore clean). Test added asserting `list_members` surfaces `deletedAt`. On `memberCount`: left as total membership (the rows are members, just deactivated); the per-row badge carries the distinction, consistent with the display.

**Beyond the list — same defect class you flagged, found at the workspace level.** The workspace last-admin guards (`handleRemoveMember` line 559, `handleUpdateMember` line 638) had the identical "counts deactivated principals" bug. Left unfixed, deactivating a workspace's admin and then removing the only other active admin strands the workspace with zero functional admins. Added `activeAdminCount(members, userStore)` + a target-active check on both paths. Tests added: can't remove/demote the last *active* admin when the co-admin is deactivated; **can** remove an already-deactivated admin (it can't reduce the active count). This is exactly the asymmetric-defense smell — fixing one guard and leaving its sibling — so I closed the whole class.

**Suggestion #1 — fixed.** `handleDelete` catch → `Failed to deactivate user:`.

**Suggestion #2 — adjudicated, not applied as proposed.** Dropping the `?.` on `this.userStore` would be type-unsafe: the field is `UserStore | null` (the WorkOS provider legitimately runs store-less, and `workos-stale-cache.test.ts` constructs it that way). A null store means soft-delete isn't even possible — no profiles exist to carry `deletedAt` — so the gate correctly no-ops rather than silently admitting a deactivated user. Kept the chain and added a comment making that reasoning explicit.

Re cache staleness: agreed and confirmed — per-tenant agent runs `platform.replicas == 1`, and the soft-delete handler calls `invalidateUser()` on that single instance, so revocation is immediate.

Green: `verify:static`, unit 3124, web 368, integration 618.

PR #311, opened 2026-05-27
fix(oauth): surface (don't swallow) interactive-auth start failures

mgoldsborough:
## Problem

A remote-OAuth connector (granola on prod/hq) hangs in **"Connecting…"** forever — no `tokens.json`, **no error log**, the connector page shows `McpSource not started`. We confirmed everything upstream is correct after the prior fixes: deploy is live (`0f44a4b`), workspace is right (#309), DCR client is clean with the bouncer redirect (#310), and the bouncer delivers a valid authorization code (the user even sees the **"You're in"** success page). Yet tokens never land.

## Root cause of the silence

In `lifecycle.startAuth`, the interactive path works like this:

1. `source.start()` runs in the background, hits `UnauthorizedError`, fires `onInteractiveAuthRequired` (captures the auth URL, records `pending_auth`, resolves the URL back to the caller), then parks in `awaitPendingFlow()`.
2. When the user returns, the code wakes it → `finishAuth(code)` (token exchange) + reconnect.

If step 2 throws (exchange rejected, reconnect failed, or the pending flow hit its 15-min TTL), the failure lands in `startAuth`'s `.catch` — which **only did anything `if (!capturedAuthUrl)`**. On the interactive path `capturedAuthUrl` is set, so the branch was the silently-dropped `else`: **no state change, no log, stuck in `pending_auth`.** That's why this has been invisible — the real exchange/reconnect error is computed and discarded.

## Fix

- **Surface it.** The `.catch` now **always logs** the failure and transitions the connection to `dead` (+ `lastError`), so the UI offers a recoverable **Reconnect** instead of an indefinite spinner — and the underlying error finally appears in the logs.
- **Scope `providerAbort.abort()` to the give-up path.** It previously fired in a `finally` on the success path too, aborting a provider whose interactive flow is still pending in the background. The token exchange uses the SDK's own fetch (not this signal), so this wasn't the cause of the hang, but aborting a still-pending provider is wrong on its face.

## Why no regression test in this PR

The interactive-failure path can't be unit-tested with the current harness — `lifecycle.startAuth` builds the `McpSource` + provider internally, and the existing lifecycle tests deliberately avoid driving a live source (the OAuth-retry integration harness drives `McpSource` directly via the headless 302 path, not `startAuth`'s interactive branch). Driving `startAuth` through an interactive-then-failing exchange needs a new mock-AS harness at the lifecycle layer. Tracked as a fast-follow alongside the actual exchange-failure fix.

## How this gets us to the real fix

This is a diagnostic-enabling + silent-hang UX fix. After deploy, one reproduce will print the real error — `[lifecycle] startAuth: ai-granola-mcp start failed for _workspace in ws_nimblebrain_shared: <CAUSE>` — and the connector will show a Reconnect instead of hanging. Then we fix the underlying exchange/reconnect cause directly.

## Verification

`bun run verify:static` exit 0 (format, lint, `tsc` ×2, cycles, all guard checks). `bun run test:unit` 3141 pass / 0 fail. Related integration suites (lifecycle-startauth-disconnect, workspace-oauth-provider, workspace-oauth-provider-abort, mcp-source-oauth-retry) pass.

### DIFF
--- a/src/bundles/lifecycle.ts
+++ b/src/bundles/lifecycle.ts
@@ -1178,13 +1178,24 @@ export class BundleLifecycleManager {
       })
       .catch((err) => {
         const msg = err instanceof Error ? err.message : String(err);
-        // For UnauthorizedError, the callback path already recorded
-        // pending_auth — we don't want to overwrite that with `dead`.
-        // Other errors (network, SSRF block, server crash) → record dead.
+        // Always surface the failure. The interactive path (capturedAuthUrl
+        // set) used to be swallowed here: if the background start() failed
+        // AFTER the auth URL was returned — the token exchange or reconnect
+        // threw once the user came back, or the pending flow timed out — the
+        // connection was left stuck in `pending_auth` ("Connecting…") forever
+        // with no log and no tokens. Log it and move the connection to `dead`
+        // (+ lastError) so the UI offers a recoverable Reconnect instead of
+        // an indefinite spinner.
+        log.warn(
+          `[lifecycle] startAuth: ${serverName} start failed for ${principalId} in ${wsId}: ${msg}`,
+        );
+        this.recordConnectionStateChange(serverName, wsId, principalId, "dead", {
+          lastError: msg,
+        });
+        // `authUrlPromise` already resolved on the interactive path, so a
+        // reject there is a no-op; only the headless / pre-auth failure path
+        // (no captured URL) still needs the caller's promise rejected.
         if (!capturedAuthUrl) {
-          this.recordConnectionStateChange(serverName, wsId, principalId, "dead", {
-            lastError: msg,
-          });
           rejectAuthUrl(err instanceof Error ? err : new Error(msg));
         }
       });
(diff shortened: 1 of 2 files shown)


mgoldsborough:
## QA adjudication

**Critical (no automated test) — agreed, resolved.** You were right that "fast-follow, no issue" was too weak for a Critical. Rather than file an issue, I wrote the integration test you described — it turned out to be tractable against the existing mock-AS pattern:

- `test/integration/lifecycle-startauth-interactive-failure.test.ts` drives the **real** `lifecycle.startAuth` against a `Bun.serve` mock AS: `/authorize` returns a 200 login page (so the provider's redirect-probe falls through to the **interactive** branch), then `/token` returns 400 `invalid_grant` on resume. It asserts the Connection ends in `dead` with `lastError` set — i.e. surfaced, not stuck in `pending_auth`.

Writing it **caught a real defect in the fix**: the SDK's OAuth error classes (`InvalidGrantError`, …) carry their detail in `.name` with an **empty `.message`**, so my `err.message`-only extraction produced a *blank* diagnostic — nearly as useless as the swallow it replaced. Fixed: fall back to `err.name`, so the log + `lastError` now read e.g. `InvalidGrantError` instead of `""`. The test pins `lastError === "InvalidGrantError"`. (This also directly sharpens the prod hunt — granola's failure will now surface with the exact SDK error name.)

**Suggestion (parallel `startup.ts:449` boot path) — agreed, filed as #312.** Confirmed it's not a fold-in: `startBundleSource` doesn't own the connection state machine (`recordConnectionStateChange` is a `BundleLifecycleManager` method; the boot path is fire-and-forget from `workspace-runtime` with no lifecycle handle in scope), so the `dead` transition needs its own plumbing. #312 carries the acceptance criteria including the `err.name` extraction this PR established.

`verify:static` exit 0, `test:unit` 3141/0, new integration test green.

PR #316, opened 2026-05-28
fix(oauth): make WorkspaceOAuthProvider safe under concurrent SDK auth() calls

mgoldsborough:
## Summary

Roots out the `{\"error\":\"invalid_code\"}` failure on every interactive Connect for bouncer-mode tenants (Granola, Notion, Dropbox).

The MCP SDK's `StreamableHTTPClientTransport` opens a connection with multiple parallel HTTP requests (POST initialize, POST initialized notification, GET for SSE stream). When OAuth is required, each 401 independently invokes `auth()` on the same provider — 2-3 concurrent invocations per logical Connect.

`WorkspaceOAuthProvider` was written as if `auth()` is called serially. Every state-mutating method (`state()`, `saveCodeVerifier()`, `saveClientInformation()`, `redirectToAuthorization()`) blindly overwrote prior state with no concurrency check. N concurrent chains produced N different `(state, verifier, client_id, URL)` tuples; disk held whichever wrote last; the user opened whichever URL returned first; vendor bound the code to one tuple but exchange POSTed another. `invalid_code`. Every. Single. Connect.

## What changes

Make the provider idempotent within an in-flight flow via sync (no-await-between-check-and-claim) coalesce points:

- **`state()`** — returns existing state if a flow is in progress, so all chains join one flow
- **`saveCodeVerifier(v)`** — first writer wins; subsequent concurrent calls no-op
- **`saveClientInformation(info)` / `clientInformation()`** — DCR coalesce via `Deferred<...>`. First DCR's result is returned to all concurrent callers; subsequent calls skip their own DCR
- **`redirectToAuthorization(url)`** — claims `urlCaptured` AND verifies the URL's `code_challenge == SHA256(claimed verifier)`. Couples the two races so the captured URL matches disk verifier — what makes the exchange succeed.

JS single-threading makes the check-then-claim atomic. No locks.

Combined with the lifecycle-level coalesce (in `lifecycle.ts`, `authFlowsInFlight` Map bounded by terminal connection-state transitions), the end-to-end invariant is: **AT MOST ONE OAuth flow per `(serverName, wsId, principalId)` at a time, with ONE coherent `(state, verifier, client_id, challenge, captured URL)` tuple, regardless of how many concurrent SDK auth() invocations or lifecycle.startAuth calls fire.**

## Commit walkthrough

1. **`fix(oauth): make WorkspaceOAuthProvider safe under concurrent SDK auth() calls`** — provider + lifecycle coalesce points
2. **`test(oauth): pin concurrent-auth() coalesce contract for WorkspaceOAuthProvider`** — 6 unit tests pinning the invariant (state, verifier, client_id, URL all coherent across N concurrent chains)
3. **`diag(oauth): env-gated AUTHORIZE URL logging for token-exchange correlation`** — permanent diagnostic (one line, gated on `NB_DEBUG_OAUTH_EXCHANGE`) so future regressions can be diagnosed in one log pair with the existing token-exchange logger

## Verification

- ✅ `verify:static` exit 0 (format / lint / tsc / 11 path/cycle/codegen guards)
- ✅ `test:unit` — 3153 pass / 0 fail
- ✅ `test:integration` — 648 pass / 0 fail / 6 skip
- ✅ New `workspace-oauth-provider-concurrency.test.ts` — 6 pass / 0 fail
- ✅ Validated end-to-end against production tenant-hq with Granola — full diagnostic trace showed `not_authenticated → starting → pending_auth → running` cleanly, TOKEN POST returns 200, tokens land

## Out of scope (separate issues)

- Web UI multi-fire on Install (4-7 `manage_connectors` calls per click) — the server now tolerates it, but the frontend defect is real
- Per-flow disk storage for verifier/client — structural alternative, unnecessary now that the invariant holds upstream

## Test plan

- [x] Unit + integration green
- [x] End-to-end manual verification on tenant-hq prod
- [ ] One more clean Install round on a fresh test connector post-merge to confirm no regression


mgoldsborough:
Addressed QA review. Four new commits pushed:

| Commit | QA item | What |
|---|---|---|
| `4dfb426` | **Critical #2** | Subscribe `dcrInFlight` deferred to `abortSignal` — releases awaiters on abort so a first-caller DCR failure can't deadlock a second concurrent SDK auth() chain on the same provider. CAS guards against late-firing abort overwriting a saveClientInformation success. Two new tests pin the contract. |
| `a36f953` | **Critical #1** | New `test/unit/lifecycle-startauth-coalesce.test.ts` (5 tests) pinning the lifecycle-level invariants: concurrent coalesce (verified by spying on `startAuthInner` invocation count), `.catch` fallback clears on sync failure, slot release on every terminal state, slot held across `starting` / `pending_auth`, CAS doesn't clear a reassigned slot. |
| `a945dce` | **Suggestion 1** | PKCE coupling check gated on `code_challenge_method === "S256"` — defensive against a future `plain` flow falsely failing the SHA256 derivation. |
| `4d6e569` | **Suggestion 2** | Truncate `state` to 8 chars in AUTHORIZE diagnostic log — matches existing convention on the next log line. Reduces correlation surface for operators reading logs. |

Skipped suggestion 3 (per-flow `isFirstSave` flag): the current ambient `!this.cachedClientInfo` check is correct for actual usage — the lifecycle constructs a fresh provider per `startAuth`, and `invalidateCredentials("client")` (which is what would run mid-flow) explicitly clears `cachedClientInfo` AND releases `dcrInFlight` so the next DCR is a genuine fresh start. Per-flow tracking here would break the SDK's `InvalidClientError` retry path, not improve it. Happy to revisit if there's a concrete scenario I'm not seeing.

Verification: `verify:static` exit 0 · `test:unit` 3160 pass / 0 fail (+7 new) · `test:integration` 648 pass / 0 fail · all 8 provider concurrency tests + 5 lifecycle coalesce tests green.

PR #329, opened 2026-05-29
Automations: Task primitive + V3 reader + mobile pass

mgoldsborough:
## What

Turns the automations bundle from a cron UI into an automation product. Three layered changes:

1. **Task is now a kernel primitive.** `runtime.executeTask()` sits beside `runtime.chat()` as the unattended sibling for scheduled work. Composes the system prompt with a new `mode: "task"` branch that prepends `TASK_IDENTITY` — so the model produces a finished deliverable instead of a chat reply. The runtime owns this framing; bundles can't spoof it by wrapping user messages.
2. **The automations bundle calls `executeTask` instead of `chat`.** The Daily Briefing no longer opens with "Sure, let me pull that together for you now" or closes with "Want me to dig deeper?" — the contract forbids both. Five places in the UI that had been sending the old flat `{ name, prompt, schedule, ... }` shape now correctly wrap as `{ manifest, body }` (Create / Test Run / Pause / Resume / inline-edit were all broken before).
3. **The UI is a two-pane reader.** Rail on the left (automations + run history), reader on the right with the briefing rendered as actual markdown. The 500-char output cap is gone. Templates render as proper stacked cards (no more overlap). Mobile collapses cleanly with a real back affordance and `userOpenedReader` decoupling so auto-selection doesn't hijack navigation.

## Why

Three things were wrong before:
- The Daily Briefing was returning chat-shaped output ("Would you like me to do any of those?") because automations called `runtime.chat()`, the runtime composed the conversational system prompt, the model responded conversationally. The fix is at the kernel: invoke a different primitive.
- The reader pane didn't exist. Run output was a 500-char monospace `<pre>` truncated at write time, buried behind a chevron expand. The whole payoff of an automation had no home.
- Mobile responsiveness, the create-form chips, and the schema mismatch on the create tool were all real bugs surfaced during a UX audit and a frontend-design audit done in this session.

## Architecture — kernel/extension boundary

Reviewed in-session by the chief-architect lens. Verdict: **Task is a kernel primitive.** It's about *how* the agent runs (an invocation mode), not *what* it does. Same engine loop, two contracts:

\`\`\`
┌─────────────────────────────────────────────────────────────────┐
│                consumers (bundles + clients)                    │
│  chat UI / REST   │   automations    │   evals    │   webhooks  │
│       chatFn ⤴       executeTaskFn ⤴       ⤴            ⤴       │
└──────────┬─────────────────┬──────────────────┬─────────────────┘
           │                 │                  │
           ▼                 ▼                  ▼
┌─────────────────────────────────────────────────────────────────┐
│                          runtime                                │
│   chat(req): ChatResult        executeTask(req): TaskResult     │
│   ─ conv history loading        ─ fresh conversation per call   │
│   ─ SSE streaming               ─ optional EventSink (obs.)     │
│   ─ content parts / files       ─ ownerId → identity hydrate    │
│   ─ in-flight lock              ─ no continuation semantics     │
│   ─ mode = "chat"               ─ mode = "task"                 │
│                                                                  │
│              composeSystemPrompt({ mode, ... })                 │
│              ├─ mode === "chat" → DEFAULT_IDENTITY              │
│              └─ mode === "task" → TASK_IDENTITY                 │
│                                                                  │
│              engine.run() — pure agent loop, unchanged          │
└─────────────────────────────────────────────────────────────────┘
\`\`\`

### Optional `workspaceId` semantics

- **Set** → focused: active tools are that workspace's tools + identity tools; briefing for that workspace; system prompt frames the agent as operating in it.
- **Omitted** → active tools are the owner's personal-workspace tools + identity tools; no focused-workspace briefing. The full cross-workspace tool union is the *discoverable corpus* reached on demand via \`nb__search\`, not preloaded into the active set (progressive disclosure — same shape as chat at the identity-level home). Layer 3 bundle workflow guidance *does* aggregate across every workspace the owner can see, so a discovered tool's usage skill is available when it lands. \`TASK_IDENTITY\` carries the framing.

The no-focus behavior intentionally mirrors chat's identity-level home so the two siblings stay semantically consistent. *(Corrected from an earlier draft of this PR body that overstated this as "cross-workspace reach"; see #329 QA round 3.)*

## Deferred — substrate extraction

Documented inline in a NOTE block on \`runtime.executeTask()\`. The clean refactor — extracting a shared \`_agentInvoke()\` substrate that both \`chat()\` and \`executeTask()\` sit on — was deliberately deferred to a follow-up PR. Doing it in this PR would have touched ~500 LOC of the most-trafficked code path in the runtime in the same diff as the UI work, multiplying regression risk for chat. The current \`executeTask\` is honest about its duplication; the next PR decomposes both methods together.

## Commits

1. \`feat(prompt): task mode in composeSystemPrompt with TASK_IDENTITY\`
2. \`feat(runtime): add executeTask primitive for unattended agent invocation\`
3. \`refactor(automations): wire executor through executeTask, drop 500-char cap\`
4. \`feat(automations-ui): V3 reader, mobile pass, chip fix, schema shape fix\`

## Verification

- \`bun run verify:static\` — clean (format, lint, types, 9 path-check scripts, codegen)
- \`bun run test:unit\` — 3143 pass, 0 fail
- \`bun run test:integration\` — 642 pass, 12 skip, 0 fail
- \`bun run test:web\` — 369 pass, 0 fail
- \`bun run build:bundles\` — automations bundle 316 kB / 94 kB gzipped (was 311 kB; +18 kB for marked + dompurify)

6 new unit tests cover task-mode compose (precedence, fallback suppression, contract phrases). The existing 14 executor tests were rewritten against the \`TaskFn\` shape and all pass.

## Test plan for review

1. Open the bundle on a wider viewport — two-pane reader with rail + reader.
2. Resize below 720 px — rail takes full width; tap a run; reader takes over with a back arrow top-left.
3. Click Create → templates render as stacked cards, no overlap.
4. Pick Daily Briefing → Test Run. Output reads like a briefing (heading, sections, lists), not a chat reply.
5. Click "Open conversation →" on a run; lands on the conversation in the chat shell.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,14 @@ jobs:
         run: bun install
       - name: Install web dependencies
         run: cd web && bun install
+      # Bundle UI deps must be installed so unit tests that import from
+      # src/bundles/*/ui/src/ (e.g. the markdown sanitizer test) can
+      # resolve their npm dependencies (marked, dompurify, etc.). Mirrors
+      # the install loop inside build:bundles. Local `bun run build:bundles`
+      # installs these as a side effect; CI doesn't build bundles for the
+      # unit job, so the install must be explicit.
+      - name: Install bundle UI dependencies
+        run: bun run install:bundles
       - name: Verify (unit + web tests)
         run: bun run verify:test-unit
 
--- a/bun.lock
+++ b/bun.lock
@@ -36,6 +36,7 @@
         "@types/bun": "latest",
         "@types/react": "^19.2.14",
         "@types/react-dom": "^19.2.3",
+        "happy-dom": "^20.0.0",
       },
       "peerDependencies": {
         "typescript": "^5",
@@ -113,6 +114,10 @@
     "@types/react": ["@types/react@19.2.14", "", { "dependencies": { "csstype": "^3.2.2" } }, "sha512-ilcTH/UniCkMdtexkoCN0bI7pMcJDvmQFPvuPvmEaYA/NSfFTAgdUSLAoVjaRJm7+6PvcM+q1zYOwS4wTYMF9w=="],
 
     "@types/react-dom": ["@types/react-dom@19.2.3", "", { "peerDependencies": { "@types/react": "^19.2.0" } }, "sha512-jp2L/eY6fn+KgVVQAOqYItbF0VY/YApe5Mz2F0aykSO8gx31bYCZyvSeYxCHKvzHG5eZjc+zyaS5BrBWya2+kQ=="],
+
+    "@types/whatwg-mimetype": ["@types/whatwg-mimetype@3.0.2", "", {}, "sha512-c2AKvDT8ToxLIOUlN51gTiHXflsfIFisS4pO7pDPoKouJCESkhZnEy623gwP9laCy5lnLDAw1vAzu2vM2YLOrA=="],
+
+    "@types/ws": ["@types/ws@8.18.1", "", { "dependencies": { "@types/node": "*" } }, "sha512-ThVF6DCVhA8kUGy+aazFQ4kXQ7E1Ty7A3ypFOe0IcJV8O/M511G99AW24irKrW56Wt44yG9+ij8FaqoBGkuBXg=="],
 
     "@vercel/oidc": ["@vercel/oidc@3.1.0", "", {}, "sha512-Fw28YZpRnA3cAHHDlkt7xQHiJ0fcL+NRcIqsocZQUSmbzeIKRpwttJjik5ZGanXP+vlA4SbTg+AbA3bP363l+w=="],
 
@@ -214,6 +219,8 @@
 
     "encodeurl": ["encodeurl@2.0.0", "", {}, "sha512-Q0n9HRi4m6JuGIV1eFlmvJB7ZEVxu93IrMyiMsGC0lrMJMWzRgx6WGquyfQgZVb31vhGgXnfmPNNXmxnOkRBrg=="],
 
+    "entities": ["entities@7.0.1", "", {}, "sha512-TWrgLOFUQTH994YUyl1yT4uyavY5nNB5muff+RtWaqNVCAK408b5ZnnbNAUEWLTCpum9w6arT70i1XdQ4UeOPA=="],
+
     "environment": ["environment@1.1.0", "", {}, "sha512-xUtoPkMggbz0MPyPiIWr1Kp4aeWJjDZ6SMvURhimjdZgsRuDplF5/s9hcgGhyXMhs+6vpnuoiZ2kFiu3FMnS8Q=="],
 
     "es-define-property": ["es-define-property@1.0.1", "", {}, "sha512-e3nRfgfUZ4rNGL232gUgX06QNyyez04KdjFrF+LTRoOXmrOgFKDg4BCdsjW8EnT69eqdYGmRpJwiPVYNrCaW3g=="],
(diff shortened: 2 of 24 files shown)


mgoldsborough:
QA fixups pushed (3 commits on top of the original 4):

- `fix(runtime): honor request.maxInputTokens in executeTask` — wires the field through so per-automation token caps actually take effect. Comment flags the parallel gap in chat() as a separate scoped follow-up.
- `chore(automations-ui): allowlist-based markdown sanitizer; drop deprecated @types shim` — switched DOMPurify from forbidden-list to default-deny ALLOWED_TAGS / ALLOWED_ATTR / ALLOWED_URI_REGEXP. The reviewer's call on `<object>`/`<embed>` was the right read: under the html profile + FORBID_TAGS those slipped through depending on context. Allowlist makes that impossible. Also dropped @types/dompurify — DOMPurify 3.x ships its own types.
- `test: cover executeTask end-to-end and markdown sanitizer contract` — the critical-fix the reviewer asked for: 6 integration tests for executeTask (fresh conversation, deliverable persistence, focused vs cross-workspace, source: 'task' metadata stamp) and 6 unit tests pinning the sanitizer's removals. Pattern mirrors `ambient-context-cross-workspace.test.ts`; needed happy-dom as a root devDep for DOMPurify's import-time DOM. The integration test is the safety net for the deferred substrate extraction — chat() and executeTask() can't diverge silently anymore.

Verification across the whole branch:
- `bun run test:unit` — 3149 pass, 0 fail (+6 from the QA fixups)
- `bun run test:integration` — 660 pass, 12 skip, 0 fail (+6 from the QA fixups)
- `bun run test:web` — 369 pass, 0 fail

Adjudication: all four QA items addressed. The `<object>`/`<embed>` happy-dom limitation is documented in the test comment so the coverage stays honest — the real browser DOM (where the bundle actually runs) enforces the allowlist faithfully.

mgoldsborough:
Second-round QA adjudicated:

| Item | Call | Action |
|---|---|---|
| Watch-item (substrate duplication) | Agree on tracking | Filed #334 with full extraction boundaries (substrate vs chat-specific vs task-specific). Linked from the NOTE block on `executeTask` |
| Suggestion 1 (chat `maxInputTokens` parity) | Agree, separate PR | Filed #335 with the one-line scope. Comment in `runtime.ts` updated to reference it |
| Suggestion 2 (raw markdown in RunRow + AutomationDetailView) | **Fixed in this PR** | Both surfaces now use `renderMarkdown` and `.out-md` for consistency with ReaderPane. Error blocks stay as `<pre>` (stack traces) |
| Suggestion 3 (`legacyTruncated` false-positive on length===500) | Wontfix | True edge case. Cost of misfire is one extra "older build" note. Better heuristic possible but not worth the LOC |
| Suggestion 4 (no e2e test of `as TaskRequest` cast bridge) | Wontfix | Reviewer themselves verified the cast is structurally safe and risk is low. `executor.test.ts` covers the bundle side, `execute-task.test.ts` covers the runtime side; the cast is the seam between |

New commit: `71af0e8 feat(automations-ui): render markdown in RunRow + AutomationDetailView; link follow-ups`

Branch state: 8 commits on top of main. Verification still clean (3149 unit / 660 integration / 369 web, 0 fail; `verify:static` clean).

mgoldsborough:
Round 3 QA adjudicated:

| Item | Call | Action |
|---|---|---|
| **Critical 1** — `TaskRequest.workspaceId` docstring contradicts code | **Agreed, fixed** | Code uses `toolsWsId = focusedWsId ?? sessionWsId` and `toolsRegistry.availableTools()` for that single workspace + identity tools — same shape as chat. The docstring claimed cross-workspace tool aggregation. Updated the type-level and field docstrings to describe the real progressive-disclosure pattern (active = personal-workspace tools + identity tools; cross-workspace = discoverable via `nb__search`). Also corrected the misleading integration-test name ("cross-workspace tools are reachable" → "the orchestrator dispatches a namespaced cross-workspace tool call") and updated the PR body's "Optional workspaceId semantics" section to match. **No code change** — the behavior is sensible and matches chat. Commit: `c517bec` |
| **Suggestion 1** — silent error swallowing in rail handlers | **Acknowledged, filed #339** | Reviewer correctly noted these are pre-existing on main (6 catches there too, not introduced by this PR). Tracking as a follow-up to keep this PR's scope coherent |
| **Suggestion 2** — `maxInputTokens` asymmetry | **Already tracked as #335** | No new action; #335 will close the gap on the chat side |

Verification after the doc fix:
- `bun run test:unit` — 3149 pass, 0 fail
- `bun run test:integration` — 6 pass on the updated execute-task suite (and full integration suite stays green)

On the test-rename: this was a real find. My `cross-workspace tools are reachable` test passed only because the echo model is scripted to emit a namespaced tool call by name — a real LLM would not see that tool in its active list with `workspaceId` omitted. The test still pins something real (orchestrator dispatch), and the comment now says so explicitly. The discoverability story (`nb__search`) is what would actually surface cross-workspace tools to a live model.

Branch state: 10 commits. CI is recovering from the earlier bundle-deps install fix; this push should land clean since static + unit + integration all pass locally.

mgoldsborough:
Heads up — a second pass of this same review just came in pinned at `71af0e8`. That SHA is two commits behind the current tip `c517bec`. Every finding was already addressed:

- **Critical 1** (`TaskRequest.workspaceId` docstring vs code) — fixed in `c517bec` (this comment is what announced it: https://github.com/NimbleBrainInc/nimblebrain/pull/329#issuecomment-4580983956). Both docstrings rewritten, integration test renamed, PR body's semantics block rewritten.
- **Suggestion 1** (rail-handler silent catches) — tracked as #339. Pre-existing on main; scoped out of this PR.
- **Suggestion 2** (`maxInputTokens` asymmetry) — tracked as #335.

If you can re-pull and re-verify against the current tip you should get the green-on-everything outcome the original adjudication described.

mgoldsborough:
Round 4 adjudicated. The reviewer this round refreshed against the current tip — appreciate that.

| Item | Call | Action |
|---|---|---|
| **No criticals; #334 carrying-cost note** | Acknowledged | The reviewer accepts the deferral and pins the operational risk on 'don't let #334 slip.' Agreed |
| **Suggestion 1** — biome `noUselessEscapeInRegex` on `\-` in `RailItem.tsx:66` | **Fixed in this PR** | One-char delete. `bun run lint` clean after. Commit: `fe90e78` |
| **Suggestion 2** — `resultPreview` field name now misleading | **Filed #340** | Reviewer themselves said 'future rename is the cleaner end state' — not asking for it here. And it's migration-level work (existing JSONL run records on disk are written with that field), so it doesn't belong bundled with the UI/runtime work. #340 captures scope + 3 migration options + the operator runbook step |

Branch state: 11 commits. Verification still clean (3149 unit / 660 integration / 369 web; `verify:static` clean). CI re-running.

Worth saying: this round caught two things I'd want a reviewer to catch — a real biome rule violation (info-level, but a real signal) and a naming-debt landmine that will get worse the longer we wait. Both went into proper homes (fixed here, or tracked with full scope). The pattern is paying off.

PR #331, opened 2026-05-29
fix(tools): refresh discovery union when a source comes online

mgoldsborough:
## Problem

Found auditing prod conversation `conv_f09ff1a4548042a5` (tenant-hq): the agent burned 59 tool calls / 154s thrashing on `nb__search` and never found the tools it needed. `nb__status` reported `synapse-collateral` healthy with 31 tools, but `nb__search` returned **zero** of them across every workspace, and `nb__manage_tools` rejected every promotion as *"not found in the current tool registry."* `files__read` (a kernel identity tool the user was explicitly told to use) was absent too.

## Root cause

The cross-workspace tool union — the `nb__search` corpus and the set `nb__manage_tools` / the engine's `allToolSchemaMap` validate against — is a per-identity **memoized projection** of each workspace's `ToolRegistry`. Its only invalidation channels were a `workspace.json` fs.watch + a membership-set change. Both are **disjoint from the real mutation surface**:

- A bundle subprocess becoming enumerable (slow cold start, **HealthMonitor crash-restart** — which reuses the same source object and never re-enters `addSource`, deferred/pending-auth start) changes a source's tools **without touching `workspace.json`**.
- So a union computed while the source was unreachable is memoized and served **stale for the process lifetime**. `nb__status` reads the live source (healthy); the union doesn't.

Identity half: `getIdentityDescriptors` memoized its first result, clearing only on rejection. An empty read at boot (in-process identity sources not yet enumerable) stranded `files__read`/`conversations__*`/`automations__*` out of **every** session permanently.

## Fix — make source readiness the trigger, via MCP-native notifications

- **`McpSource`** registers the client handler for the server's native `notifications/tools/list_changed` (we ignored it before — even a *dynamic* tool change went stale) and emits a tools-changed signal at `start()`'s single success seam, which also covers `tryRestart()` and deferred/pending-auth starts. New optional `subscribeToolsChanged()` on `ToolSource`.
- **`ToolRegistry`** bridges both source-set membership changes (`add`/`removeSource`) and per-source readiness transitions to one invalidation listener; `SharedSourceRef` forwards the subscription to the shared inner source.
- **`Runtime`** wires each per-workspace registry's listener to `aggregator.invalidateWorkspace(wsId)` at the boot and JIT seams — after boot population so the boot-time `addSource` storm doesn't thrash.
- **aggregator** exposes `invalidateWorkspace` (per-workspace cache drop; does not touch the membership stamp).
- **identity descriptors** no longer memoize an empty list as terminal — retry-while-empty is bounded (kernel identity sources always exist, so empty is always transient; stops the instant a non-empty list resolves).

The `workspace.json` fs.watch stays as a backstop for out-of-process edits.

## Validation

This fix was produced via the triage-first bug-fix workflow including an **independent adversarial review**, which corrected the initial mechanism (it's source-*readiness* transitions, not source-set membership — the `addSource`-only hook would have missed the HealthMonitor restart path) and rejected an unsafe unbounded identity retry.

- `bun run test:unit` — 3206 pass
- `bun run test:integration` — 656 pass (incl. aggregator watcher + mcp-source OAuth-retry rebuild path)
- New regression tests: aggregator re-lists after `invalidateWorkspace` once a source is online; no-op when nothing cached; identity descriptors retry-while-empty then memoize; registry bridges add/remove + readiness emits and detaches cleanly on removal.

## Follow-up (tracked, not in this PR)

`tool-list-cache.ts`'s header claims `workspace.json` is "the canonical persistence target for bundle writes" — false for the lifecycle install path (`atomicConfigAdd/Remove` writes `configPath`) and all respawn/restart paths. Now that invalidation is reactive, either correct the comment to "external-edit backstop" or remove the fs.watch entirely. Deferred until the reactive channel soaks in prod.

### DIFF
--- a/src/orchestrator/tool-list-aggregator.ts
+++ b/src/orchestrator/tool-list-aggregator.ts
@@ -109,6 +109,17 @@ export interface ToolListAggregator {
    * `aggregateToolList` call to discover the drift.
    */
   invalidateIdentity(identityId: string): void;
+  /**
+   * Drop the cached tool list for one workspace and every identity union that
+   * read from it. The reactive counterpart to {@link invalidateIdentity}:
+   * that handles membership changes (a workspace left the identity's set);
+   * this handles tool-set changes WITHIN a workspace that the
+   * `workspace.json` watcher can't see — a bundle subprocess (re)connecting,
+   * a deferred/pending-auth start completing, or a native
+   * `tools/list_changed`. Wired from `ToolRegistry.setInvalidationListener`.
+   * Does NOT touch the membership stamp (membership is unchanged).
+   */
+  invalidateWorkspace(wsId: string): void;
   /** Number of active per-workspace watchers — for the leak test. */
   activeWatcherCount(): number;
   /** Close every watcher, clear every cache. Idempotent. */
@@ -156,15 +167,15 @@ export function createToolListAggregator(options: ToolListAggregatorOptions): To
    * Bare descriptors for the kernel identity sources, listed once. Identity
    * tools are static (code-defined, not per-tenant install), so unlike the
    * per-workspace cache they need no FS-watch invalidation — list on first
-   * use, memoize, reuse for every identity. A failed listing drops the memo
-   * so the next call retries.
+   * use, memoize, reuse for every identity. A failed OR empty listing drops
+   * the memo so the next call retries (see below).
    */
   let identityDescriptorsPromise: Promise<readonly NamespacedToolDescriptor[]> | null = null;
   const getIdentityDescriptors = (): Promise<readonly NamespacedToolDescriptor[]> => {
     const lister = options.listIdentityTools;
     if (!lister) return Promise.resolve([]);
     if (!identityDescriptorsPromise) {
-      identityDescriptorsPromise = lister().then((tools) =>
+      const p = lister().then((tools) =>
         // Bare: `name === toolName`, `wsId === null`. The orchestrator routes
         // a bare `<source>__<tool>` through the identity door.
         tools.map((t) => ({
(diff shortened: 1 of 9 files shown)


mgoldsborough:
## Addressed QA Critical finding (commit 2dcdf7c)

**Critical — OAuth-retry seam skipped `emitToolsChanged()`: confirmed and fixed.** You're right — `start()` has two success seams, and the headless OAuth auto-resolve retry returns early (after `connectWithTimeout` succeeds) before the bottom-seam emit. A remote OAuth source becoming enumerable via that retry fired no readiness signal, leaving the exact stale-union bug live on that branch. My "single success seam" PR claim was wrong.

- Added `this.emitToolsChanged();` on the retry success path before the early `return;`, and corrected the comment.
- Extended `test/integration/mcp-source-oauth-retry.test.ts` to assert **exactly one** readiness emit on the retry path — this would have been `0` before the fix, so it's a genuine regression test for the branch you flagged.

**Pre-existing asymmetry — tracked, not fixed here (matching your scoping).** As you noted, that same early-return also skips `dead = false`, the remote `onclose` rewiring, `instructions` capture, and `startTaskSweeper()`. That's pre-existing on `main` and the OAuth path has subtle single-use-transport / crash-detection state, so converging the retry path onto the bottom seam is its own change with its own risk surface — left as a documented follow-up (in-code comment + will file an issue) rather than smuggled into this PR.

**Suggestion — identity retry-while-empty bound: acknowledged, intentionally left as-is.** Valid in the abstract. In practice it's bounded: the kernel identity sources are code-defined and always present, so a non-empty list resolves within the first call or two of boot and is then memoized permanently — steady state is one memoized promise, not a re-list loop. The only way to hit unbounded per-request re-listing is a deployment that *wires* `listIdentityTools` yet has it return `[]` durably, which the contract already says should instead omit the lister entirely. I'd rather keep the invariant "never strand identity tools" than add a TTL/counter that could re-introduce staleness if sources are slow. Happy to add a short-window bound if you feel strongly.

Green after the change: `verify:static` clean, affected unit + integration tests pass (20/20 across the three affected files; full suites were green pre-change and nothing here widens scope).

PR #338, opened 2026-05-30
fix(skills): report per-request Layer-3 skills in status scope:skills

mgoldsborough:
## Problem

`nb__status scope:skills` reported only the boot-time platform/core skills, even when workspace- and user-tier Layer-3 skills (`loading_strategy: always`) had been composed into the prompt. Surfaced in a shared-workspace report: the agent's prompt received 7 workspace skills, but `status:skills` showed only the 5 platform defaults.

## Root cause

Two paths read skills from two sources:

- **Prompt composition** (`chat()`) loads the per-request Layer-3 pool fresh — org + workspace + user tier (`loadConversationSkills`) plus bundle `skill://<name>/usage` skills — then runs `selectLayer3Skills`.
- **`handleSkillStatus`** read a boot-time static cache (`getContextSkills()` / `getMatchableSkills()`), which never contains per-request workspace/user-tier skills.

So the status surface and the prompt could disagree — the reporting half of the same divergence class as the tool-discovery cache bug (#331).

## Fix

- Extract pool construction + selection into `Runtime.selectRequestLayer3(...)`, now the **single source of truth** for both `chat()` composition and status reporting — the two surfaces can't drift again.
- Add `Runtime.describeRequestSkills(wsId)`, which reports through that same path (active-tool input = the focused workspace's available tools).
- `handleSkillStatus` now renders the composed **Workspace & User Skills (always loaded)** and **Tool-Affined Skills (active)** sections alongside the existing core / context / matchable sections, deduped by name.

## Scope

Reporting only. Prompt composition behavior is unchanged (`chat()` calls the extracted method with identical inputs). Does not address tool-affined skills failing to load when their bundle's tools are down — that is the separate tool-discovery cache fix in #331.

## Tests

- `workspace-tier-skills.test.ts`: asserts the focused workspace's `always` skill is reported by `describeRequestSkills` (the status surface), not just `skills.loaded`.
- `system-tools.test.ts`: new case for rendering per-request Layer-3 workspace skills; existing skill-status cases updated for the `describeRequestSkills` seam.
- `bun run verify` green.

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -83,7 +83,7 @@ import {
   partitionSkills,
 } from "../skills/loader.ts";
 import { SkillMatcher } from "../skills/matcher.ts";
-import { selectLayer3Skills } from "../skills/select.ts";
+import { type SelectedSkill, selectLayer3Skills } from "../skills/select.ts";
 import { approxTokens } from "../skills/tokens.ts";
 import { truncateMarkdownToBudget } from "../skills/truncate.ts";
 import type { Skill } from "../skills/types.ts";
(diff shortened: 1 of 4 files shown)


mgoldsborough:
Applied review suggestions #1 and #2 in 589d198..896dbc6:

- **#1 (asymmetric dedup):** made **Core authoritative** — core-marker skills are now excluded from the Layer-3 (`always`/`tool_affined`) sections, so a core skill that ever carried a loading strategy lists only under *Core Skills (immutable)*, never twice. (Chose this direction over deduping core *against* layer3, since Core is the more meaningful category.)
- **#2 (coverage gap):** added a unit test asserting a skill present in both boot context and the per-request Layer-3 set renders exactly once.

**Declining #3 (sort sections by name):** intentionally not doing this. Status section order currently mirrors the prompt's composition order — the property that makes `status:skills` a faithful view of what the agent actually received. Sorting by name would break that mirror in exchange for byte-stability, which isn't worth it for a diagnostic surface. It's also pre-existing FS-order behavior, not introduced here. `bun run verify` green.

PR #341, opened 2026-05-30
fix(skills): keep core skills out of Layer-3 status sections; test dedup

mgoldsborough:
Follow-up to #338 (merged) — applies the two QA-review suggestions that didn't make it into the squash before merge.

## Changes

1. **Core is authoritative in `status:skills`.** Core-marker skills are excluded from the Layer-3 (`always` / `tool_affined`) sections, so a core skill that ever carried a Layer-3 loading strategy lists only under *Core Skills (immutable)* — never twice. Removes the dedup asymmetry between the core and `userContext` branches the reviewer flagged.
2. **Test the dedup path.** New unit case: a skill present in both boot context and the per-request Layer-3 set renders exactly once (under Layer-3, not also under User Context).

## Not done (deliberately)

The reviewer's third suggestion — sort each section by name for byte-stable output — is intentionally declined. Status section order currently mirrors the prompt's composition order, which is what makes `status:skills` a faithful view of what the agent received; sorting would break that mirror for marginal byte-stability on a diagnostic surface. Pre-existing FS-order behavior, not introduced here.

## Tests

`bun run verify` green. Display-only change; no behavior change to prompt composition.

### DIFF
--- a/src/tools/system-tools.ts
+++ b/src/tools/system-tools.ts
@@ -471,8 +471,12 @@ async function handleSkillStatus(
   const userContext = context.filter(
     (s) => !s.sourcePath.includes(CORE_SKILL_MARKER) && !layer3Names.has(s.manifest.name),
   );
-  const alwaysLoaded = layer3.filter((s) => s.loadedBy === "always");
-  const toolAffined = layer3.filter((s) => s.loadedBy === "tool_affinity");
+  // A core skill is shown only under "Core Skills (immutable)", the more
+  // meaningful category. Should one ever carry a Layer-3 loading strategy,
+  // drop it from the Layer-3 sections so it can't list twice — Core wins.
+  const layer3Visible = layer3.filter((s) => !s.skill.sourcePath?.includes(CORE_SKILL_MARKER));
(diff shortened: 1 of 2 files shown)


mgoldsborough:
Applied both review suggestions — at the root rather than as polish:

- **#1 (marker fragility → real regression):** confirmed reachable. `loadScopedSkills` recurses (`loader.ts:138`), so a workspace/user skill under a `core/` subfolder gets a `/skills/core/` sourcePath; the old substring filter would drop it from the Layer-3 sections and — not being in boot context — hide it entirely. **Fixed by deduping against Core by *name*** (`coreNames`), not by path marker: a Layer-3 entry is suppressed only when a same-named skill already renders under *Core Skills*. Keeps Core authoritative, mirrors the name-based `userContext` dedup, can't hide a distinct skill.
- **#2 (optional-chaining inconsistency):** obviated — the name-based dedup no longer touches `sourcePath` on Layer-3 entries, so the redundant `?.` is gone.

Added a unit test for the guard (a core skill that also enters the Layer-3 set renders once, under Core). `bun run verify` green.

PR #348, opened 2026-05-30
fix(conversations): broadcast new chat to list live (#155)

### LINKED ISSUE #155: New chats do not appear in Conversations list until refresh or tab switch

mgoldsborough:
## Root cause

A brand-new conversation didn't appear in the Conversations list until the user refreshed or switched tabs (#155). The runtime emitted the new-conversation `data.changed` signal onto the **per-request chat sink**, which never reaches the `/v1/events` SSE channel that `useDataSync` consumes. This routes the emit through the runtime's **default sink** instead (the one `api/server.ts` wraps for `/v1/events`).

It fires twice for a new conversation:
- right after the user message is persisted → conversation surfaces with its first-message-preview label
- after auto-title generation settles → label flips to the generated title

No broadcast on resume (the pre-turn emit is guarded by `if (!request.conversationId)`, and the title block only runs when `conversation.title === null`).

## Hardening

- `FaultIsolatedSink` wraps each sink in the `defaultEvents` fan-out so a logger throw can't abort the SSE broadcast fan-out.
- `AbortSignal.timeout` in `generateTitle` hard-caps a hung fast-model call so the title block's `.finally` always fires (the list re-pull happens even if title generation/persist fails).
- Removed a dead chat-stream `data.changed` branch in `handlers.ts` (nothing relays it).

## Note

This is a rebased + reduced version of #255. The conversation-index `pull-on-demand` refactor from that PR is deferred to a separate follow-up — it needs mtime+size staleness handling, an ownership-gating caller audit, and a search/stats limit cap.

## Test plan

- `bun run verify` green for everything touching the changed files; the only red is a pre-existing 15s-timeout flake in `body-limit-routes.test.ts` (a multipart-upload HTTP test) that reproduces identically on clean `origin/main` and is unrelated to this change.
- `test/integration/runtime/chat-start-event.test.ts` asserts the new-conversation `data.changed` reaches the default (SSE-bound) sink and that resume does not broadcast (12 pass, 0 fail in isolation).

Closes #155

### DIFF
--- a/src/api/handlers.ts
+++ b/src/api/handlers.ts
@@ -319,8 +319,7 @@ export async function handleChatStream(
           event.type === "tool.preparing.done" ||
           event.type === "tool.start" ||
           event.type === "tool.done" ||
-          event.type === "llm.done" ||
-          event.type === "data.changed"
+          event.type === "llm.done"
         ) {
           if (event.type === "chat.start") {
             broadcastUserMessageOnce();
--- a/src/conversation/auto-title.ts
+++ b/src/conversation/auto-title.ts
@@ -1,5 +1,8 @@
 import type { LanguageModelV3 } from "@ai-sdk/provider";
 
+/** Hard cap so a hung fast-model call can't leave the caller's promise pending. */
+const TITLE_TIMEOUT_MS = 10_000;
+
 /**
  * Generate a short conversation title using the provided model.
  * Non-blocking — call fire-and-forget after first turn.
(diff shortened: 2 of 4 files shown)


mgoldsborough:
OBE by #302

PR #388, opened 2026-06-08
fix(identity): configurable WorkOS admin-role slug mapping; preserve app owner across login

mgoldsborough:
## Problem

`resolveOrgRole` (the WorkOS provider's external→app role mapping) matched only the **exact** slug `"admin"` → app `admin`, and collapsed everything else to `member`:

```ts
const roleSlug = (membership.role as { slug?: string })?.slug;
if (roleSlug === "admin") return "admin";
return "member";
```

Consequences:

- A WorkOS org whose admin role carries a **custom slug** (e.g. `org-admin`), or differs in case, silently maps every such user to `member` — so the tenant has **no working org admins**, and there is **no log line** explaining the downgrade.
- `owner` is **unreachable** from WorkOS — yet the rest of the platform treats it as first-class (`ORG_ADMIN_ROLES`, the last-owner guards, the `manage_users` enum).
- `syncLocalProfile` overwrites `orgRole` on **every** login, so a `manage_users`-set `owner` is durable on OIDC/dev but **silently reverted** on WorkOS — and the login-sync path bypasses the last-owner guard entirely, so a WorkOS tenant can decay to zero owners.

## Fix

- **`resolveOrgRole`** — match a **configurable, case-insensitive** set of admin slugs (default `["admin", "owner"]`). Any unrecognized non-`member` slug that falls through is **logged** (with the actual slug + the configured set), turning a silent privilege downgrade into a diagnosable one. The normal `member` slug stays quiet.
- **`WorkosAuth.adminRoleSlugs`** — new optional `instance.json` config so operators with custom WorkOS admin slugs map them without a code change. An explicit list replaces the defaults; the unmatched-slug warning makes a forgotten `admin` visible in logs.
- **`syncLocalProfile`** — never **downgrade** a local `owner` on login. `owner` becomes a coherent **app-internal elevation**: created/removed only via the guarded `manage_users` path (which enforces the last-owner invariant), never stripped by a WorkOS membership change. `admin`/`member` continue to track WorkOS.

### Design note

This deliberately keeps `owner` out of the WorkOS-derived set: WorkOS grants `admin`/`member`; `owner` is the app's break-glass tier layered above. A WorkOS owner-slug role therefore grants app `admin` (sane default), and app `owner` is managed in-app. This makes the last-owner guard actually hold on WorkOS tenants.

## Test plan

New `test/unit/identity/workos-org-role.test.ts`:
- default `admin`/`owner` slugs → `admin`; case-insensitive match
- unrecognized slug → `member` **and** logs the downgrade
- custom `adminRoleSlugs` honored; explicit list replaces defaults
- no `organizationId` → `member`; no membership → `null` (deny)
- `syncLocalProfile` preserves a local `owner` when WorkOS reports a lesser role; still syncs `admin`/`member` for non-owners

Existing `workos-provisioning.test.ts` unchanged and green. `bun run verify:static` + `bun run test:unit` pass.

### DIFF
--- a/src/identity/instance.ts
+++ b/src/identity/instance.ts
@@ -30,6 +30,16 @@ export interface WorkosAuth {
    * exposes /.well-known/oauth-protected-resource for MCP client discovery.
    */
   authkitDomain?: string;
+  /**
+   * WorkOS organization-membership role slugs that map to the app `admin`
+   * role, matched case-insensitively. Defaults to `["admin", "owner"]` when
+   * omitted. Set this when your WorkOS org's admin role carries a custom slug
+   * (e.g. `org-admin`) — otherwise it silently maps to `member`. Any slug not
+   * listed maps to `member` (and is logged). Note: `owner` is an app-internal
+   * elevation assigned via `manage_users`, not a WorkOS-derived role; listing
+   * an `owner` slug here grants app `admin`, not app `owner`.
+   */
+  adminRoleSlugs?: string[];
 }
 
 export type AuthConfig = OidcAuth | WorkosAuth;
@@ -106,6 +116,16 @@ function validateAuthConfig(raw: unknown): AuthConfig {
         if (typeof auth.authkitDomain !== "string")
           throw new Error("instance.json: workos auth 'authkitDomain' must be a string");
         workos.authkitDomain = auth.authkitDomain as string;
+      }
+      if (auth.adminRoleSlugs !== undefined) {
+        if (
+          !Array.isArray(auth.adminRoleSlugs) ||
+          auth.adminRoleSlugs.some((s) => typeof s !== "string")
+        )
+          throw new Error(
+            "instance.json: workos auth 'adminRoleSlugs' must be an array of strings",
+          );
+        workos.adminRoleSlugs = auth.adminRoleSlugs as string[];
       }
       return workos;
     }
--- a/src/identity/instance.ts
+++ b/src/identity/instance.ts
@@ -32,12 +32,14 @@ export interface WorkosAuth {
   authkitDomain?: string;
   /**
    * WorkOS organization-membership role slugs that map to the app `admin`
-   * role, matched case-insensitively. Defaults to `["admin", "owner"]` when
-   * omitted. Set this when your WorkOS org's admin role carries a custom slug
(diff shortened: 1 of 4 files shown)


mgoldsborough:
Thanks for the review — both points were right; addressed in d641c40.

**Critical (store/session divergence) — confirmed and fixed.** Verified against source: `resolveUser` is the single session-identity construction site (both `verifyRequest` JWT paths funnel through it; `exchangeCode`/`listUsers`/`provisionUser` don't build a session identity), and it set `identity.orgRole` from the raw `resolveOrgRole` value while `syncLocalProfile` discarded the effective role. So a preserved owner was durable in the store but inert in the live session that authz reads — exactly the break-glass case granting nothing.

Fix: `syncLocalProfile` now returns `{ preferences, orgRole: effectiveRole }`; `resolveUser` builds `identity.orgRole` from it. The race-on-create branch preserves a raced owner the same way. New test asserts `resolveUser` yields `orgRole: "owner"` for a stored owner whose WorkOS slug maps to `member`.

**Suggestion (empty `adminRoleSlugs`) — fixed by rejecting, not documenting.** An empty/blank-only list now throws at config load (`instance.ts`), consistent with the "no silent surprises" intent of the change — silently falling back to defaults (or silently locking everyone out) were both worse. Tests cover the rejection and a valid custom-slug round-trip.

`verify:static` green; `test:unit` 3312 pass / 0 fail.

mgoldsborough:
Both optional-polish items addressed in 43f31ec.

1. **Blank-only `adminRoleSlugs` → empty set.** Chose "fall back to defaults" over "tighten the comment": `normalizeAdminRoleSlugs` now returns the defaults when normalization yields an empty set, so it can never produce a zero-admin set (the lockout footgun). Config-level empties stay rejected in `instance.ts`; this is the defense-in-depth invariant for direct callers. Test: blank-only `["  "]` resolves `admin` via the defaults.

2. **Per-login warn volume.** The unmatched-slug warning now fires once per slug per process (`warnedUnmatchedSlugs` set), so a tenant with legitimate non-admin slugs (`viewer`, `guest`) gets the diagnostic on first sight without recurring per-login noise. Test: a repeated non-admin slug warns exactly once.

`verify:static` green; `test:unit` 3314 pass / 0 fail.

PR #395, opened 2026-06-08
fix(authz): unify workspace-scoped write gates behind one strict helper (#389)

### LINKED ISSUE #389: Org-admin override into non-member workspaces is enforced inconsistently across workspace-scoped writes
## Summary

Workspace-scoped write gates disagree on whether an **org admin/owner who is not a member of the target workspace** may act. The same conceptual operation — "an org admin writes workspace-scoped content in a workspace they haven't joined" — is implemented four times with two different policies.

## The four gates

| Workspace-scoped write | Org-admin override (non-member)? | Site |
|---|---|---|
| Instructions | **Yes** (org-admin bypass first) | `src/tools/platform/instructions.ts` (~L71) |
| Workspace identity | **Yes** | `src/tools/core-source.ts` (~L643) |
| Member management | **Yes** | `src/tools/workspace-mgmt-tools.ts` (~L366) |
| Skills | **No** — must be a workspace member with `admin` role | `src/tools/platform/skills.ts` (~L1167) |

Skills' strict stance is deliberate (it carries a comment to the effect of "no org-admin override into a workspace the operator isn't a member of"). The other three grant a bypass. So today an org admin can rewrite a workspace's instructions and identity, and manage its members, without joining — but cannot touch its skills without being added as a member.

## Why it matters

This is a security-relevant inconsistency: the authorization boundary for "org admin reaching into a workspace they're not a member of" depends on which resource is being written, which is surprising and hard to reason about. It is almost certainly not intentional that skills diverges from the other three.

## Proposed fix

1. Decide one policy: either org-admin/owner may override into any workspace's scoped writes, or all workspace-scoped writes require explicit membership (the stricter skills behavior).
2. Unify the four gates behind a single shared helper, e.g. `canWriteWorkspaceScoped(identity, wsId)`, so the policy lives in one place and cannot drift again.

_Surfaced during the role audit behind #388, which fixed a separate WorkOS role-mapping seam._

mgoldsborough:
Closes #389.

## Problem

The same conceptual operation — "an operator writes workspace-scoped content in a workspace" — was authorized in several independent places with two different policies. Most gates let an org admin/owner bypass workspace membership; skills deliberately did not. So an org admin could rewrite a workspace's instructions/identity and manage its members without joining, but could not touch its skills. The authorization boundary depended on which resource was written — surprising and hard to reason about.

## Decision

Adopt the **strict** policy everywhere: a workspace-scoped **write** requires the operator to be a workspace member with `role === "admin"`. `orgRole` (org admin/owner) grants **no** bypass. This matches the deliberate skills behavior and the HTTP middleware (`resolveWorkspace`) that already requires membership to resolve a workspace.

### 1. One strict helper, five gates
- **New helper** `canWriteWorkspaceScoped(identity, ws)` in `src/workspace/authz.ts` — pure, fail-closed (malformed/`undefined` members → deny), no I/O, no `src/tools/*` imports. The single source of truth for the policy.
- **Five gates** now delegate to it (the org-admin bypass is removed from each):
  - Instructions — `instructions.ts` `checkScopePermission` (workspace scope)
  - Workspace identity — `core-source.ts` `manage_identity`
  - Member management — `workspace-mgmt-tools.ts` `canManageMembers`
  - Skills — `skills.ts` `checkPathAccess` (workspace **write** branch; **read** semantics unchanged — any member may read)
  - Connector installs — `connector-tools.ts` `isWorkspaceAdmin` (a **fifth** gate surfaced by a drift grep during the audit; it still carried the bypass)

### 2. Bootstrap fix
`handleCreate` now seats the workspace creator as an `admin` member. Without this, removing the org-admin bypass from member management would strand a freshly created shared workspace with no one able to manage it (it was created with `members: []`). Personal-workspace seating was already correct and is untouched.

### 3. `claim_admin` recovery action (added after QA)
QA flagged that the bootstrap fix only helps *new* workspaces — pre-existing shared workspaces with members-but-no-admin (or content-but-no-members, created under the old bypass) would become permanently unmanageable, with delete+recreate as the only, lossy recovery. There is no migration framework to hook a one-off script into.

Added a `claim_admin` action on `manage_workspaces` (org-admin gated, like the other CRUD actions): an org admin seats **themselves** as an admin member of a shared workspace **that has no admin member**. It refuses when an admin already exists, so it is a narrow, auditable recovery lever — not a reinstated per-write bypass into healthy workspaces. Personal workspaces are rejected (always owner-admin). Covers both the membered-adminless and memberless-with-content cases. `isAdmin` was made a type guard so the narrowed identity flows to the handler.

## Deliberately left org-admin gated (not workspace-scoped writes)

Org-scope instruction/skill writes, `set_model_config` (org-wide), workspace CRUD (create/delete/list), user/registry/usage org operations, and skills reads.

## Tests & verification

- New `test/unit/workspace/authz.test.ts` covers the strict policy, the org-admin-without-membership denial (the core regression guard), and the fail-closed case.
- Each gate's tests updated to assert the new strict behavior; stale tests asserting the old bypass were inverted/removed. Bootstrap deadlock has an explicit test (creator manages members with org role dropped to `member`).
- `claim_admin` tests: succeeds on membered-adminless and memberless shared workspaces; recovery proven end-to-end; refused when an admin already exists and when targeting a personal workspace; non-org-admin denied.
- `bun run check` ✓ · `bun run lint` ✓ · `bun run test:unit` ✓ (3336 pass, 0 fail)


mgoldsborough:
## QA adjudication — Critical #1 (adminless shared-workspace lockout)

**Verdict: valid and confirmed.** Verified against source: pre-PR shared-workspace `create` seats zero members, `add_member` defaults to `member`, and post-PR every member-seating caller is strict-gated — so an existing shared workspace with members-but-no-admin (or content-but-no-members) had no in-product recovery once the org-admin bypass was removed. Two refinements to the QA writeup: there is **no migration framework** to hook a one-off script into, and a load-time self-heal would only cover the *membered*-adminless subset (not memberless workspaces that hold instructions/identity/skills).

**Fix shipped (commit `0ff532d`):** a `claim_admin` action on `manage_workspaces` (org-admin gated, like the other CRUD actions). An org admin seats *themselves* as an admin member of a shared workspace **that has no admin member**. It refuses if an admin already exists, so it is a narrow, auditable recovery lever — not a reinstated per-write bypass into healthy workspaces. Covers both the membered-adminless and memberless-with-content cases; personal workspaces are rejected (always owner-admin). This keeps the strict per-resource write policy intact while giving org admins a deliberate way to un-strand a workspace.

Tests: claim succeeds on membered-adminless and memberless shared workspaces; recovery proven end-to-end (operator manages members after claim with org role dropped to `member`); refused when an admin already exists and when targeting a personal workspace; non-org-admin denied.

**Suggestion #1** (single-write create via `opts.members`) — not taken: passing `members` for a non-personal workspace is fine, but the two-step create+addMember mirrors the personal-provisioning path and keeps the seating logic in one obvious place. Noted as optional; can fold in if preferred.

`bun run check` ✓ · `bun run lint` ✓ · `bun run test:unit` ✓ (3336 pass, 0 fail)

PR #425, opened 2026-06-11
fix(auth): derive the public origin in one seam to fix custom-domain OAuth

mgoldsborough:
## Problem

Connector OAuth callbacks (Composio, static-auth MCP) and the post-callback redirect were built from `NB_API_URL` — the platform subdomain — while users on a custom-domain tenant sign in on the branded domain. The OAuth return leg therefore landed cross-origin, with no session cookie, and 401'd (connectors stuck "not connected").

Root cause: "what is this tenant's public origin?" was smeared across `NB_API_URL`, `NB_WEB_URL`, `WORKOS_REDIRECT_URI`, and `ALLOWED_ORIGINS[0]` — values in multiple layers that had to agree by hand and didn't.

## Change

One kernel seam, `src/oauth/public-origin.ts`:
- `publicOrigin()` — vendor/OAuth callback URLs
- `webOrigin()` — user-facing SPA returns (dev API/SPA port split)
- `canonicalOrigins()` — CORS auto-include

It is **config-derived, never request-Host-derived** (host-header injection / open-redirect vector), and **fail-closed** (asserts a bare `https` origin, localhost-exempt). It is the only sanctioned reader of `NB_API_URL` / `NB_WEB_URL`, enforced by `scripts/check-public-origin.ts` (wired into `verify:static`).

Every consumer now resolves through it:

| Consumer | Origin |
|---|---|
| Composio callback | `publicOrigin()` |
| MCP static-auth callback | bouncer-first (untouched), else `publicOrigin()` |
| connectors-redirect | `webOrigin()` |
| WorkOS login redirect | derived; `instance.json` `redirectUri` now optional, explicit value still overrides |
| `appOrigin` + CORS | `publicOrigin()` / `canonicalOrigins()` — decoupled from `ALLOWED_ORIGINS[0]` |

The Helm chart forwards facts (`NB_PLATFORM_HOST`, `NB_CUSTOM_DOMAIN`, `NB_CUSTOM_DOMAIN_CANONICAL`, optional `NB_PUBLIC_ORIGIN`); origin policy lives in the runtime.

## Carve-outs / trust

- Bouncer-mediated static-auth connectors keep their fixed vendor-registered `redirect_uri` — not switched to the custom domain.
- Vendor registration (WorkOS dashboard) is still a manual pre-step; deriving the value doesn't register it.
- Existing **non-bouncer** static-auth connectors reconnect on cutover (their cached `redirect_uri` changes — the DCR drift check handles it). Composio unaffected (per-request callbackUrl).

## Verification

- Unit: 3437 / 0 · Integration: 590 / 0
- Typecheck (server + web), lint, all guards incl. `check:public-origin`: clean

## Rollout

Per-tenant via `customDomainCanonical` (default true; an unverified tenant can be pinned `false`). Roll the dogfood tenant first, verify Composio end-to-end on its custom domain, then enable the rest. Deploy/tenant specifics live in the private deployments repo.

Reviewed against the runtime-kernel thesis (chief-architect): `publicOrigin()` is the floor everything resolves through, guarded so it stays the floor.


mgoldsborough:
Thanks — QA was correct on all three, addressed in the latest push (force-pushed to scrub the leak from branch history):

- **Critical (real tenant names in a public repo):** fixed. `public-origin.test.ts` now uses `acme` / `tenant-b` / `tenant-c` placeholders with `brain.acme.com`-style domains (assertions were name-agnostic — pure rename, 16/16 still green). Scrubbed the customer names from this PR description too.
- **Suggestion 1 (cached/uncached docstring contradiction):** fixed the stale module-header line — the uncached behavior is intended.
- **Suggestion 2 (`dev-worktree.ts` → `src/cli/dev.ts`):** corrected the reference.

Separately flagging a *pre-existing* leak not in this PR's scope: `test/integration/workspace-security.test.ts:270` carries a real tenant name in a comment on `main` — worth a standalone cleanup.

mgoldsborough:
Addressed in 1a09630.

**Critical (untested WorkOS derivation): fixed — this was a real gap.** `publicOrigin()` was covered but the WorkOS *wiring* wasn't, and a regression there would break login for exactly the custom-domain tenants this targets while the suite stayed green. Added both layers:
- `instance.test.ts` — a `workos` config with `redirectUri` omitted loads with the field absent (not defaulted, not thrown); a non-string `redirectUri` is rejected.
- `workos-redirect-uri.test.ts` — the provider derives `${publicOrigin()}/v1/auth/callback` when omitted (custom-domain + platform-host cases), and an explicit value still overrides. Observed through the public `getAuthorizationUrl()`'s `redirect_uri` param (no network).

**Suggestion 1 (canonicalOrigins doc):** reworded — now states the custom domain is included even when pinned non-canonical (the ALB still serves it, so CORS must allow it).

**Suggestion 2 (appOrigin):** went ahead and switched it to `webOrigin()`. You're right it's moot in prod (identical), but `appOrigin` is a browser-landing value like the connectors return, so aligning it gives one clean rule — browser-facing → `webOrigin()`, vendor-facing → `publicOrigin()` — and future-proofs a dev-WorkOS path. Zero behavioral change (no tenant sets `NB_WEB_URL`).

Full suite green: 3442 unit / 0, lint + typecheck + `check:public-origin` clean.

mgoldsborough:
Both polish items applied in ca2887b.

**Suggestion 1 (canonical pin reliability):** good catch — this protects the rollout safety valve. `NB_CUSTOM_DOMAIN_CANONICAL` now parses case-insensitively (`false`/`False`/`FALSE` all pin) and **fails closed on anything unrecognized** (`no`, `0`, `off` → throws at boot) rather than silently defaulting to canonical. An unverified tenant pinned `false` can no longer be flipped to its custom domain by a typo — consistent with the module's other fail-closed asserts. Tests cover case-insensitivity, the unset default, and the typo→throw path.

**Suggestion 2 (IPv6 loopback):** added `http://[::1]` to the localhost-dev exemption + a test.

Full suite green: 3448 unit / 0, lint + typecheck + `check:public-origin` clean. Thanks for the thorough pass — the failure-mode checklist (asymmetric defense, host-injection, fail-closed availability) is exactly the right lens for this change.

mgoldsborough:
Fixed in 5d1f9d4 — worth doing, because the drift had already happened.

You're right that they can't merge wholesale (bare origin vs full callback URL with a path), but the *scheme/loopback* rule is the shared part, and it had already diverged: my harden commit added `[::1]` to `assertOrigin` and not to `resolveWorkosRedirectUri`. Extracted `isAllowedOriginScheme(url)` into `public-origin.ts` (the owner of origin validation); both call it now. The path-handling difference stays local to each (assertOrigin rejects real paths; the WorkOS one allows `/v1/auth/callback`). Side effect: the existing drift is fixed — WorkOS now accepts an http loopback override including `[::1]` — with a regression test asserting it.

This was thematically the right call to flag on *this* PR specifically — leaving a duplicated validation predicate undercuts the "one source of truth" thesis the whole change is built on. Full suite green: 3449 / 0.

PR #437, opened 2026-06-12
feat(deep-research): durable deep_research + OutputStore provider seam

mgoldsborough:
Adds the runtime-driven **deep_research** capability and the **OutputStore provider seam** behind it. Proven end-to-end live on staging (tenant-hq).

## What this does
A new `nb__deep_research` system tool: the runtime mints a tenant-key token, POSTs a durable task to **nimbletasks**, polls the baked **web** worker (Exa search → fetch → cited report), persists the report through a kernel **OutputStore**, and returns a `resource_link` the UI renders. The runtime is the only tenant-bound principal — bundles/workers never hold a data-plane credential.

## Key pieces
- **`nb__deep_research`** trigger — create task → poll → persist → return a `files://` resource_link (+ one-line summary). Fail-closed: an error returns no body and never invites the model to fabricate a substitute.
- **`OutputStore`** (`src/files/output-store.ts`) — one kernel interface, pluggable backends: `dataplane` (artifacts service, RLS), `local` (workspace-scoped PVC), `null` (disabled). Explicit provider selection, default `local` (zero-infra / off-platform). One resource primitive; uploads vs outputs differ by `kind`/`scope`/`producedBy` metadata, not a parallel store/scheme.
- **`nb__get_output`** + an `outputs` resolvable source so `read_resource(files://…)` resolves stored outputs (dataplane + local).
- **Result shape fix** — the report rides as a `resource_link` (jsonl stores the *ref*, not ~17KB), the model is told not to reproduce it (the UI renders it), and `resourceLinks` is now **persisted** so the artifact survives conversation reload.
- **Scope fix** — the local backend writes to a workspace-scoped area and is workspace-fenced (was leaking into the identity-owned user file store).

## Verification
Full unit suite **3571 pass, 0 fail**; `tsc` + biome clean. Proven live on staging: mint → task → worker → artifact → resource-link render, no fabrication / duplication / per-turn bloat, workspace-fenced. Design + decisions in `research/SPEC-output-store-provider.md`; task tree in `.tasks/output-store-provider/`.

## Follow-ups (tracked)
- Topology: agent-chart `dataPlane.enabled` flag + tenant-template egress (retire per-tenant hand-wiring); CA distribution decision (trust-manager vs gateway).
- Platform: writable `source`/`producedBy` column on the artifacts service.
- Stacked: `feat/artifact-panel` (the document-panel UI for these resource_links).


mgoldsborough:
Closing without merging. The `deep_research` capability is moving **out of the runtime kernel** and onto the `web` MCP server as a task-augmented MCP tool. The runtime drives the async result over the generic MCP task rail it already has (`src/tools/mcp-source.ts`) and renders the returned `resource_link` generically, resolving it via `resources/read` — zero domain code in the kernel.

The kernel `OutputStore` seam (`src/files/output-store.ts`), `nb__get_output`, and the dataplane client introduced here are intentionally **dropped**: `web` writes its report to the artifacts service and the runtime reads it back through the existing generic resource path, so there is no surviving producer or consumer for an output store in the kernel.

The generic primitives this PR also touched (tenant-key mint, remote transport, the MCP task driver/store, the resource-link UI) already landed separately on `main` and stay. A standing grep-gate test now guards `src/` against any net-new domain hits so this stack cannot reland in the kernel.

Design and migration tasks: `research/SPEC-deep-research-mcp-migration.md` and `.tasks/deep-research-migration/`.

PR #440, opened 2026-06-13
test: standing grep-gate keeps the runtime kernel domain-clean

mgoldsborough:
## What

Adds a standing unit test (`test/unit/runtime-domain-clean.test.ts`) that guards the runtime kernel against domain capability creeping back into `src/`.

The test:
- Runs `git grep --untracked -niE 'deep.?research' -- 'src/*'` and asserts the hit count stays at or below the current benign baseline (**8**: third-party model catalog names — `o3-deep-research`, `o4-mini-deep-research`). Any net-new domain hit pushes it over the ceiling and the test fails loudly, listing the offending lines. `--untracked` ensures a not-yet-committed file is caught too.
- Asserts `src/files/output-store.ts` and `src/tools/get-output.ts` do **not** exist, so the dropped output-store seam cannot reappear.

The word-gate deliberately does **not** match `artifact`: the kernel uses that word pervasively for unrelated infra (a live `artifacts` data-plane service, plus build/cache/tmp/oauth artifacts), so gating on it by count would fail on benign growth while blaming a "domain capability leak." The removed output-store seam — the actual artifact-producing capability — is guarded precisely by the file-existence test instead.

## Why

Capability comes from MCP servers; the kernel stays generic. This is a ratchet test — if you legitimately remove a benign hit, lower the ceiling in the same change. Wanting to raise it is the signal that a capability is being put in the wrong place.

## Verification

- `bun test test/unit/runtime-domain-clean.test.ts` — 2 pass.
- Verified the guard fails (9 > 8, offender listed) when a domain hit is injected, and passes again once removed.
- No production code changes.

### DIFF
--- a/test/unit/runtime-domain-clean.test.ts
+++ b/test/unit/runtime-domain-clean.test.ts
@@ -7,11 +7,17 @@ import { spawnSync } from "bun";
  * Standing guard: the runtime kernel stays generic.
  *
  * Capability lives in MCP servers, not in `src/`. A handful of generic
- * mentions of "deep research" (an English phrase, third-party model
- * catalog names) and "artifact" (the English word, plus service names)
- * legitimately occur in the kernel and form a fixed baseline. Anything
- * above that baseline means a domain capability has leaked back into the
- * runtime — this test fails loudly so it cannot ride in unnoticed.
+ * mentions of "deep research" (third-party model catalog names) legitimately
+ * occur in the kernel and form a fixed baseline. Anything above that baseline
+ * means a domain capability has leaked back into the runtime — this test fails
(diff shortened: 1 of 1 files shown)


mgoldsborough:
Superseded by #442, which landed `test/unit/runtime-domain-clean.test.ts` with a stronger matcher: it gates on domain *identifiers* (`deep_research`, `research.report`, `report_artifact`, `ResearchReport`) at a hard ceiling of 0, drops the noisy `artifact` word entirely (the generic `artifact://` resolver is kernel infrastructure), and adds an affirmative guard that the resolver is present and domain-free. main's file carries the explicit reconciliation note to keep its matcher over this branch's. Closing as obsolete.

PR #449, opened 2026-06-15
fix(bundles): boot-start tenant-key fleet sources (static auth)

mgoldsborough:
## Symptom
A workspace MCP source registered as a url bundle with `transport.auth.type: "tenant-key"` (a fleet data-plane service that mints a service token on demand) never auto-starts — its tools never surface, and the agent reports "no tools installed." Reproduced in staging: `[bundles] Skipping boot start for URL bundle "web" — no tokens yet (state: not_authenticated)`.

## Root cause
Two gates decide "is this url bundle authenticated?" purely on OAuth/Composio **token presence on disk**:
- boot-start gate — `src/runtime/workspace-runtime.ts` (`hasPersistedWorkspaceOAuthTokens`)
- `seedInstance` — `src/bundles/lifecycle.ts`

A **static-auth** source (`bearer`/`header`/`tenant-key`) carries its own credential and has *no* persisted OAuth tokens, so both gates treat it as unauthenticated: skipped at boot, and seeded `not_authenticated` — which `deriveConnectorStatus` renders as a **"Connect"** button that (via `startAuthInner`) would spin a **bogus OAuth flow against a server with no OAuth**. `startBundleSource` already short-circuits the OAuth provider for static auth (`hasStaticAuth`), so the source runs fine once started — only the two gates blocked it. The OAuth-centric check was never generalized when `tenant-key` was added to `RemoteTransportConfig`.

## Fix (right-layer, structural)
Extract the static-auth check into a shared `bundleHasStaticAuth(ref)` predicate (new `src/bundles/bundle-auth.ts`) and apply it at **both** gates: a static-auth url source now boot-starts (mints on demand) and seeds `running`. `deriveConnectorStatus` is left untouched — fixing the upstream **state** makes it derive `ready` correctly. `startup.ts`'s inline check now uses the same predicate (three copies → one).

A failed boot-start is filtered before `seedInstance` runs (`workspace-runtime.ts` drops undefined result slots), so recording `running` stays accurate — verified by an independent adversarial review.

### Composio guard (caught in QA)
Composio bundles also carry `header` auth, so `bundleHasStaticAuth` returns true for them too — correct for `startup.ts` (Composio never uses the OAuth provider), but at the two boot gates an **unconnected** Composio connector would then seed `running`/`ready` and lose its Connect button (every tool call fails "not connected"). Both gates now check the **Composio marker FIRST** and fall back to `bundleHasStaticAuth || persisted-OAuth-tokens` only for non-Composio url bundles. `STATIC_AUTH_TYPES` is an explicit allowlist (`bearer`/`header`/`tenant-key`), not `!== "none"`, so a future interactive auth type defaults to the OAuth path. Regression test: an unconnected Composio connector must seed `not_authenticated`, not `running`.

## SSRF guard: in-cluster HTTP for tenant-key fleet sources (commit 3)
With the fleet on internal HTTP, a tenant-key source (the deep_research `web` mcp-service) is reached at `http://mcp-web.mcp-shared.svc.cluster.local` — but `validateBundleUrl` (the SSRF guard) rejected all non-localhost http, even with `allowInsecureRemotes`, so the boot-start above would throw "must use HTTPS" and the fix wouldn't work end to end.

Add a scoped `fleetInternal` allowance: http is permitted for in-cluster (`.svc.cluster.local` / `.svc`) hostnames **only for tenant-key sources**. Trust argument:
- The fleet's trust boundary is **NetworkPolicy + the verified token** (ARCHITECTURE P4), not TLS — TLS here is defense-in-depth, the token is the control.
- Gated on the **source class** (`auth.type === "tenant-key"`), set at exactly one site (`startup.ts`), **not** the dev-only `allowInsecureRemotes` flag — a production posture for the fleet rail, not a dev escape hatch.
- A tenant **cannot self-assign** tenant-key auth: the install tool only emits remote-oauth / mpak-bundle / direct-url; tenant-key refs originate solely from operator-written config (`admin-add-source`). So no tenant-facing path can reach `fleetInternal`.
- The embedded-credential and private/reserved-IP rejections stay **above** the protocol check, so a raw private IP or credentialed URL can never be reached over http even for a fleet source. Safe because CoreDNS owns `cluster.local` (no external host resolves there; the validator does no DNS).

Tests: in-cluster http allowed for `fleetInternal`; external http + private IP still rejected; `.svc.cluster.local` NOT unlocked by `allowInsecure` alone.

## Tests
- `test/unit/tenant-key-source-boot.test.ts`: the predicate across all auth kinds; `seedInstance` records `running` (not `not_authenticated`) for a tenant-key ref; the unconnected-Composio regression. The real `seedInstance` state-derivation was previously **untested** (the existing helper shadowed it).
- `test/unit/url-validator.test.ts`: 5 new `fleetInternal` cases.
- Full unit suite green (3604 pass / 0 fail); root `tsc --noEmit` + biome clean.

## Verdict
Confirmed-with-correction (two sites, not one; fix is upstream state, not `deriveConnectorStatus`). Adversarial review: aligned. Unblocks tenant-key fleet sources (the deep_research `web` mcp-service) end to end.


mgoldsborough:
Good catch on the Composio regression — confirmed valid and fixed in 283e64c.

**Critical 1 (Composio regression):** Verified against source — Composio bundles carry `transport.auth.type: "header"` (`buildComposioWiring`, connector-tools.ts:1053-1058), so `bundleHasStaticAuth` is true for them, and the `bundleHasStaticAuth(ref) ||` prefix short-circuited `hasPersistedComposioConnection` → unconnected Composio connectors would seed `running`/`ready` and lose their Connect button. Real regression in a hot path. Applied your reorder exactly: **composio marker checked FIRST** at both gates (`workspace-runtime.ts`, `lifecycle.ts`), static-auth fallback only for non-Composio url bundles. `startup.ts` keeps the predicate directly (Composio correctly counts as static-auth there — it never uses the OAuth provider).

**Critical 2 (missing test):** Added — an unconnected Composio connector (header auth + composio marker, no persisted connection) now asserts `not_authenticated`, not `running`.

**Suggestion 1 (allowlist):** Adopted — `bundleHasStaticAuth` uses an explicit `[bearer, header, tenant-key]` allowlist so a future interactive auth type defaults to the OAuth path.

Verification: 3599 unit tests pass (incl. the new composio guard + the composio-install suite); tsc + biome clean.

PR #456, opened 2026-06-16
fix(conversations): orphaned run no longer swallows later turns in chat UI

mgoldsborough:
## Problem

A run killed mid-turn (process death, e.g. a pod restart/deploy bounce) emits no `run.done`/`run.error`. In the chat-UI JSONL projection (`src/bundles/conversations/src/jsonl-reader.ts`), `collectRun()` only breaks its event loop on a terminator matching the **same** runId. With no terminator, an orphaned run consumes the **rest of the event array** and the outer loop ends — so **every turn after the orphaned run is silently dropped from the rendered transcript on reload**, even though all turns are fully persisted on disk. Users perceive losing the tail of an actively-used conversation.

The sibling LLM-replay projection (`src/conversation/event-reconstructor.ts:227-229`) already guards exactly this case (the "implicit run end" break) — the two projections had drifted.

## Fix

- Port the implicit-run-end guard into `collectRun`: break **without consuming the event** on a foreign `user.message` / `run.start`, so the outer loop renders subsequent turns.
- Distinguish an **abandoned** run (a later turn follows — it can never complete) from a **trailing in-flight** run (ran off the end of the array). The former is stamped `stopReason: "interrupted"` and is **not** `pending`, avoiding a perpetual "still thinking" spinner on an old turn. The latter stays `pending` for the resume path to reconcile.

## Tests

Two regression tests in `test/unit/bundles/conversations/jsonl-reader.test.ts`:
- a mid-transcript orphaned run preserves all later turns and is marked `interrupted`;
- a genuinely trailing in-flight run stays `pending`.

Verified the first test **fails without the guard** and passes with it. `bun test` (25/25), `tsc`, lint, format all clean.

## Follow-ups (tracked, not in this PR)

- Extract a shared run-segmentation primitive so the two projections can't drift again.
- Emit a terminal event for abandoned runs at the source (graceful shutdown and/or a startup reaper) so orphaned runs never reach disk.

### DIFF
--- a/src/bundles/conversations/src/jsonl-reader.ts
+++ b/src/bundles/conversations/src/jsonl-reader.ts
@@ -486,9 +486,11 @@ function collectRun(
   let endTs = events[start]?.ts ?? "";
   let stopReason: string | undefined;
   // A run is "terminated" only when we see its run.done/run.error. If the loop
-  // runs out of events first, the turn was still in flight when the file was
-  // read → mark the message pending.
+  // runs off the END of the array first, the turn was still in flight when the
+  // file was read → mark the message pending. A run cut short by a LATER turn
+  // (foreign user.message / run.start) is `abandoned`, not pending — see below.
   let terminated = false;
+  let abandoned = false;
 
   let i = start + 1;
   while (i < events.length) {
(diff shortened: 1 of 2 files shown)


mgoldsborough:
QA's one suggestion (comment hard-references the sibling `event-reconstructor.ts` by path) is acknowledged and **deferred** — the durable fix is the already-tracked shared run-segmentation primitive (so the two projections can't drift again). Rewording the comment now would relocate the rot risk, not remove it. No changes; ready as-is.

PR #457, opened 2026-06-16
fix(web): recover from stale lazy-chunk 404s after deploy (white screen)

mgoldsborough:
## Problem

A browser tab open across a deploy holds the **previous** build's module graph. Its hashed lazy chunks (e.g. the `streamdown`/Shiki code-block module — `highlighted-body-<hash>.js` — loaded on demand when a message contains a code block) are gone from the server after a new build ships. The next dynamic import 404s:

```
Uncaught TypeError: Failed to fetch dynamically imported module: /assets/highlighted-body-<hash>.js
```

This is an unhandled rejection, and `main.tsx` rendered `<App/>` with no top-level boundary and no `vite:preloadError` handler — so the page goes **fully white** instead of recovering or degrading gracefully. (The only `ErrorBoundary` wraps `<Routes>`, leaving the shell/bootstrap layer above it unprotected too.)

## Fix (`web/src/main.tsx`)

- Listen for Vite's `vite:preloadError` and **reload once** to pick up the new `index.html` + current chunk hashes. A 10s `sessionStorage` time guard prevents a reload loop when the asset is genuinely unreachable (offline, or a deploy still mid-flight).
- Wrap the app root in `ErrorBoundary` so any render-time throw that escapes the inner route-level boundary degrades to the fallback card rather than a blank page.

Verified: `tsc --noEmit` clean, Biome clean.

## Follow-ups (tracked, not in this PR)

- Explicit "backend unavailable / reconnecting" UI for API/SSE failures instead of bounce-to-Login; auto-retry bootstrap on transient failure.
- Retain prior builds' `/assets/` chunks across a deploy (grace window) so an in-flight tab can still fetch its old chunk.

### DIFF
--- a/web/src/__tests__/stale-chunk-recovery.test.ts
+++ b/web/src/__tests__/stale-chunk-recovery.test.ts
@@ -1,0 +1,56 @@
+// ---------------------------------------------------------------------------
+// shouldReloadForStaleChunk — the loop guard
+//
+// The only real branch logic in the stale-chunk recovery: a reload fires at
+// most once per debounce window. A regression here (reload loop) would hammer
+// the asset server and trap the user on a blank page, so it's the one piece
+// worth pinning. Pure over an injected clock + Storage — no DOM needed.
+// ---------------------------------------------------------------------------
+
+import { beforeEach, describe, expect, test } from "bun:test";
+import { shouldReloadForStaleChunk } from "../lib/stale-chunk-recovery";
+
+/** Minimal in-memory Storage stub (only the two methods under test). */
+function fakeStorage(): Storage {
+  const m = new Map<string, string>();
+  return {
+    getItem: (k) => m.get(k) ?? null,
+    setItem: (k, v) => {
+      m.set(k, v);
+    },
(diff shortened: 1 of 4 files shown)


mgoldsborough:
Addressed all three QA suggestions in 4287b6e:

1. **Loop-guard test** — extracted the reload guard into `shouldReloadForStaleChunk` + `registerStaleChunkRecovery` (`web/src/lib/stale-chunk-recovery.ts`) and pinned the debounce with unit tests (first failure reloads; a second inside 10s does not; reloads again after the window). This is the one piece with real branch logic, so it's now covered.
2. **Root fallback centering** — `ErrorBoundary` default fallback `h-full` → `min-h-screen`, so the root boundary centers (#root has no explicit height). Inner usages tolerate it.
3. **Telemetry** — emit `stale_chunk_reload` before reload; PostHog persists it across the reload, giving visibility into how often tabs hit a deploy boundary (sizes the deferred asset-retention work).

`bun test` (3/3), `tsc`, Biome all clean.

PR #461, opened 2026-06-17
Artifact rehydration fix + read_artifact/list_artifacts retrieval tools

mgoldsborough:
Two related changes on the host artifact read path (first-principles MVP).

## 1. Fix: artifact rehydration on conversation re-open
A deep_research report rendered in the sidebar on first run but vanished when the conversation was reopened (observed: hq prod `conv_08c471e37b464b0a`). Root cause: the engine emits `resourceUri` + `resourceLinks` on `tool.done`, but `event-sourced-store.mapEngineEvent` dropped them on persist — while the jsonl-reader + UI already *read* them. Fix persists them (small references, not bytes) and declares them on `ToolDoneEvent`. Regression test added. (Fixes conversations created *after* this lands; the dropped link can't be recovered for older ones, though their bytes are still in `artifacts`.)

## 2. Feature: read_artifact + list_artifacts host tools
Generic artifact retrieval so an agent can find + read artifacts a prior turn or past conversation produced:
- `list_artifacts(type?, cursor?)` — discovery over the existing `GET /v1/artifacts` (filter + keyset), workspace-scoped, RLS-fenced.
- `read_artifact(uri|id)` — reuses the host `artifact://` resolver.

**Placement:** the trust model forbids a bundle in the artifacts read path (the host reads as the viewing user), so these are **host system tools** (kernel), not a fleet bundle — zero new trust surface, reuses `ArtifactReadClient`. The kernel stays domain-agnostic (passes the `runtime-domain-clean` grep gate). Free-text search (vs type filter) is a tracked follow-on.

## Verification
`bun test test/unit/` → 3658 pass / 0 fail; pinned tsc + biome clean. New tests: `ArtifactReadClient.list` (parsing, type filter, RLS-fence, fail-closed) + the tool.done resource-link persistence regression.

### DIFF
--- a/src/conversation/event-sourced-store.ts
+++ b/src/conversation/event-sourced-store.ts
@@ -9,6 +9,7 @@
 import { appendFileSync, existsSync, mkdirSync } from "node:fs";
 import { readFile, rename, unlink, writeFile } from "node:fs/promises";
 import { join } from "node:path";
+import type { ResourceLinkInfo } from "../engine/content-helpers.ts";
 import type { EngineEvent, EventSink } from "../engine/types.ts";
 import { ConversationCorruptedError } from "../runtime/errors.ts";
 import { assertNoBinaryPayloads } from "./binary-guard.ts";
@@ -553,6 +554,15 @@ export class EventSourcedConversationStore implements ConversationStore, EventSi
         // model-context bound. Replay uses it verbatim so the replayed prompt
         // matches what the model saw live. See boundToolResultForModel.
         const modelOutput = typeof d.modelOutput === "string" ? d.modelOutput : undefined;
+        // UI-binding resource references. The engine emits these on the live
+        // tool.done; persist them so a reopened conversation rehydrates its
+        // artifact viewers (the panel a tool's `artifact://` resource link renders into).
+        // They are small references, not bytes — the body is fetched on view.
+        const resourceUri = typeof d.resourceUri === "string" ? d.resourceUri : undefined;
+        const resourceLinks =
+          Array.isArray(d.resourceLinks) && d.resourceLinks.length > 0
+            ? (d.resourceLinks as ResourceLinkInfo[])
+            : undefined;
         const e: ToolDoneEvent = {
           ts,
           type: "tool.done",
@@ -563,6 +573,8 @@ export class EventSourcedConversationStore implements ConversationStore, EventSi
           ms: (d.ms as number) ?? 0,
           ...(output !== undefined ? { output } : {}),
           ...(modelOutput !== undefined ? { modelOutput } : {}),
+          ...(resourceUri !== undefined ? { resourceUri } : {}),
+          ...(resourceLinks !== undefined ? { resourceLinks } : {}),
         };
         return e;
       }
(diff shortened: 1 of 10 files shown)


mgoldsborough:
QA adjudication — all findings addressed (`e699769`):

**Critical 1 (typecheck fails on pinned tsc 5.9.3) — fixed, and you were right.** I reproduced it: `result.contents` is `TextResourceContents | BlobResourceContents`, so `c.text`/`c.blob` need narrowing. My PR-body 'pinned tsc clean' was **false** — I verified with `bun run check` (bunx-latest tsc), which masked it; the worktree had no local `node_modules` at the time. Narrowed with `'text' in c` / `'blob' in c`; `node_modules/.bin/tsc --noEmit` is now clean (only the pre-existing tsconfig `baseUrl` deprecation). Lesson logged: verify with the pinned `node_modules/.bin/tsc`, never bunx.

**Critical 2 (no handler tests) — fixed.** Added `test/unit/tools/core-artifact-tools.test.ts` driving the real handlers via the `setArtifactResolver` seam: text read, bare-id→`artifact://` normalization, binary placeholder, `ArtifactNotFoundError` mapping, empty-uri; list happy-path + type-filter threading + empty case. The text-extraction test exercises the exact lines that failed #1.

**Delete 1 (unused `source` filter) — dropped.** Removed from `ArtifactListOptions` + the query threading (YAGNI; the tool only surfaces type/cursor/limit).

**Suggestion 1 ('newest first') — verified accurate, keeping it.** Not an assumption: the store does `order_by(artifacts.c.created_at.desc())` (`artifact_store.py:259`), and the keyset cursor is `created_at`. It's a real data-plane contract.

**Suggestion 2 (gate name) — correct, noted.** There's no script-level gate; the check is the `test/unit/runtime-domain-clean.test.ts` guard (a `git grep` over `src/*`, ceiling 0). Substance holds — no domain terms in kernel src (`research.report` appears only in tests).

Full suite: **3665 pass / 0 fail**; pinned tsc + biome clean.

mgoldsborough:
QA round 2 — adjudication (`8c2846a`; no Critical issues this round):

**Suggestion 1 (newest-first ordering) — verified accurate, now documented.** Not an unenforced promise: the artifacts store does `order_by(created_at.desc())` and keysets on `created_at` (`platform … artifact_store.py:259`). Added a doc comment on `list()` recording that contract. Kept the wording.

**Suggestion 3 (non-positive limit reaches the data plane) — fixed.** `list()` now only forwards a positive limit (drops 0/negative → data-plane default). Test added: `limit=5` forwarded, `limit=0`/`-3` dropped.

**Suggestion 2 (binary body materialized then discarded) — valid, deferred.** Real but bounded (8 MB cap) and off the primary path (reports are text/markdown; a binary artifact via read_artifact is rare). A clean short-circuit needs a metadata-only read leg on the client (the resolver fetches metadata+content in one call), which is more surface than the MVP warrants. Tracked as a follow-on optimization, not done here.

**Suggestion 4 (read_artifact returns only `content`, no `structuredContent`) — intended.** The artifact body *is* the payload the agent wants, so it's the human-readable content; list returns a collection, so it gets structuredContent. No change.

Verified across **both** suites this round (the gap that bit the last review): unit + integration 4267 pass / 0 fail; pinned tsc + biome clean.

**On the necessity/bundling question** — deferring to the author; see the PR thread.

PR #479, opened 2026-06-18
fix(auth): pin org on WorkOS refresh + instrument verify rejections

mgoldsborough:
## Problem

Sentry **NIMBLEBRAIN-PRODUCTION-1** — `involuntary logout: retry_401`. One user on tenant `hq` (Matthew Gerard) bounced to login roughly hourly, mid-conversation.

`retry_401` means: a request 401'd → silent token refresh **succeeded** → the retried request **still 401'd** → the client logs the user out. The signal's framing ("a token we just minted is already invalid") is correct — but the cause was not the transient case it sounds like.

## Investigation (production logs, tenant-hq)

- **All 23 auth failures over 17h came from one user's IP** — user-specific, not global.
- Pod up **17h / 0 restarts** → rules out JWKS cold-cache; caches were warm.
- **No `[workos]` lines at all** — no `resolveUser failed`, no `DENIED`, no JWKS/signature errors. The only log was the generic `[auth] authentication failed`.
- **Refresh succeeded** (no refresh-failure logs), then the retried request hit a **silent** `return null` in `verifyRequest`.
- Incident clusters (16:25–26, 17:27) line up exactly with the two Sentry events.

The only `verifyRequest` exit that is **silent + user-specific + on a freshly-refreshed token** is the `org_id` gate (`workos.ts`). Config confirms `organizationId` is set, so the gate is active. Not clock-skew (would hit every user); not transient (logs warm, single-replica).

## Root cause

`WorkosIdentityProvider.refreshToken` called `authenticateWithRefreshToken` **without `organizationId`**, so WorkOS minted the refreshed access token against the user's *default* org — which for a **multi-org** user differs from the org the session was established under ((`buildAuthorizationUrl` pins `organizationId` on the authorization request; `exchangeCode` then enforces membership — the mint itself does not pass it)). The drifted token fails `verifyRequest`'s `org_id` gate on the very next request → `retry_401` → involuntary logout, about once per access-token lifetime.

## Fix

1. **Pin `organizationId` on the refresh** so token mint and token verify agree (omitted when no org is configured — single-org/open deployments unaffected).
2. **Instrument the silent rejection paths.** `verifyRequest`'s `return null` exits now log a structured `reason` (`org_mismatch` / `token_expired` / `malformed_jwt` / `bad_alg` / `missing_exp` / `missing_sub` / `jwks_unavailable`). Routine self-healing reasons (`no_token`, `token_expired`) log at debug; anomalies at warn. Only token-derived ids are stamped (`sub`, org ids) — never the token, email, or name (per the file's trust rule). This both fixes the observability gap that made this take log-spelunking and **confirms** the diagnosis on the next occurrence.

## Tests

`test/unit/identity/workos-org-scope.test.ts`:
- refresh passes the configured `organizationId`; omits it when unconfigured.
- `verifyRequest` rejects an org-mismatched token (and a no-`org_id` token) and emits the `org_mismatch` reason with `token_org` / `expected_org`.

`bun run verify:static` ✓ · `bun run test:unit` ✓ (3663 pass) · existing identity suite ✓ (131 pass).

## Scope / follow-up

OIDC provider has no token refresh, so it's unaffected. A separate latent issue — the client retry leg has no transient tolerance (a transient server 401 also triggers logout) — is **not** this bug and is left for a follow-up.


mgoldsborough:
Root-cause precondition confirmed: the affected user (the only one hitting this on `hq`) is **intentionally** a member of multiple WorkOS organizations. That's exactly the condition that makes an un-pinned `authenticateWithRefreshToken` drift to the user's default org and fail the `org_id` gate — so this isn't hypothetical, it's the mechanism. The fix pins each deployment's refresh to its configured org, keeping multi-org users on the right org per tenant. (No behavior change for single-org users — `organizationId` is omitted when unconfigured.)

mgoldsborough:
Addressed QA review in d84f877 — all three points accepted (verified each against source first):

- **Critical (bad_signature):** confirmed the WorkOS-branch `if (!verified) return null` was genuinely silent — `verifySignature` only `warn`s when `candidates.length > 0`, so the 0-candidate case (forged/unknown `kid`, JWKS rotation) was fully unlogged. Routed through `reject("bad_signature", { sub })`, symmetric with the AuthKit branch. The "all silent paths now logged" claim is now accurate. Added a `bad_signature` regression test (valid org gate + key not in JWKS).
- **Unverified org claim:** renamed `token_org` → `claimed_org` and added a comment noting the gate runs pre-signature-verification, so the value is unverified and not authoritative.
- **Namespace docs:** documented the new `auth` NB_DEBUG namespace in both `src/cli/log.ts` and `AGENTS.md` (CLAUDE.md symlink), per the keep-in-sync convention.

`bun run verify:static` ✓ · `bun run test:unit` ✓ (3664 pass, +1) · tsc/biome clean.

mgoldsborough:
Addressed QA round 2 in 20ca6c4:

- **Comment accuracy (the one finding):** confirmed against source — `exchangeCode` calls `authenticateWithCode({ clientId, code, codeVerifier })` with no `organizationId`; it uses `this.organizationId` only for the `resolveOrgRole` membership gate. Corrected the `refreshToken` comment (and the PR body) to credit `buildAuthorizationUrl` for pinning org on the authorization request, with `exchangeCode` enforcing membership. Narrative-only; fix logic unchanged.
- **AuthKit asymmetry (suggestion #1) — pulled in rather than deferred:** the AuthKit branch's JWKS/signature failures now route through `reject("authkit_jwks_unavailable" | "authkit_bad_signature")` at `warn` with a `sub` field, matching the User Management branch. Previously they logged at `error` with no reason field — invisible to the reason-keyed view this PR creates, which would defeat the instrumentation during an AuthKit incident. Existing AuthKit tests still pass (behavior unchanged: still `return null`); the bad-sig test now also shows the reason line.
- **CHANGELOG verbosity (suggestion #2):** left as-is — it matches the surrounding (uniformly verbose) entries in that section; trimming one bullet against the grain would be inconsistent.

`verify:static` ✓ · `test:unit` ✓ (3664 pass) · authkit+org-scope suites ✓ (15 pass) · tsc/biome clean.

Issue #503, opened 2026-06-20
Remove host-meta `category` field across schema + 11 app manifests (currently unused)


mgoldsborough:
## Context

The `ai.nimblebrain/host` `_meta` `category` field (was enum: sales/marketing/operations/research/finance/hr/engineering/support/custom) is **declared by all 11 UI apps but consumed by nothing**:

- The shell groups by **slot** (`ShellLayout` → `sidebar.<group>`), not category.
- Briefing prioritization uses facet `type` / section category (`attention/recent/upcoming`), not the app category.
- The read path (`extractUiMeta`/`extractBundleMeta`) doesn't persist it onto `BundleUiMeta`/`BundleInstance`.

Its only live effect was **install-gate failures** on drift (`synapse-reef`=`data`, `synapse-astro-editor`=`content` — outside the enum; `db-query` was already force-migrated `data`→`operations`).

## Interim (done in host-schema v1)

`category` was **neutered**: enum dropped, kept as an unconstrained optional string marked "reserved — not consumed by the host," so it stops gate-failing and doesn't break the 11 apps that declare it. Dead surface, but non-breaking.

## This issue — full removal (coordinated)

1. Drop `category` from `src/bundles/schemas/host-manifest.schema.json` + `HostManifestMeta` (types.ts).
2. Strip `category` from all 11 manifests: 10 × `synapse-apps/*/manifest.json` + `platform/mcp-servers/people/manifest.json`.
3. Re-add later via a version bump **only if** grouping-by-category is actually implemented and wired.

PR #504, opened 2026-06-20
Host placement discovery for fleet connectors (ServerDetail._meta)

mgoldsborough:
Fleet connectors (remote MCP services like People) never registered their host UI: the runtime read placement from a bundle **manifest** (mpak path) but never from a connector's **ServerDetail._meta**. So People shipped a correct `ui://` app + placement and still never appeared in the shell. This closes the gap on the **conventional MCP extension interface mpak already uses** — `ServerDetail._meta["ai.nimblebrain/host"]`, read once at install. **No resource scanning; the mpak path is untouched.**

See `specs/host-placement-discovery.md` for the full design + why `ServerDetail._meta` over resource-`_meta` scanning.

## Changes
- **server-detail.ts** — `ai.nimblebrain/host` on the `ServerDetail._meta` type + `getNimbleBrainHostMeta(s)` accessor (mirrors `getNimbleBrainConnectorMeta`).
- **defaults.ts** — one `hostMetaToUiMeta(_meta → BundleUiMeta)` helper; both extract paths (mpak manifest + fleet) delegate to it.
- **projection.ts** — catalog entry carries `ui` from the server-authored host meta; the **`interactive` badge is now derived** (explicit flag OR host placements) so it can't drift to false while a UI ships (the exact People bug).
- **connector-tools.ts** — install resolves `trustedUi` from the **operator catalog by id**, never the untrusted client-supplied entry (mirrors the provider-auth precedent).
- **lifecycle.ts** — `sanitizePlacements`: server-declared chrome validated fail-closed (own `ui://<thisServer>` namespace only, bounded strings, per-placement drop, never throws); `registerPlacements` uses it.
- **schema** — `host-manifest.schema.json` is now the published `$id: …/v1/nimblebrain-host.schema.json`; `settings` object dropped (settings is a `slot:"settings"` placement); `category` neutered to a reserved string (full removal: #503).
- **schema-deploy.yml** — publishes the host schema to `schemas.nimblebrain.ai/v1`. ⚠️ **Needs the OIDC IAM role granted the new S3 key (infra) or it 403s.**
- **types.ts** — drop `settings` from `HostManifestMeta`.
- **tests** — 14: fleet `ServerDetail` → placements, interactive derivation, sanitize/anti-spoof.

## Verification
typecheck 0 errors · lint clean · host-placement 14/14 · projection/connector/placement suites 113/113 · full unit suite green (one unrelated `@sentry/react` web-dep miss in this worktree).

## Pairs with
- **deployments#77** — the static-config bridge (People `_meta` in `platform.yaml`). Merge this first, then that makes People's app appear.
- **infra** — IAM key grant for the schema publish (task tracked).
- **platform#106** — fleet discovery (retires the bridge hand-copy).


mgoldsborough:
## QA adjudication + fix (pushed 709d05f)

Traced all three findings against the call graph. One was a real gap (fixed), one is incorrect, one is valid (addressed). Details:

**#1 — boot path bypasses sanitize → VALID (fixed).** Correct that `runtime.ts` (the workspace-bundle boot loop) called `placementRegistry.register` directly with the raw `instance.ui.placements`. Since sanitize returns a filtered copy and doesn't mutate `instance.ui`, a spoof dropped at install would re-register verbatim on the next restart. **Fix:** extracted `sanitizePlacements` to `defaults.ts` (pure module) and the boot loop now sanitizes before registering — same fail-closed guard as install. (The framing "sanitizePlacements is dead code" wasn't accurate — it *does* run at install; the gap was boot-only.)

**#2 — "fresh install needs a restart to surface the app" → INCORRECT.** `handleInstallRemoteOAuth` (the provider path People takes) calls `lifecycle.notifyInstalled` immediately after `seedInstance` (`connector-tools.ts:1264`), and `notifyInstalled` calls `registerPlacements` (sanitized) **and** emits `bundle.installed` so the web shell refreshes without a reboot. So a fresh fleet install surfaces the app live, sanitized — no restart. The review stopped at `seedInstance` and missed the next line.

**#3 — no test on the real registration path → VALID (added).** Added registry-level tests exercising the contract both paths share (`sanitize → register → forWorkspace`): a mixed-authority spoof never reaches the registry; an all-malformed set registers nothing.

**Residual surfaced by the new tests (worth a follow-up, not a blocker):** `sanitizePlacements`' anti-spoof is "one authority per server, first wins." It blocks *mixing* a foreign authority (the documented threat), but a connector declaring a *sole* foreign authority (e.g. only `ui://home/dashboard`) is accepted — there's no binding of the `ui://` authority to the server's identity. Severity is bounded: the host fetches a placement's resource **from the registering server**, so this is a mislabeled-sidebar-entry phishing risk (renders the connector's own content under a "Home"-ish label), not a takeover of the real host surface. A clean fix (bind authority ↔ server slug, or a reserved-host-authority denylist) needs care to not break legit bundles whose `ui://` namespace differs from their server name. Recommend tracking separately rather than expanding this PR.

`installRemote` deletion: it's prod-dead (test-only) but pre-existing, not introduced here — agree it's a separate cleanup.

Verification: typecheck clean; 375/375 across the bundle/runtime/registry/connector/placement suites.

mgoldsborough:
## QA round 3 adjudication + fix (pushed 56f560b)

**Critical #1 (published contract contradicts its own tolerance lock) — VALID, fixed.** Confirmed against current code: the schema was strict (`host_version` enum, root `additionalProperties:false`, the `if/then`) and the mpak gate hard-fails on schema-invalidity (`manifest-gate.ts:62`), while the fleet path is ungated/tolerant — so the two paths the spec calls "unified" had opposite tolerance, and dropping `settings` under `additionalProperties:false` was a breaking narrowing. The spec's "tolerance lock" promised the opposite.

Reconciled by relaxing the schema to honor the lock (the property we'd already committed to for a published, third-party-copied contract), strict only where a typo is dangerous:
- root `additionalProperties: false → true` (unknown keys ignored; the `settings` removal is now non-breaking — a bundle still declaring it validates)
- `host_version`: enum → open `string` (a newer version validates, not rejected)
- dropped the `host_capabilities ⇒ host_version "1.1"` `if/then` (forcing an exact version conflicts with forward-compat; the rule is advisory/documented now)
- **kept** `HostCapabilityRequirement` `additionalProperties: false` — a `requierd:true` typo would silently downgrade a security gate, so it's still rejected at install

The gate validates against this schema, so mpak **and** fleet are now both tolerant of unknown version/keys while the security-sensitive sub-object stays strict. Spec's "Versioning & tolerance" section updated to state the schema enforces this.

**Suggestion #1 (interactive derived from raw placements) — fixed.** The badge now derives from `sanitizePlacements(ui.placements)`, so a sole spoofed/foreign-authority placement (dropped at registration) no longer lights the chip while rendering nothing. Also tidies the sole-foreign-authority residual from round 2.

**Suggestion #2 (stray blank line) — already fixed** in 5fda6a2 (the lint commit); your review predated it.

Tests: added schema/gate tolerance-lock tests (unknown `host_version` + unknown top-level keys validate; capability-requirement typo still rejected; version↔capability pairing now advisory). Updated the two pre-existing tests that asserted the old `if/then` rejection. 293/293 affected suites pass; tsc + lint clean.

PR #510, opened 2026-06-22
prompt: evict volatile layers from the cached system block (P1 PR-1)

mgoldsborough:
## What

The composed system prompt is sent as **one system block under a single 1-hour cache breakpoint**, but it bakes in per-turn-volatile layers — the **current date**, **app / focused-app state**, and the per-message **matched skill**. Any one of those changing rewrites the *entire* cached system prefix at the premium write rate, every turn.

This splits composition into a **stable system prefix** and a **volatile head**:

- **`compose.ts`** — adds `composeSystemSegments()` returning `{ stableSystem, volatileHead }`. It delegates to `composeSystemPromptTraced` (single source of truth for layer order), tags each layer `stable | volatile` by kind, and splits the date out of `## User` into its own `## Current Date` layer. The volatile head is wrapped in a single `<runtime-context>` block, with the same closing-tag escape discipline as the existing containment tags.
- **`runtime.ts`** (chat + task) — sends `stableSystem` as the system block and **prepends the volatile head to the latest user message** (a leading text part), falling back to folding it into the system string if there's no user message to carry it. Budget and telemetry still size on every segment.

**Effect:** the stable system block becomes byte-identical across turns within a workspace, so it stops being rewritten — and re-cached — every turn.

## Scope (PR-1)

- **No `model/cache-policy.ts` change.** The breakpoint layout is untouched (system stays one block, the rolling history anchor is intact). This slice only moves volatile content off the cached system block; the physical stable/frozen split and breakpoint re-budget are **PR-2**, deterministic skill ordering + per-segment attribution are **PR-3**.

## Safety / backward-compat

- **No escape or sanitize logic changed** — only *which block each formatted section lands in*. Every per-block containment escape (`<app-state>`, `<app-guide>`, `<app-instructions>`, `<layer3-skill>`, `<skill-instructions>`, scope overlays) moves verbatim.
- `composeSystemPrompt` / `composeSystemPromptTraced` still return the full prompt (both tiers), so sizing/telemetry callers are unaffected.

## Tests

- New `test/unit/compose-segments.test.ts`: per-layer tier routing, stable/volatile partition, non-lossy reconstruction, and `<runtime-context>` escaping of a forged closing tag.
- `prompt.test.ts` / `prompt-injection.test.ts`: all containment/escape assertions kept; only date-location assertions updated (date is now its own `## Current Date` section; with no identity fields there is no empty `## User`). Added a `segment` integrity check and `current_date` / `task_identity` to the known-kinds set.
- `bun run verify:static` ✓ · `bun run test:unit` ✓ (3738 pass).

### DIFF
--- a/src/prompt/compose.ts
+++ b/src/prompt/compose.ts
@@ -21,6 +21,13 @@ const SEPARATOR = "\n\n---\n\n";
 export interface TracedLayer {
   kind: TracedLayerKind;
   /**
+   * Volatility tier. `volatile` layers (current date, app/focused-app state,
+   * matched skill) change per turn and are evicted from the cached system
+   * prefix onto the latest user message; `stable` layers form the cached
+   * system prefix. See `composeSystemSegments`.
+   */
+  segment: "stable" | "volatile";
+  /**
    * Stable identifier. Filesystem path for file-backed layers; `nb:<slug>`
    * for runtime-derived layers; `instructions://<scope>` for overlays.
    */
@@ -50,6 +57,7 @@ export type TracedLayerKind =
   | "core_skill"
   | "user_context_skill"
   | "user_prefs"
+  | "current_date"
   | "workspace_context"
   | "org_overlay"
   | "workspace_overlay"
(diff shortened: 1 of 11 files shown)


mgoldsborough:
Thanks for the review. Dispositions:

- **#1 (DRY the 4× fold)** — done in the latest commit. Extracted `foldVolatileHead(stable, head)` as the single source of the fold separator (so the budget/telemetry sizing and the engine-side fallback can't drift), plus `resolveEngineSystem(messages, stable, head)` for the prepend-or-fold. Behavior-preserving; `verify:static` ✓, `test:unit` 3738 pass / 0 fail.
- **#2 (authority shift for matched-skill / app-state → eval)** — valid concern. There's no existing eval covering skill adherence, and `matched_skill` moves to the pull model in PR-3, so the right place for a skill-adherence eval is PR-3 (alongside that change). Tracking it there rather than blocking this PR.
- **#3 (back-search edge)** — acknowledged; leaving as-is (normal path is the current turn; the edge mispositions but never loses content, and the no-user case folds back).

PR #512, opened 2026-06-22
prompt/cache: split system into frozen + per-workspace segments (P1 PR-2)

mgoldsborough:
## What

Splits the cached stable system prompt into two segments:
- **Frozen** — identity + core skills (stable per process/model).
- **Per-workspace** — scoped skills, overlays, apps (stable per workspace).

Under the Anthropic provider, on the **first iteration** of a run the cache policy emits them as **two consecutive cached system messages**, so switching workspace or installing a connector rewrites only the workspace segment — frozen identity stays warm. From **iteration 2** (once the rolling step-anchor exists) it **fuses** them back into one system message so the step-anchor keeps its breakpoint.

## Breakpoint budget (Anthropic max = 4)

- Split (iteration 1, no anchor): tools + frozen + workspace + tail = 4.
- Fused (iteration 2+): tools + system + anchor + tail = 4.

A 5th `cache_control` is silently dropped by the SDK, so a unit test asserts ≤4 in both modes, and that the two system messages are first + consecutive (avoids the `mid-conversation-system` beta).

## Unchanged

PR-1's volatile-head move (date / app-state / focused-app / matched-skill → `<runtime-context>` head on the latest user message) is untouched. `composeSystemSegments` now also returns `{frozen, workspaceStable}`; `stableSystem` stays the byte-identical fusion for sizing/back-compat. Non-Anthropic providers, transformed prompts, and plain callers keep the single-system-message path.

## Test surface

Under the Anthropic provider the prompt now has two system messages on iteration 1. Capturing-model test helpers that read a single system message would miss workspace-tier content, so they join all system messages via a shared `capturedSystemText` helper.

## Verification

All four gates green: `verify:static`, `test:unit` (3745), `test:integration` (601), `smoke` (23).

## Follow-up

PR-3: deterministic `layer3_skills` ordering + per-segment cache attribution.

### DIFF
--- a/src/engine/engine.ts
+++ b/src/engine/engine.ts
@@ -227,6 +227,12 @@ export class AgentEngine {
     systemPrompt: string,
     messages: LanguageModelV3Message[],
     tools: ToolSchema[],
+    // Optional frozen/workspace split of `systemPrompt` for the prompt cache.
+    // When provided (and the provider is Anthropic), the cache policy emits the
+    // two segments as separate cached system messages on the first iteration so
+    // the frozen identity stays warm across workspace switches. `systemPrompt`
+    // remains the source of truth for sizing and the fused (iteration 2+) path.
+    systemSegments?: { frozen: string; workspaceStable: string },
   ): Promise<EngineResult> {
     // Never mutate the caller's array
     const history = [...messages];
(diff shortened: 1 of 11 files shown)


mgoldsborough:
Closing. The implementation is correct and fully tested, but the two-mode design doesn't deliver a net cache benefit under Anthropic's 4-breakpoint budget:

- The frozen/workspace split only activates on the **first model call of a conversation** (no step-anchor yet). Every later call — the rest of the run, and all subsequent turns — takes the fused path (identical to the merged PR-1). Workspace switches / connector installs happen on those later turns, so the intended "frozen stays warm across switches" benefit never applies where it would matter.
- The split→fused transition changes the system representation on the second call of each conversation, forcing a partial system re-write — a small regression vs PR-1.

Root cause: the 4-breakpoint ceiling forces a choice between keeping the rolling step-anchor (this PR) and persisting the A/B split (would require dropping the anchor). PR-1 already captured the dominant win (per-turn system rewrite eliminated).

Revisit only if prod shows conversation-start / workspace-switch system writes are a meaningful share of cache writes — and if so, evaluate the always-split / drop-anchor variant with measurement, not assumption. The segment plumbing (composeSystemSegments) is preserved on main for reuse.

PR #514, opened 2026-06-22
skills: per-skill body cap (full body by default, cap only on prompt-load) [P2]

mgoldsborough:
## What

- **Per-skill body cap for the prompt.** `MAX_SKILL_BODY_CHARS` (shared constant) caps a skill body when it is rendered into the prompt — a backstop against **unvetted user/workspace Layer-3 skills** (injected uncapped before this change). Vendored skills are trusted and well under it.
- **Full body is the default; capping happens ONLY on the prompt-load path** (`parseSkillFileGuarded`). Every read/inspect/round-trip caller — `skills__read`, `writer.readSkill`/`listSkills` — returns the **full stored body**, so an agent that reads-then-rewrites a skill (`skills__read` → `skills__update`) can never silently truncate user-authored content. A future read door fails toward "uncapped prompt" (perf), not data loss.
- **The drift audit caps too — on purpose, not a read door.** `compose_effective_context`'s `bodyFromSkillFile` passes `cap: true` because the recorded `contentHash` is computed over the *capped* body (Layer-3 skills are hashed after the prompt-load cap, `skills-loaded-payload.ts`). It deliberately reproduces the **prompt-load body**, not the stored body — re-hashing the full body would false-report `drift` on every skill over the cap and snapshot recovery would never match.
- **Truncation signal.** `log.warn` when the prompt-load cap fires, deduped by `sourcePath` (once per process, not per chat turn).

## Reverted / dropped earlier in review
- **Kill-`always` default** — reverted (it silently dropped workspace/user-tier `type: context` skills, which reach the prompt only via Layer 3). The correct double-injection fix (de-dup the Layer-0/1-covered context skills from the L3 pool) lands in its own PR.
- **Aggregate Layer-3 budget** — deferred (inert ahead of a measured need).

## Tests
- `skills__read` of a skill larger than the cap returns the full body (the regression this fixes).
- Default `parseSkillContent` returns full; `cap:true` caps; the prompt-load path (`loadScopedSkills`) caps; the authoring round-trip preserves the full stored body.

## Verify
`verify:static` ✓ · `test:unit` 3744/0 · `test:integration` 613/0 · `smoke` 23/0.


mgoldsborough:
Thanks — the Critical was a real silent regression and I've reverted the cause. Dispositions:

- **Critical (workspace/user `type: context` skills silently stop loading)** — agreed; **reverted the kill-`always` change**. You're right that the dual-path safety only covers core/org (Layer 0/1); workspace/user-tier context skills reach the prompt only via Layer 3, so killing the default dropped them. The correct way to kill the double-injection — **de-dup the Layer-0/1-covered context skills from the Layer-3 pool**, plus a workspace-scope safety-net test — will land as its own focused PR (the double-injection is pre-existing token waste, not a regression, so it's safe to separate).
- **Delete Instead (aggregate Layer-3 budget)** — **deferred**. Same carrying-cost reasoning that pulled out the directory loader; will land with a measured trigger. The per-skill cap stays.
- **Suggestion #2 (docstring overclaim)** — resolved by the revert (the `loading.ts` wording is back to pre-change).

Net: this PR is now just the **per-skill body cap + truncation `log.warn`**. `verify:static` ✓ · `test:unit` 3742/0 · `test:integration` 613/0 · `smoke` 23/0.

mgoldsborough:
Fixed — and I went broader than the suggested door-patch on purpose; here's why.

- **Critical (`skills__read` silently truncates → data-loss on read-then-rewrite)** — fixed by **inverting the cap default**: `parseSkillContent`/`parseSkillFile` return the full body by default, and the *only* place the per-skill cap is applied is the prompt-load choke point (`parseSkillFileGuarded`, `{ cap: true }`). This is the third read-door from the same cap-at-parse root (writer was patched earlier, this one slipped through), so rather than patch door #3 I made full-body the safe default — now `skills__read`, `writer.readSkill`/`listSkills`, and the debug audit all return the full stored body, and any *future* read-door fails toward "uncapped prompt" (perf), never truncated content (data loss). No `skills.ts` change was needed; the inverted default fixes `skills__read` automatically. New tool-layer test reads a >12K skill via `skills__read` and asserts the end-marker survives.
- **Necessity note (threat model)** — comment restated: the cap defends the prompt against *unvetted user/workspace Layer-3 skills* (injected uncapped before this work), not vendored ones. That "untrusted" framing is also why the per-skill cap stays while the aggregate budget was deferred.
- **Suggestion #1 (per-turn log spam)** — deduped the truncation warn by `sourcePath` (once per process, not every turn).
- **Suggestion #2 (wording)** — no more "(0 section(s) omitted)" on a hard cut.

`verify:static` ✓ · `test:unit` 3744/0 · `test:integration` 613/0 · `smoke` 23/0.

PR #515, opened 2026-06-23
feat(connectors): API-key Composio connectors (PostHog, etc.)

mgoldsborough:
## What

Adds a non-redirect **API-key** auth path for Composio-backed connectors. Until now every Composio connector assumed the OAuth aggregator flow (redirect → vendor consent → callback). Toolkits like **PostHog** (and Exa, Firecrawl, SerpAPI, Perplexity, …) authenticate by API key — there is no redirect; Composio expects the key submitted at connect time. This unlocks that whole class, not just PostHog.

## How

- **`ComposioConnectorConfig`** (`server-detail.ts`) — one shared type for the composio block (was duplicated inline across `server-detail`, `registries/types`, and `projection`). Adds two fields:
  - `authScheme?: "OAUTH2" | "API_KEY"` — defaults to `OAUTH2`, so every existing connector entry is unchanged.
  - `fields?: ComposioConnectField[]` — what to collect from the user for non-redirect schemes.
- **`connectComposioApiKey()`** (`composio/sdk.ts`) — `AuthScheme.APIKey(fields)` → `connectedAccounts.initiate()` → `waitForConnection()` to verify the account reaches ACTIVE; deletes the half-created account on failure so a retry starts clean. `initiate` is the correct, **non-deprecated** call for non-OAuth schemes (the 2026-07-03 `link()` sunset explicitly excludes API key / bearer / basic).
- **`manage_connectors.connect_api_key`** (`connector-tools.ts`) — the API-key sibling of the OAuth `/v1/composio-auth/initiate` route. It's a **tool action, not a route**: there's no state cookie and no redirect/callback, so neither reason that path is a route applies. Server-trusted catalog lookup (`catalogById`), default-deny field validation (required present, unknown keys rejected), membership-gated, then reuses the OAuth connect tail (`ensureSourceRegistered` → `saveComposioConnection` → `recordConnectionStateChange("running")`).

## Trust posture

Identical to the OAuth path: the user's key is handed to Composio and **never persisted by the platform** — `connection.json` keeps only the opaque `connectedAccountId`. This deliberately does **not** reuse the stdio-bundle credential-store / env-inject convention; that would put the platform in needless custody of a key it never replays. The handler logs only `catalogId` / `wsId` / error text — never field values, and `api_key` is in the logger's auto-redact set.

## Testing

- `bun run verify` green (static: format/lint/tsc/cycles; 564 unit + web; smoke).
- New `test/unit/connector-tools-composio-apikey.test.ts` (6 tests): SDK forward-fields + verify-failure cleanup; action unknown-field / missing-required / non-API_KEY / unconfigured rejections.

## Out of scope / follow-ups

- **Web form** — the shell still needs to render the declared `fields` and POST `connect_api_key`. The backend is complete and exercisable via the tool action; the form is a separate PR.
- **PostHog catalog entry + auth-config** are deployment config (`deployments/nimblebrain/connectors/composio.yaml`, tracked separately), gated on **staging verification** of the exact field keys (`getConnectedAccountInitiationFields("posthog","API_KEY")` — may be `generic_api_key`) and the curated tool slugs.
- **OAuth-path `link()` migration** — the existing `initiateComposioConnection` is hit by the same 2026-07-03 deprecation; unrelated to this PR, worth a separate ticket.

### DIFF
--- a/src/composio/sdk.ts
+++ b/src/composio/sdk.ts
@@ -34,7 +34,7 @@
  *     short-circuits when `COMPOSIO_API_KEY` is unset.
  */
 
-import { Composio } from "@composio/core";
+import { AuthScheme, Composio } from "@composio/core";
 import { log } from "../cli/log.ts";
 import { getBouncerMode } from "../oauth/bouncer-config.ts";
 import { publicOrigin } from "../oauth/public-origin.ts";
@@ -51,6 +51,16 @@ export const COMPOSIO_CALLBACK_PATH = "/api/v3.1/toolkits/auth/callback";
 
 /** Max time a single Composio SDK call may run before we abort. */
 const COMPOSIO_TIMEOUT_MS = 10_000;
+
+/**
+ * Verification budget for an API-key connect. `waitForConnection` polls the
+ * freshly-created connected account until it reaches ACTIVE (or throws on a
+ * terminal FAILED/EXPIRED). Non-redirect schemes normally resolve on the
+ * first poll, so a few seconds is ample. Kept below `COMPOSIO_TIMEOUT_MS` (the
+ * outer per-call backstop) so the SDK's own typed timeout/failure error
+ * surfaces first instead of our generic abort.
+ */
+const COMPOSIO_APIKEY_VERIFY_MS = 8_000;
 
 // ── Config validation ────────────────────────────────────────────────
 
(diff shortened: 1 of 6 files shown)


mgoldsborough:
## QA adjudication — addressed in 511ba0a

**Critical 1 — handler tail untested (valid, fixed).** Ported the `stubCtx` pattern from `composio-auth.test.ts` and added 4 tests for the lifecycle tail:
- happy path — asserts `connection.json` is written (account + toolkit + status, and that the submitted key is *not* on disk) and `recordConnectionStateChange(serverName, wsId, "_workspace", "running")` fires once;
- `ensureSourceRegistered` failure → "install first" message, no persistence, no state change;
- auth-config env unset → operator-config error;
- Composio connect-failure → generic message, no persistence, dangling account cleaned up.

**Suggestion 1 — `initiatedId` fallback (valid, fixed).** Now reads `connRequest.id ?? connRequest.connectedAccountId`, mirroring the OAuth sibling, so the failure-cleanup delete can't be silently skipped if a future SDK shape moves the id.

**Suggestion 2 — early `isValidConnectorId` (declined, with reason).** `catalogById()` is the real gate: an unknown/malformed `catalogId` returns null → clean error *before* any path construction (`slugifyServerName`/`saveComposioConnection`), so the id is known-safe by the time it's used. The OAuth route needs the pre-check because its unauthenticated callback builds a filesystem path from `cid` before any catalog lookup; this tool path never does. Adding it would be redundant noise.

Suite: 3748 unit pass (0 fail), `verify:static` green.

mgoldsborough:
## QA round 2 — adjudication (71e8126)

**Critical 1 — re-connect / key-rotation orphans the prior account — valid, fixed.** Confirmed: `handleConnectApiKey` always minted a fresh account (`allowMultiple`) and overwrote `connection.json`, so the replaced account stayed ACTIVE at Composio (orphaned, still holding the old key) and `cleanupComposioBundle` could never reclaim it. Adopt-existing isn't viable here — a re-submit may carry a rotated key — so I implemented **delete-then-replace**: capture the prior account before connecting, and best-effort `deleteComposioConnectedAccount` it **after** the new account is verified + persisted (a failed rotation leaves the working account intact). New test: a second `connect_api_key` revokes the first account and keeps the new one (`ca_old` deleted, `ca_new` not). Suite now 3749 pass.

**Suggestion 1 — member vs admin authz — keep as-is (deliberate).** `connect_api_key` is member-gated to mirror the OAuth connect route; you flagged it's consistent and not a regression. Tightening connect to admin (like `disconnect`) is a defensible product call, but it must be made **across both connect paths (OAuth + API-key) in one change** — doing it only here would create the asymmetry you warned against. Leaving member-level; flagging for a separate deliberate decision if the team wants connect admin-gated.

**Suggestion 2 — `?? "ACTIVE"` status fallback — keep (with reasoning).** `waitForConnection` only resolves when the account is ACTIVE (it throws on FAILED/EXPIRED/timeout, per `createConnectionRequest`), so by contract `account.status` is "ACTIVE" on the resolve path; the `??` is belt-and-suspenders, not load-bearing. Left as-is — happy to revisit if the SDK's resolve contract changes.

mgoldsborough:
## QA round 3 — adjudication (5405939)

No criticals — thanks for the verification pass. Two of the three suggestions were worth taking:

**Suggestion 1 — admin-gate the rotation case — done.** Agreed the rotation branch (a prior account exists → replace + revoke the shared credential) is destructive like `disconnect` and shouldn't be member-level. Now **only the rotation case is admin-gated**; a first connect (no prior) stays member-level, matching the OAuth connect route. This doesn't create the OAuth asymmetry flagged last round — the OAuth path has no destructive replace+revoke branch, so the file's connect=member / disconnect=admin split stays consistent. Test added: a non-admin member rotating → `permission_denied`, with no connect, the old account intact, and state unchanged.

**Suggestion 2 — assert ACTIVE before `running` — done, at the SDK layer.** `connectComposioApiKey` now throws on a non-ACTIVE resolve instead of the `?? "ACTIVE"` default, so "bundle running" can't drift from the boot-state gate (`status === "ACTIVE"`). The throw routes through the existing delete-the-dangling-account cleanup. Test added: non-ACTIVE resolve → reject + cleanup. (This also closes last round's `?? "ACTIVE"` nit.)

**Suggestion 3 — concurrent double-rotation orphan — leaving as-is.** Per your call: best-effort-cleanup territory, requires concurrent rotation of the same connector, and a lock would be overkill. Noted for completeness.

Suite: 3751 unit pass (0 fail); tsc / biome / cycles clean. The web-deps `@sentry/react` failure is stale `web/` deps (no web/ or sentry code touched) — passes after `cd web && bun install`.

mgoldsborough:
## QA round 4 — adjudication (c4c3793)

**Suggestion 1 — misleading "install first" on a transient start failure — fixed.** The `ensureSourceRegistered` catch now distinguishes the two causes: not-installed (no ref) keeps "must be installed before connecting"; an installed connector whose source transiently fails gets "installed but its MCP source could not start — try Disconnect, then Connect again" (mirrors the OAuth adopt path). Test added for the installed-but-failed branch.

**Suggestion 2 — raw `err.message` in `log.warn` — declined, on technical grounds.** Redaction here is **key-name based** (fields named `api_key`/`secret`/`token`/…), not content-scanning — so moving `err.message` into a structured field wouldn't redact a secret embedded in its text either; it'd just look safer. The actual protections already hold: the user-facing message is generic, submitted field values are never interpolated into logs, and Composio's errors don't echo the key. I'd rather not add motion that implies a guarantee the redactor doesn't provide. (If we want real coverage here it'd be a separate, content-based log-scrubber change across the codebase, not this line.)

Suite: 3752 unit pass (0 fail); tsc / biome / cycles clean.

PR #516, opened 2026-06-23
feat(ci): promote the stable channel on a GA release

mgoldsborough:
## What

A GA release tag now **promotes the stable channel to prod**, completing the release→deploy path for the agent.

After `build-and-push` publishes `runtime`/`web:vX.Y.Z`, a new `promote-to-prod` job dispatches `prod-image-promoted` (`target: agent`, `version:` the tag) to the deploy-config repo's prod-promote receiver, which records the stable image; the per-tenant deployment auto-syncs it.

## Mirrors the existing staging dispatch

Same shape as `ci.yml`'s `roll-staging-edge`: a GitHub App gated on `PROD_PROMOTE_APP_ID` / `PROD_PROMOTE_APP_KEY` + `STAGING_DEPLOY_REPO`, **no-op when unconfigured** so forks and pre-setup CI stay green. Repo/App come from vars/secrets, never literals.

## Scope

- **GA only** — `if: !contains(github.ref_name, '-')`. Pre-releases publish the image but don't touch the stable channel.
- **Still rebuilds** — `NB_VERSION` is baked at build time (for `/about`), and the edge `:sha` image carries no version, so the release can't promote-by-retag; this dispatch points prod at the freshly built `:vX.Y.Z`. (A byte-for-byte retag would require moving the version to deploy-time injection — a separate change.)

## Validation

YAML parses; no literal private names (all via vars/secrets); GA-gate + `target: agent` payload verified.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,7 +162,7 @@ jobs:
   #
   # Fully optional: if the App isn't configured (vars/secrets unset) the job
   # no-ops, so forks and pre-setup CI stay green. Config:
-  #   vars.STAGING_DEPLOY_REPO       — deploy-config repo name (same org)
+  #   vars.DEPLOY_CONFIG_REPO       — deploy-config repo name (same org)
   #   vars.EDGE_BUMP_APP_ID          — GitHub App id (Contents: write on that repo)
   #   secrets.EDGE_BUMP_APP_KEY      — that App's private key
   roll-staging-edge:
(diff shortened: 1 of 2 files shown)


mgoldsborough:
Folded the deploy-repo var rename into this PR (`5c9c74b`): `STAGING_DEPLOY_REPO` → `DEPLOY_CONFIG_REPO` across the new prod-promote dispatch **and** the existing `roll-staging-edge` in `ci.yml`. The var is the single deploy-config repo for both staging and prod, so the `STAGING_` prefix read wrong on a prod path. The new repo var is already set on both repos alongside the old one; platform's two refs are in NimbleBrainInc/platform#120; the old var is retired once both land.

PR #518, opened 2026-06-23
ci: agent release promotes by retag instead of rebuilding

mgoldsborough:
## What

Stop baking the version into the runtime image, and switch the release from **rebuild-at-tag** to **promote-by-retag** — so prod runs byte-for-byte the same image staging verified, and releases stop rebuilding.

- **Dockerfile** — drop the `VERSION` `ARG`/`ENV`. `NB_VERSION` is now injected at **deploy time via env** (the image is version-agnostic). `NB_BUILD_SHA` stays baked — it's a genuine build fact, correct through a retag.
- **`handlers.ts`** — refresh the `/about` `VERSION` comment; the runtime already read `process.env.NB_VERSION || pkg.version`.
- **`web/Caddyfile`** — read `{$NB_VERSION:}` for the web client's Sentry release (was `{$NB_RELEASE:}`). Version and release are one fact, so web uses the same env as the runtime. The `release` JSON field name stays (it maps to Sentry's `release`). Pairs with the deploy side setting `NB_VERSION` on the web container (platform#127).
- **`release.yml`** — replace the verify + build-and-push jobs with a retag job:
  - **ECR** (deploy images): retag the `ci.yml`-built, staging-verified `:<sha>` runtime + web images to `:vX.Y.Z` via `aws ecr put-image`. Guards: immutable repo / source-exists / idempotent. Validated **all-or-nothing across both repos** (pre-flight both, then mutate) so a missing web source can't half-promote runtime.
  - **GHCR** (public OSS pulls): copy the same ECR `:<sha>` manifests to the GHCR tags via `docker buildx imagetools create` (no rebuild). GA also moves `:latest`. `nimblebrain` alias preserved.
  - `workflow_dispatch` gains an **EMERGENCY `rebuild=true`** escape hatch.
  - **No verify job**: `ci.yml` already gated the commit; the source-image-exists guard is the proof.
- **`RELEASING.md`** — refresh §1 (deploy-time `NB_VERSION`, not a build arg) and §7 (the new job list), since this PR makes them stale and `handlers.ts` points at §1.

## Merge gate ⚠️

Depends on the deploy side injecting `NB_VERSION` at deploy time — it **must be live in prod first** (platform#127, **merged + rolled**; verified zero-diff on the runtime-chart tenant). Otherwise a retagged image reports `0.0.0-dev` at `/about` **and** the web client's Sentry release tag goes empty (the Caddyfile now reads `NB_VERSION`). Both surfaces depend on the same env, so the gate covers both.

## Verification

- `release.yml` YAML validated; semver regex accepts `vX.Y.Z` (+ `-prerelease`/`+build`), rejects 4-part (`v0.5.1.7`) and junk; GA-vs-prerelease `:latest` gating and `imagetools` assembly dry-run as expected; two-pass retag confirmed to abort before mutating runtime when the web source is absent.
- Only `src/` change is a comment, so `bun run verify` (biome/tsc) is unaffected — CI green.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,7 +37,10 @@ on:
         default: false
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.inputs.tag || github.ref }}
+  # Normalize across triggers: ref_name is `v0.5.1` on a tag push, matching the
+  # workflow_dispatch `tag` input — so a tag-push and a manual dispatch of the same
+  # version share one group and can't race on the immutable ECR :vX.Y.Z.
+  group: ${{ github.workflow }}-${{ github.event.inputs.tag || github.ref_name }}
   cancel-in-progress: false
 
 permissions:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -78,7 +78,7 @@ jobs:
           VERSION="${{ github.event.inputs.tag || github.ref_name }}"
           # Require vX.Y.Z (optional -prerelease / +build). A garbage version would
           # burn an immutable tag, so reject it before anything is written.
-          if ! [[ "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+([-+.][0-9A-Za-z.-]+)?$ ]]; then
+          if ! [[ "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+([-+][0-9A-Za-z.-]+)?$ ]]; then
             echo "::error::Tag '$VERSION' is not vX.Y.Z (with optional -prerelease)"; exit 1
           fi
           SHA="$(git rev-parse --short HEAD)"
@@ -157,9 +157,17 @@ jobs:
           SRC: ${{ steps.r.outputs.sha }}
         run: |
           set -euo pipefail
-          for REPO in nimblebrain/nimblebrain-runtime nimblebrain/nimblebrain-web; do
-            DST="$VERSION"
+          REPOS=(nimblebrain/nimblebrain-runtime nimblebrain/nimblebrain-web)
+          declare -A MAN ACTION
 
+          # Two passes so the retag is all-or-nothing across BOTH images: validate
+          # every repo's guards FIRST, then mutate. Otherwise a missing web :<sha>
+          # (runtime built green but the web build failed on that ci.yml run) would
+          # leave runtime:vX.Y.Z already retagged while web errors out — a wedged,
+          # immutable, half-promoted version that even rebuild=true can't recover.
+          #
+          # Pass 1 — validate guards for every repo, mutate nothing.
+          for REPO in "${REPOS[@]}"; do
             # Guard 1: the byte-for-byte promise needs an immutable repo.
             MUT="$(aws ecr describe-repositories --repository-names "$REPO" --query 'repositories[0].imageTagMutability' --output text)"
             [ "$MUT" = "IMMUTABLE" ] || { echo "::error::ECR repo $REPO is $MUT, not IMMUTABLE — refusing to promote-by-retag"; exit 1; }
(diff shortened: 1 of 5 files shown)


mgoldsborough:
Added (1b18f8f): the web Caddyfile now reads `NB_VERSION` for the Sentry release instead of the separate `NB_RELEASE` — version and release are one fact, so web uses the same env the runtime does. The `release` JSON field name stays (it maps to Sentry's `release`).

Order-safe via a transitional dual-set on the deploy side (the chart keeps emitting `NB_RELEASE` too, same value, until all web images carry this Caddyfile), so there's no empty-release window during rollout. No new merge gate beyond the existing one.

mgoldsborough:
Simplified (ea654fc): the Caddyfile just reads NB_VERSION now — no transitional NB_RELEASE fallback. Pairs with the chart setting NB_VERSION (platform#127). Shipping together, so no shim needed.

mgoldsborough:
Adjudicated — pushed 01442cc.

**Critical 1 — RELEASING.md staleness → fixed.** Confirmed both spots, and that `handlers.ts` points at the now-stale §1 (self-referential). Updated §1 (version injected at deploy time via `NB_VERSION`, image version-agnostic) and §7 (rewrote the job list: promote-by-retag with the three guards + GHCR imagetools copy + emergency `rebuild=true`, then GitHub Release, then prod stable promote — also folded in the prod-promote job §7 was already missing). Artifact-map table left as-is — still accurate, as you noted.

**Suggestion 1 — partial-promotion wedge → fixed.** Real footgun: a missing web `:<sha>` would leave `runtime:vX.Y.Z` retagged + immutable + unrecoverable (rebuild=true would conflict on the existing runtime tag). Restructured to two passes — validate guard-1/guard-2/idempotency for **both** repos first, mutate only in pass 2. Dry-ran the missing-web-source case: it aborts in pass 1 before any `put-image`. Worth the small restructure on the release path.

**Suggestion 2 — PR body omits Caddyfile → fixed.** Added the `NB_RELEASE`→`NB_VERSION` rename to *What* and widened the gate note (web Sentry release goes empty too, not just `/about` — same env).

**Suggestion 3 — regex → partially fixed.** Dropped `.` from the prerelease lead char-class, so `v0.5.1.7` is now rejected (valid `-rc.1`/`+build` still pass) — that `.` was a genuine bug in my regex. Left the leading-zero case (`v01.2.3`) alone: full strict-semver adds real regex complexity for a tag no human writes, and you flagged it low-priority.

**Suggestion 4 — SHA-length heuristic → acknowledged, not changing.** It's the existing `ci.yml` convention (not introduced here), it fails **safe/loud** (guard-2 errors, never silently mistags), and both workflows checkout shallow (depth 1 → `git rev-parse --short HEAD` is always 7 chars, can't drift). Switching only `release.yml` to `${GITHUB_SHA::7}` would *introduce* a divergence risk vs. `ci.yml`; doing it right means changing both in lockstep, which is `ci.yml` scope creep for this PR. Left as the accepted convention.

mgoldsborough:
Adjudicated — pushed 19097fa.

**1 — concurrency key not normalized → fixed.** Good catch, and it was also inconsistent with the rest of the file (which already uses `inputs.tag || github.ref_name`). Changed the group to `github.event.inputs.tag || github.ref_name` so a tag-push and a manual dispatch of the same version share one group. Not just race-avoidance — `cancel-in-progress: false` means the second now queues behind the first and then hits the idempotency guard (same digest → skip), instead of both firing at once.

**2 — `--short=7` SHA coupling → agreed, deferring as you scoped it.** This is the same contract from last round; it's fail-loud and stable today (both checkouts are `fetch-depth: 1` → always 7 chars). `--short=7` in both `ci.yml` + `release.yml` is the right way to make it explicit and depth-proof, but it's a paired change and out of scope here, as you said. I'll cut a tiny follow-up PR for the pair rather than half-do it here.

**3 — `-rc` dry-run before prod GA → adopting it as the rollout plan.** Exactly right: a prerelease skips `promote-to-prod` and `:latest`, so a `workflow_dispatch` on a throwaway `-rc` tag exercises the ECR retag + GHCR imagetools copy end-to-end without touching prod. That's now the first step when we cut the release (after this + the infra ECR-perms PR land).

Issue #96, opened 2026-04-27
Reduce input tokens: convert automation-authoring.md from always-on context to triggered skill


Ovaculos:
Two structural changes that meaningfully reduce baseline token cost but have tradeoffs to evaluate before shipping.

## 1. Convert `automation-authoring.md` from context to skill

`automation-authoring.md` is `type: context, priority: 16`, so it's always merged into the system prompt on every request — adding ~900 tokens to every call, even 'what's the weather?'. Its content (cron syntax, token budget guidance, confirmation flow) is only relevant when the user asks about scheduling or automations.

**Fix:** Change `type: context` → `type: skill` and add triggers/keywords so it fires only when relevant. Suggested triggers: `schedule`, `automate`, `recurring`, `every day`, `every hour`. Keywords: `schedule, automate, cron, recurring, interval, automation`.

**Tradeoff:** If triggers are too narrow, the automation guidance won't activate when it should. Needs testing across the range of ways users ask about automations before shipping. `bootstrap.md` stays as `type: context` — its search-first tool discovery rules are needed on every call.

## 2. Never include MCP bundle tool schemas in the base tool set

Up to 30 tool schemas from all installed bundles are included in every LLM call regardless of whether the conversation uses them. MCP bundle tools are specialized — most conversations won't touch a given bundle — but their full JSON schemas still occupy 200–700 tokens each.

**Fix:** Filter out `McpSource` tools from `modelTools` in the runtime. MCP bundle tools stay in the router for execution. `hasProxiedTools` becomes always-true when any MCP bundle is installed. The agent uses `nb__search` to discover bundle tool names (already instructed by `bootstrap.md`), then calls them by name. The router executes normally — `nb__search` returns full schemas so the model can construct valid inputs.

**Tradeoff:** The agent now needs one `nb__search` round-trip before it can call any bundle tool. Multi-bundle workflows that previously called tools directly will add one extra iteration. Needs evaluation of whether the token savings outweigh the latency cost across typical usage patterns.

Ovaculos:
## Root cause

Two things inflate the token baseline on every request even for simple conversations that don't need them:

[...]

PR #520, opened 2026-06-23
feat(web): API-key connect form for Composio connectors

mgoldsborough:
## What

Adds the web UI so a client can **Install an API-key Composio connector and enter its credentials in-app** — the missing front end for the `connect_api_key` backend that landed in #515. Until now, clicking Install on an `authScheme: "API_KEY"` connector (e.g. PostHog) routed into the OAuth redirect path, which has no redirect to follow, and failed.

## How

- **`client.ts`** — the server already ships `authScheme` + `fields` on `install.composio` / `catalog.composio`; widen the two web types to expose them (new `ComposioField` / `ComposioConfig`). Add `connectComposioApiKey(catalogId, fields)` → `manage_connectors.connect_api_key`.
- **`ComposioApiKeyModal.tsx`** (new) — a field-driven form rendered from the connector's declared `fields`: titles, descriptions, password inputs for `sensitive`, required-field validation mirroring the server's `required !== false` default. Reuses the `OperatorSetupModal` chrome. Submits via an `onClick` handler (Enter-to-submit preserved via a keydown handler) rather than native form submission.
- **Connect flow branch** — in `ConnectorBrowsePage` (post-install) and `ConnectorStatusHero` (connect/reconnect), when `authScheme === "API_KEY"` open the modal instead of `initiateComposioOAuth`. First connect is member-level; reconnect/rotation is admin-gated server-side and the error surfaces inline.

## Trust posture

Unchanged from the backend: field values flow to `connect_api_key` → Composio and are **never persisted** by the platform (only the opaque `connectedAccountId`). `sensitive` fields render as password inputs and aren't logged.

## Testing

- `bun run verify` green (static + 3763 unit + web + smoke).
- New tests in `connector-sections.test.tsx`: `connectComposioApiKey` wiring, modal field rendering (+ password for sensitive), required-field validation (no call on empty), and filled-submit calls the helper with the right `(catalogId, values)` + fires `onConnected`.

## Notes

- A small UX detail worth a look: the modal intentionally is **not** a `<form>` — happy-dom crashes on `requestSubmit`/`checkValidity`, and we do validation in JS, so the primary button drives submit via `onClick` with Enter handled per-input. Same end-user behavior, testable.
- This unblocks **deployments#96** (the PostHog catalog entry), which was held in draft pending this form. Sequence after merge: roll the runtime image → undraft + merge #96 → provision `COMPOSIO_POSTHOG_AUTH_CONFIG_ID` per env.
- Validated end-to-end locally against real PostHog via the same `connect_api_key` action this form calls (install → connect ACTIVE → live tool read).

### DIFF
--- a/docs/src/content/docs/config/connectors-catalog.mdx
+++ b/docs/src/content/docs/config/connectors-catalog.mdx
@@ -166,10 +166,39 @@ NimbleBrain-specific fields under `_meta["ai.nimblebrain/connector"]`:
 | `requiredScopes` | no | List of OAuth scope strings. |
 | `additionalAuthorizationParams` | no | Extra authorize-URL params (e.g. Google's `access_type: offline`). Reserved keys (`client_id`, `state`, `redirect_uri`, `response_type`, `code_challenge`, `code_challenge_method`, `scope`, etc.) are rejected. |
 | `operatorSetup` | required when `auth: static` | `{ portalUrl, hint, clientSecretKey }`. |
-| `composio` | required when `auth: composio` | `{ toolkit, authConfigEnv, tools? }`. `tools` is an optional allowlist of upstream tool names — strongly recommended to keep agent context bounded. |
+| `composio` | required when `auth: composio` | `{ toolkit, authConfigEnv, tools?, authScheme?, fields? }`. `tools` is an optional allowlist of upstream tool names — strongly recommended to keep agent context bounded. `authScheme` defaults to `OAUTH2` (the redirect flow); set it to `API_KEY` for toolkits that authenticate by key (see below). `fields` is required for `API_KEY` — the inputs collected from the user at connect. |
 | `tags` | no | Strings used for filter / search in the Browse page. |
 | `interactive` | no | Sets the "Interactive" badge. |
 | `docsUrl` | no | Connector-specific docs link surfaced on the Configure page. Must be http(s). |
+
+### auth: composio with API keys
+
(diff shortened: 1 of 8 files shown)


mgoldsborough:
## QA adjudication — all three suggestions deferred to a follow-up (no #520 changes)

Thanks for verifying the load-bearing facts (server ships `authScheme`/`fields`, wire contract matches, both dispatch sites guarded, server-side allowlist is the real gate). No criticals; agreed on all of it.

The three optional suggestions share one root — the connector modals now triplicate the same chrome — so they're best fixed together, not piecemeal in this PR:

1. **Extract a shared `<Modal>` shell** — agreed it's the signal, but extracting now would scope-creep this PR into refactoring `OperatorSetupModal` + `BundleCredentialsModal` + their tests. Adding the 3rd copy is the right call here (matches precedent).
2. **a11y `noStaticElementInteractions` on the overlay** — matches both sibling modals, and `web/src/` isn't in `bun run lint`'s scope, so fixing only here would *diverge* from the siblings. Fix once in the shell.
3. **Same-tick double-submit window** — real but negligible (button disabled on `busy`; server re-validates; same property as both siblings and the already-accepted server-side concurrent-rotation case). Add a synchronous `ref` guard once, in the shell.

Fixing 2 and 3 only in the new modal would create exactly the asymmetry the shell extraction avoids. Tracked together in **#522** so they're not dropped. Leaving #520 as-is.

PR #528, opened 2026-06-24
feat: connector ↔ skill binding (surface-once-into-history) [P4]

mgoldsborough:
## P4 — Connector ↔ Skill Binding (v1)

Installing a connector now automatically applies a curated **usage overlay** for its tools — relevant (only when the connector's tools are used) and cache-friendly (it never re-varies the cached system prefix). This is the SPEC's open "move conditional skill bodies into history" item.

### The two locked design axes
- **Source:** the connectors that need overlays most are the ones we don't author (Composio + third-party). v1 sources from a NimbleBrain-curated **public** overlay repo, keyed by connector identity (`composio/<toolkit>` else server name), at a pinned version.
- **Loading model:** **dynamic + tool-affinity, surfaced ONCE into the conversation history** on the first matching tool call, deduped if already present. NOT a stable system-prefix tier; NOT per-turn prefix re-injection.

### Commits (ordered)
1. **`07efb90` — surface-once-into-history mechanism (the centerpiece).** New `connector.skill.injected` engine event; the engine matches a run's connector-overlay candidates by tool-affinity and emits the event on first match (deduped across runs via a history scan for the synthetic marker, and within a run via a set). The reconstructor turns it into a synthetic assistant message wrapped in `<connector-skill>` containment — the Anthropic provider merges it into the preceding assistant block and a user turn always follows on replay, so it never becomes a trailing prefill and role alternation stays valid. `EngineConfig.connectorSkillCandidates` carries the per-run pool; candidates never reach `composeSystemPrompt`.
2. **`08a37a0` — identity + provenance + config.** `connectorSkillIdentity` / `connectorSkillIdentityFrom`; `SkillProvenance.source`; `src/config/connector-skills.ts` (pinned repo coords + opt-in `CONNECTOR_SKILLS_ENABLED`, env-driven). The resolver (landed earlier as `f7ad130`) sources its defaults here.
3. **`4d36b1f` — candidate store + loader + routing.** `connector-skills/<server>/<skill>.md` store (a SIBLING of `skills/`, never read by the authored loader); `Runtime.loadConnectorSkillCandidates` threads the pool into both `engine.run` sites; `skills__list` filters `origin: connector` as defense-in-depth.
4. **`b22b99d` — bind at install + dedupe + `list_bound_skills`.** `BundleLifecycleManager.syncBoundSkills` / `removeBoundSkills` (best-effort, NON-FATAL, injectable fetch); the remote/Composio install path records the returned `skillsLock` on the persisted `BundleRef`; `loadBundleSkills` skips a server's `skill://usage` when an overlay is bound; `manage_connectors list_bound_skills`.
5. **`fea931d` — docs + CHANGELOG.**

### Notable adaptations from the plan (codebase reality)
- **Install hook lives where the connector shape is known.** `ServerDetail` is not in scope inside `lifecycle.ts`'s install methods; the remote/Composio install completes in `handleInstallRemoteOAuth`, which has the toolkit + server name directly. `syncBoundSkills` is a lifecycle method (per the plan) called from there. Named/in-bundle connectors are the deferred fast-follow per the locked decisions, so only the remote path binds in v1 (no `installNamed` / respawn hook needed yet).
- **`syncBoundSkills(identity, …)` not `(detail, …)`** — the caller computes the identity; the lifecycle manager stays decoupled from connector shapes.
- **Binding state is disk-authoritative** (the materialized `connector-skills/` dir); `skillsLock` rides the ref for restart-survivable dedupe + cleanup traceability.

### Tests (hermetic — fixtures, no network)
- Engine surfacing + cross-run/per-run dedup + no-match + inert-without-candidates; reconstruction into a synthetic message with containment + alternation; `<connector-skill>` containment-escape; an engine+event-store integration test proving surface-once, present-next-turn, no-re-injection, never-in-system-prefix.
- Identity derivation; config resolution (opt-in / override / fail-closed); store materialize/load/remove; lifecycle sync/remove (enabled / disabled / 404 no-op / fetch-error non-fatal / removal) with an injected fetch; a Runtime-level lifecycle test (bind → candidate loaded → listed → absent from `skills__list` → removed).

### Gates
`bun run verify` (incl. codegen drift), `test:integration`, `smoke`, and the docs build (internal links valid) all green.

### Out of scope (follow-ups)
- Creating + seeding the public `NimbleBrainInc/connector-skills` repo (separate, operator-confirmed step). Until it exists and `CONNECTOR_SKILLS_ENABLED=true`, this is fully dormant (fail-closed).
- Generalizing surface-once-into-history to `matched_skill` / `skill://usage`; named/in-bundle connector sources; everything in the plan's Deferred section.


mgoldsborough:
Thanks — sharp review. Addressed in `763e197`.

**Critical (workDir divergence on uninstall) — fixed.** You're exactly right: install (`runtime.getWorkDir()`) and the per-turn loader (`WorkspaceContext`) honor `config.workDir`, but step-4d cleanup used `defaultWorkDir()`, so under a `nimblebrain.json` `workDir` with `NB_WORK_DIR` unset they diverged and uninstall removed nothing. The lifecycle manager now takes the runtime's resolved workDir via `setWorkDir` (wired in `Runtime.start` with `resolveWorkDir(config)`, mirroring the existing `setPlacementRegistry`/`setBundleMcpDepsFactory` injectors), and step-4d uses it. New regression test drives the **real** `uninstall` path (not a direct `removeBoundSkills` call) with a workDir provably `≠ defaultWorkDir()`, so the divergence can't come back. (Scoped to the connector-skill cleanup; the sibling credential/OAuth cleanup's use of `defaultWorkDir()` is pre-existing and left untouched.)

**Suggestions:**
1. **Reactive timing** — added a sentence to the Skills doc noting the first matching tool call is unguided (overlay takes effect from the next turn) and pointing preemptive-precondition cases at an authored `always` skill / tool-side guard.
4. **Resolver cache key** — added the "pinned versions are immutable; moving a tag serves the cached body until the cache dir is cleared" comment.
2 (per-turn re-read caching) and 3 (single-value `scope`) — left as-is per your "if it ever matters" / "fine to keep"; the bound is connector-count-small and `scope` is deliberately forward-looking.

Gates re-run green: `verify`, `test:unit` (3798), `test:integration` (598), `smoke`, docs build (links valid).

mgoldsborough:
Adjudicated — `be045ee`. The Critical is real and now fixed + locked with a runtime-level test; one of the two suggestions doesn't hold (details below).

## Critical #1 (rehydrate strips metadata → re-injects every turn) — VALID, fixed
Confirmed against source: `rehydrate.ts:73` destructures `{ role, content, providerOptions }` from non-user messages, dropping `metadata` — so the synthetic message's `synthetic`/`skill` marker is gone before the engine's history scan runs on the `runtime.chat` path. My engine+store test missed it precisely because it passes `store.history()` straight to `engine.run`, bypassing rehydrate. Good catch.

Took your **preferred** fix (keeps rehydrate's "exactly `LanguageModelV3Message`" contract): the runtime computes the already-injected set from the UN-rehydrated `compactedHistory ?? history` and passes it via a new `EngineConfig.alreadyInjectedConnectorSkills`; the engine seeds from it. The history scan stays as a fallback for the direct-call (test) path.

**New regression test** drives two real `runtime.chat` turns (each calling a matching connector tool) and asserts `connector.skill.injected` fires exactly once. I verified it **fails without the fix** (length 2) and passes with it — so the runtime-layer gap you flagged is now covered.

## Delete #1 (`ConnectorSkillLockEntry.path` unused for removal) — VALID, doc corrected
Right — cleanup is keyed on `serverName` (removes the whole `connector-skills/<server>/` dir), never reads `path`. Kept the field (traceability/debugging) and fixed the comment to stop claiming it drives removal.

## Delete #2 (`expectedSha` no production caller) — acknowledged, kept
Fair — `syncBoundSkills` always fetches fresh and never passes `expectedSha`. It's the integrity mechanism for the recorded sha (re-resolve verification) and is tested; I'd rather keep the defensive scaffolding than strip it for "no caller today." Noted as a follow-up surface, not removed.

## Suggestion #1 (overlay leaks into conversations__get/export/fork) — does NOT hold
Checked this against source and it doesn't reproduce: the display path uses `jsonl-reader.ts`, **not** `reconstructMessages`. Its `isRelevantLine` filter (`jsonl-reader.ts:380-389`) only admits `user.message` / `run.start` / `run.done` / `llm.response` / `tool.start` / `tool.done` / `metadata.*` — `connector.skill.injected` isn't in the set, so the event is skipped before the reducer and never becomes a `DisplayMessage`, and fork indices (`conversation.messages.slice`) are unaffected. The reviewer's read assumed `conversation.messages` came from `reconstructMessages`; it's the jsonl-reader projection. Only `reconstructMessages` (the LLM-replay projection) materializes the synthetic message — which is exactly where it's needed. No change.

## Suggestion #2 (consecutive assistant messages → real-provider check) — acknowledged
Verified `@ai-sdk/anthropic`'s `groupIntoBlocks` merges consecutive same-role messages, and a user turn always follows on replay so it's never a trailing prefill. Didn't add an eval (non-hermetic, needs a key); the merge behavior is confirmed in the provider source.

Gates re-run green: `verify`, `test:unit` (3798), `test:integration` (599, +1 = the new dedup test), `smoke`, docs build.

mgoldsborough:
Round 3 addressed — `282fd07`. No Critical; took all four suggestions (two as code fixes, two as pinned-intent comments).

**#1 — scrubbed the 34 `(P4)` phase markers from shipped `src/` comments.** Agreed: planning lineage that means nothing to an external contributor and cuts against this repo's "represent what is, not what was" + OSS-hygiene rules. Comments now describe the behavior directly; phase tags stay in commit messages. (Left the pre-existing, unrelated `ARCHITECTURE P4` in `url-validator.ts` alone.)

**#2 — empty-body overlay → dedup never engages: fixed, made symmetric with the store.** Good catch on the two-sources-disagree shape. `materializeConnectorSkill` now returns `null` on an empty body (treated as "no overlay," like an unparseable one), and `readConnectorSkillCandidates` skips empty-body files as defense for any pre-existing one. So an empty overlay never becomes a candidate, never emits, never re-loops. Test added (`connector-skill-store.test.ts`).

**#3 — non-Anthropic consecutive assistant messages: pinned the assumption.** You're right that the merge is the provider's job and the PR only reasons about Anthropic. Added a provider note at the reconstruction append site: the synthetic message produces two consecutive assistant turns that rely on Anthropic's same-role coalescing (the default + only enabled-for provider); verify provider-side coalescing before enabling overlays on a non-Anthropic default. Didn't add a test (no hermetic way to exercise a real non-Anthropic provider's message conversion).

**#4 — compaction interaction: pinned the intent.** Documented on `collectInjectedConnectorSkills`: reading from `compactedHistory ?? history` means once compaction folds the marker into a summary, the overlay re-surfaces once on the next matching call — re-establishing guidance after the verbatim block was summarized away, bounded to one re-injection per compaction. Confirmed desirable, now explicit.

Gates green: `verify`, `test:unit` (3799), `test:integration` (599), static, docs build.

mgoldsborough:
Rebased onto `main` to clear the merge conflict (`6c0de60`). Two commits had landed since this branch forked:

- **#527** (admin "rules" containment) — conflicted in `test/unit/prompt-injection.test.ts`: both it and this PR added a `describe("1.15 …")`. Kept both; renumbered the connector-skill block to **1.16**. `compose.ts` / `writer.ts` auto-merged (my `provenance.source` line + #527's additions both present).
- **#531** (connector `serverInfo.version`) — auto-merged in `connector-tools.ts` (different region from the install-hook + `list_bound_skills` action); verified both changes coexist.

No behavior change from the rebase — pure conflict resolution. Full gates re-run green on the rebased tree: `test:unit` (3809), `test:integration` (612), `verify:static`, docs build (links valid). Branch is now even with `main`.

PR #532, opened 2026-06-25
fix(mcp): log ui:// resource-read failures instead of masking them as 404

mgoldsborough:
## Problem

A `ui://` resource read (the `/v1/apps/:name/resources/*` proxy that mounts an app's UI) could return a silent **404 `resource_not_found`** when the backing MCP source's connection degraded — e.g. after a deploy/roll — and stay that way until the **entire runtime was restarted**. There was no log to say why, because the error was thrown away.

## Root cause

`McpSource.readResource` had a bare `catch { return null }`. It collapsed two different outcomes into one answer:

- the resource is **genuinely absent** (expected — a `skill://` / `app://instructions` probe against a server that has neither), and
- the read itself **failed** (torn transport, dropped connection, timeout).

Both became `null` → a 404, with nothing logged. So a degraded source was undiagnosable.

## Fix (scoped deliberately small)

Classify the error in the catch:

- a genuine MCP miss (`-32002` / `-32601` / `-32602`, or a "not found" message) stays a **silent null** — `isMcpResourceMiss` is an allowlist with everything-else-falls-through;
- a real failure is **logged** (`[mcp] readResource(...) failed: ...`).

The 404 is **unchanged** — this only makes the cause visible.

### Logging is scoped to the app-surface path (no probe spam)

The log fires **only** on the resource-proxy path (`readResourceFromSource`, whose sole callers are `readAppResource` / `readIdentityAppResource` serving a mounted app's `ui://` resource), where a failure is anomalous and low-volume. **Discovery probes** (`skill://`, `app://instructions`, facets) read a source directly and stay silent — because absence there is the *normal* case, runs against every bundle on every turn, and isn't actionable. This holds **regardless of the error code a bundle returns** for an unknown resource (which isn't standardized across MCP server implementations), so a bundle that simply lacks a probed resource can never spam the log.

## Why not the self-heal too?

The resource-read path also lacks the self-heal the tool-dispatch path has (`recoverWorkspaceSource` / restart a degraded source in `src/orchestrator/route.ts`). That asymmetry is real — but **I have not confirmed which failure mode caused the incident** (the masking destroyed the evidence, and the source was restarted before it could be captured). The plausible mechanisms — torn transport, a source that didn't re-register its resources after a roll, or a boot-time connection failure — call for **different** fixes, and a speculative recovery branch is extra kernel surface for a guess. This PR adds the log that will identify the real mechanism; the targeted recovery follows once it's confirmed. Tracked in #533.

## Tests

`test/unit/mcp-source-resource-errors.test.ts`: the `isMcpResourceMiss` allowlist (miss codes / message → miss; transport errors → not a miss; null-safe), and the log scoping — a transport failure is **silent** on a probe read but **logged** on the app-surface path, a genuine miss is silent on both, and `readResource` never throws (the null/404 contract holds for every caller).

`bun run verify` passes (static + unit + bundles + integration + smoke, 0 failures).

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -3946,16 +3946,22 @@ export class Runtime {
   ): Promise<ResourceData | null> {
     if (!(source instanceof McpSource)) return null;
 
+    // This is the app-surface (resource-proxy) read path — its only callers are
+    // readAppResource / readIdentityAppResource serving a `ui://` resource for a
+    // mounted app. A failure here is anomalous, so opt into logging; discovery
+    // probes that read directly from a source stay silent.
+    const opts = { logFailures: true } as const;
+
     if (resourcePath.includes("://")) {
-      return source.readResource(resourcePath);
+      return source.readResource(resourcePath, opts);
     }
 
     const exactUri = `ui://${resourcePath}`;
     const namespacedUri = `ui://${appName}/${resourcePath}`;
 
-    const result = await source.readResource(exactUri);
+    const result = await source.readResource(exactUri, opts);
     if (result !== null) return result;
-    if (exactUri !== namespacedUri) return source.readResource(namespacedUri);
+    if (exactUri !== namespacedUri) return source.readResource(namespacedUri, opts);
     return null;
   }
 
--- a/src/tools/mcp-source.ts
+++ b/src/tools/mcp-source.ts
@@ -1292,8 +1292,22 @@ export class McpSource implements ToolSource {
    * spec attaches ui metadata at the content level, so that's the load-bearing
    * source for iframe CSP / permissions / layout hints.
    */
-  async readResource(uri: string): Promise<ResourceData | null> {
-    if (!this.client) return null;
+  async readResource(uri: string, opts?: { logFailures?: boolean }): Promise<ResourceData | null> {
+    if (!this.client) {
+      // A torn-down client (a source degraded/crashed without re-creation)
+      // returns null HERE, before the catch below — the persistent silent-404
+      // shape this change exists to surface, and the most likely cause of a read
+      // that stays broken until a full runtime restart. Log it on the app-surface
+      // path; the probe path stays silent because a not-yet-started or tearing-down
+      // source legitimately has no client during composition.
+      if (opts?.logFailures) {
+        log.warn("[mcp] readResource: source has no active client connection", {
+          uri,
+          source: this.name,
+        });
+      }
+      return null;
+    }
     try {
       const result = await this.client.readResource({ uri });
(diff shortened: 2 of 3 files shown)


mgoldsborough:
Thanks — this review changed the PR. After re-examining, I couldn't actually confirm the failure mechanism: the masking destroyed the evidence and the source was restarted before it could be captured. The self-heal orchestration was a hypothesis-driven response to an *unconfirmed* degraded-source theory, and (as you noted) it's a lot of new kernel surface — some of which my own classifier could make a no-op for the real incident (a genuine `-32002` post-roll is classified as a miss → never heals).

So I trimmed to the part I'm certain about: **stop masking, log the real error.** The 404 behavior is unchanged; the cause is now visible. That converts the next occurrence from "blind 404 forever" into a logged code/message that tells us which fix is actually warranted.

On your specific points:

- **Critical #1 (untested heal orchestration):** resolved by removal — there's no orchestration in this PR anymore. The remaining logic (the `isMcpResourceMiss` allowlist + that `readResource` never throws) is unit-tested directly.
- **Suggestion #1 (restart tears down a live source):** good catch, and moot here since no `restart()` is added — but I captured it in the follow-up (gate restart on `!source.isAlive()`) so it's baked into the recovery when we add it.
- **Suggestion #2 (concurrent `restart()` de-dup):** also moot for this PR; noted in the follow-up for when restart-on-read lands.

The deferred self-heal parity is tracked in https://github.com/NimbleBrainInc/nimblebrain/issues/533.

mgoldsborough:
Thanks — and especially for tracing all four resource producers to confirm the allowlist holds; that's the validation I couldn't do without the masking already removed. Both suggestions taken.

**Suggestion 1 (torn-down client not covered) — fixed, and you're right that it's the one that matters.** The `if (!this.client) return null` short-circuit was silently swallowing exactly the mechanism my own PR body named (a client torn down after a roll, persistent until restart), so the "log that identifies the cause" would never have fired for it. I went with `log.warn` **scoped to `logFailures`** rather than an unconditional `log.debug`: your reason for debug was that the null-client path is hit normally during startup/teardown — but that's the *probe* path, which the scoping already silences. On the app-surface path a null client is genuinely anomalous, so it warrants a warn that's visible in prod without `NB_DEBUG` on. Added two tests pinning it (silent on the probe path, logged on the app-surface path).

**Suggestion 2 (structured fields) — taken.** Loki-queryability is the whole point of this PR, so both warns now emit `{ source, uri[, error] }` as indexed fields — an operator can filter by app. Worth the small divergence from the interpolating neighbors here.

Dropped-nit ack: agreed on leaving the raw `-32002` literals (no named SDK member; the commented literals read better than a mixed enum/literal).

Re-verified green.

PR #535, opened 2026-06-25
feat(model): provider-abstracted tool disclosure (ToolDisclosureStrategy)

mgoldsborough:
## What
Adds a per-provider **tool-disclosure** seam, `src/model/tool-disclosure.ts`, mirroring the existing `CacheStrategy` in `cache-policy.ts`. Disclosure decides *which* tools are sent and where the eager (cacheable) prefix ends; cache-policy decides *where* the breakpoints land. They compose through a single scalar, `eagerCount`.

## Why
The prompt cache is a prefix (`tools → system → messages`), so the tools block must be append-only. A provider that discloses extra tools mid-conversation (e.g. surfacing a discovered tool) must add them *after* the cacheable eager prefix, or it busts tools + system + all history. The cache-policy previously pinned the 1h tools breakpoint on the literal last tool — correct only while every tool is stable, wrong once volatile tools are appended.

## Changes
- `src/model/tool-disclosure.ts` (new) — `ToolDisclosureStrategy` + per-provider registry + passthrough default + `applyToolDisclosure`.
- `src/model/cache-policy.ts` — `CachePolicyInput.eagerToolCount?` (defaults to `tools.length`); the anthropic strategy places the 1h tools breakpoint on `tools[eagerToolCount-1]` (the last **eager** tool), never a volatile one. `eagerToolCount: 0` places no tools breakpoint.
- `src/engine/engine.ts` — runs `applyToolDisclosure` before `applyCachePolicy` and threads `eagerCount` through.
- Tests — new `tool-disclosure.test.ts`; `eagerToolCount` cases in `cache-policy.test.ts`.

## Behavior
**No-op today.** The only strategy is passthrough (send the active set, `eagerCount = length`), and cache-policy defaults `eagerToolCount` to `tools.length`, so breakpoint placement is byte-identical to before. This PR installs the contract; a follow-up adds a defer-and-append strategy (deferred union + tool-search tool) behind a flag.

## Multi-LLM
Provider-abstracted by construction: passthrough is the portable default; providers register their own strategy. No provider-specific assumption leaks into the engine.

## Verification
- `bun run verify:static` — clean (format, lint, tsc, cycles, custom guards).
- `bun run test:unit` — 3825 pass / 0 fail (stable across runs).


mgoldsborough:
## Updated per QA review

**Merged latest `main`** (resolved the import conflict from the TUI/CLI removal).

**Suggestion (clamp) — applied** (`77b45ed`): `eagerToolCount` is now clamped with `Math.min(..., tools.length)`, so a future strategy that passes a count past the array end can't silently drop the tools 1h breakpoint. Added a test.

**Critical (no-op landed ahead of its consumer) — agreed; converting this PR to draft.** Per the review, the seam shouldn't merge as a standalone no-op. Rather than land it now and add behavior later, the first real consumer — a defer-and-append strategy that registers the discoverable union as deferred tools + a tool-search tool, behind a default-OFF `features.toolSearch` flag — will be folded into **this** PR before it's marked ready, so `deferredTools`, the strategy registry, and `eagerToolCount`'s non-default path all land with a live caller doing real work.

Scope note for that consumer: it requires plumbing the cross-workspace discoverable union (today `runtime.listDiscoverableTools()`, not passed to the engine) down to the model-call site, constructing the provider-defined tool-search tool, and widening the tool types — and the `defer_loading` round-trip is validated on a live model behind the flag (the echo-model test harness can't exercise it). That's why it's flag-gated default-OFF.

mgoldsborough:
Closing per the QA reviews (three independent passes concur) and the team decision.

This PR installs a provider tool-disclosure seam but ships no consumer — at runtime it's a verified no-op (empty strategy registry, passthrough identity, eagerToolCount always equals the tool count, deferredTools never populated). No consumer is scheduled, and #537 (workspace-bounded tool reach) changes the foundation it would build on: a session now sees only its focused workspace's tools, so the deferred set would re-source from the workspace-bound router rather than a cross-workspace union, and most sessions have nothing to defer.

Rather than carry a no-op while the consumer is unscheduled and its shape is changing, we'll rebuild this seam together with its first real strategy (defer-and-append) as one behavior-bearing PR — after #537 lands and we confirm a single workspace still exceeds the active-tool budget enough to make deferred loading worthwhile. If it doesn't, the append-only discipline already merged in #534 is sufficient and no seam is needed.

The cache-policy hardening here (the eager/volatile breakpoint boundary + clamp) rides with that future PR, where it has a live caller.

PR #536, opened 2026-06-25
Drop the TUI, CLI subcommands, and the nb binary; launch via bun

mgoldsborough:
## Why

We have no embedded CLI customers — every tenant runs inside the managed runtime, which boots via `serve`. The interactive terminal client (Ink/React TUI) and the operator CLI subcommands were a second extension surface that duplicated the web UI + the agent's tool actions, and carried Commander plus the Ink terminal renderer. Removing them slims the kernel and sharpens the runtime's thesis: the interfaces are the web shell and `/mcp`, not a terminal client.

## What changed

**No `nb` binary.** The runtime is launched with **bun**:
- `bun run start` → `bun run src/cli/index.ts serve` — the managed-services HTTP API server (production entry point)
- `bun run dev` — local watch/HMR loop

`src/cli/index.ts` is now a plain `node:util` `parseArgs` dispatcher for those two commands. The `package.json` `bin: { nb }` entry is gone — nothing used it (the container already runs `bun run src/cli/index.ts serve` directly).

**Removed:** the Ink TUI (`app.tsx`, `markdown.tsx`, `tui-gate.ts`), the Commander program (`program.ts`), and every operator subcommand (`bundle`, `skill`, `config`, `credential`, `status`, `reload`, `telemetry`, `automation`, `user`). Those capabilities are managed from the web UI and the agent's tools (`manage_connectors`, workspace settings, env-var telemetry opt-out). Runtime-emitted operator hints that pointed at those commands were reworded to the web-UI / env-var equivalent.

**Logger relocated:** `src/cli/log.ts` → `src/observability/log.ts`. It's the runtime's structured logger (70+ importers across api/engine/bundles/identity/tools), not a CLI concern — it now sits beside the tracing/identity seam it already depended on.

**Telemetry `mode` removed:** it only switched PostHog flush cadence for short-lived CLI invocations (now all deleted); flush is hardcoded batched and `mode` is gone. Telemetry init is scoped to the `serve` path.

**Dependencies dropped:** `commander`, `ink`. (`react` / `react-dom` + their `@types` **stay** — `src/tools/core-resources/render.tsx` server-renders the Synapse resource shell via `react-dom/server`'s `renderToStaticMarkup`; they are not TUI-only.)

[...]

PR #538, opened 2026-06-25
Make the runtime binary serve-only; move dev orchestration to scripts/

mgoldsborough:
## Why

Follow-up to #536. With the TUI / operator subcommands / `nb` binary gone, the runtime entry was still a 2-command argv dispatcher (`serve` | `dev`). But `serve` and `dev` are different *kinds* of thing:

- **`serve`** = the runtime: one process, the product, what the container runs.
- **`dev`** = a local two-process supervisor (watch-mode serve + Vite HMR) — build tooling.

Mixing them in one binary made the runtime "know about" dev tooling. This makes the runtime binary **serve-only** and moves dev orchestration to `scripts/dev.ts`, next to the existing `scripts/dev-worktree.ts`. The split now matches what the code already is: `serve` = product, `dev` = tooling. (Camp-wise: this moves us from the framework-CLI pattern — `next dev`/`next start` — to the app-server pattern, where the binary serves and dev is a separate tool. We're an app server.)

## What changed

- **`src/cli/index.ts`** → serve-only. No subcommand dispatch. It accepts (and ignores) a leading `serve` token so the deploy command is byte-for-byte unchanged, and parses `--config`/`-c`, `--model`, `--port`, `--debug`.
- **`scripts/dev.ts`** (new) ← the dual-process dev supervisor, moved from `src/cli/dev.ts` (import paths + the resolved CLI-entry path adjusted for the new location; an argv-parsing `main()` added).
- **`src/cli/dev.ts`** deleted.
- **package.json**: `dev` / `dev:empty` / `dev:minimal` / `dev:docs-demo` point at `scripts/dev.ts`. `bun run dev` is unchanged for developers. `dev:api`, `start`, `dev:worktree` untouched.
- Docs (README tree, AGENTS structure line, `cli/overview` usage note) and one stale code comment updated.

After this, `src/cli/` is just the serve surface: `index.ts` (boot), `serve.ts`, `config.ts`.

## Deploy safety

The image/chart command `bun run src/cli/index.ts serve` is **unchanged** — the entry accepts the `serve` token and ignores it. No Dockerfile/Helm change.

One behavior delta: `bun run src/cli/index.ts` with *no* args now boots the server (it previously printed usage and exited 0). No in-repo caller invokes it bare — every deploy path (Dockerfile `CMD`, `start`, `dev:api`, the dev supervisor) passes `serve` — and booting on bare invoke is the correct app-server behavior (the binary's job is to serve). Documented in `cli/overview`.

## scripts/ is now typechecked

The relocation raised the obvious question: was `scripts/` ever typechecked? It wasn't — tsconfig `include` was `src/**` + `instrument/**`, so the codegen, migration, guardrail, and (now) dev-supervisor scripts all ran unchecked. This PR closes that: `scripts/**/*.ts` is added to the tsc program, so `bun run check` / verify:static / CI typecheck them under the same strict options as `src/`. Blast radius across all 19 scripts was a single pre-existing error (an unused destructured var in `check-cycles.ts`), fixed here. Biome-lint scope is unchanged — a separate, optional follow-up.

## Validation

- `bun run verify:static` — green (tsc, lint, all kernel checks, codegen)
- unit 3799/0 · integration 590/0 (incl. the `serve` boot smoke; ~12 env-gated skips) · smoke 21/0
- `scripts/dev.ts` bundles cleanly (relocated imports resolve)
- **Deploy-contract boot smoke:** an integration test boots the deploy command (`serve`) and polls `/v1/health`; mutation-verified (it fails if the `serve`-strip is removed). Parse errors exit 2 (usage), distinct from runtime exit 1.
- serve entry fast-fails correctly **with and without** the `serve` token; rejects unrecognized positionals
- **Live boot:** `bun run scripts/dev.ts --no-web` spawns the watch-mode serve child via the relocated path and reaches `[dev] API ready` (with `[api] Runtime ready` → listening on the chosen port)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,7 +97,7 @@ src/
 ├── prompt/        System prompt composition (identity → core → apps → skill)
 ├── model/         LLM provider registry (AI SDK)
 ├── adapters/      EventSink implementations (logs, console, debug, telemetry)
-├── cli/           Process entry: serve (HTTP API) + dev (watch/HMR) launchers
+├── cli/           Process entry: the serve HTTP API server (dev tooling is in scripts/)
 └── files/         File context extraction
 web/               Vite + React + TypeScript SPA (separate package.json)
 ```
--- a/README.md
+++ b/README.md
@@ -458,12 +458,13 @@ src/
 ├── telemetry/            Anonymous product telemetry
 │   ├── posthog-sink.ts   PostHog event mapping
 │   └── manager.ts        TelemetryManager (opt-in/out, anonymous ID)
-└── cli/                  Process entry: serve + dev launchers
-    ├── index.ts          Entry point (argv dispatch: serve | dev)
+└── cli/                  Process entry: the serve HTTP API server
+    ├── index.ts          Entry point (boots the server)
     ├── serve.ts          HTTP API server boot
-    ├── dev.ts            dev-mode dual-process supervisor
     └── config.ts         nimblebrain.json loading
 ```
+
+Local dev orchestration (watch + web HMR) is tooling, not runtime: `scripts/dev.ts` (`bun run dev`).
 
 ## Deployment
 
--- a/docs/src/content/docs/cli/overview.mdx
+++ b/docs/src/content/docs/cli/overview.mdx
@@ -34,7 +34,7 @@ The server and dev mode share most flags:
 | `--debug` | serve, dev | Verbose event logging to stderr |
 | `--no-web` | dev | Skip the web dev server (API only) |
 
-The working directory is set with the `NB_WORK_DIR` environment variable, not a flag. Running `bun run src/cli/index.ts` with no command prints a one-line usage summary; an unrecognized command prints the same summary and exits with code 2.
+The working directory is set with the `NB_WORK_DIR` environment variable, not a flag. `bun run src/cli/index.ts` boots the server directly; a leading `serve` token is accepted (the deploy command passes it) but optional.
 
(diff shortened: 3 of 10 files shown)


mgoldsborough:
QA addressed (commit `64eeaea`).

**Critical — serve-strip had no test coverage → fixed and mutation-proven.** You were right that a parse-only assertion can't catch this (`serve <garbage>` exits non-zero either way). Added an integration **boot smoke** that runs the real deploy command (`bun run src/cli/index.ts serve --config .environments/empty/nimblebrain.json`) and polls `/v1/health`. Verified it has teeth: breaking the `serve` strip makes the smoke **fail** (server exits 2, health never comes up) while the reject test still passes — i.e. it catches the exact silent-crashloop regression.

**Suggestion 1 — tighten the reject assertion → done.** Arg-parse failures now route through a `parseServeArgs` helper that exits **2** (usage), distinct from runtime exit 1; the reject test asserts `exitCode === 2`.

**Suggestion 2 — biome lint scope on scripts/ → tracking note (not in this PR).** Confirmed the gap: tsconfig now covers `scripts/`, but biome.json's `files.includes` scopes lint/format to `src/` + `instrument/`, so `biome lint scripts/` checks 0 files. Extending it means adding `scripts/` to the biome config and absorbing formatting/lint churn across 19 never-formatted scripts — a separate change with its own blast radius. Filing as follow-up so scripts/ doesn't drift unlinted.

Local: tsc + lint clean; unit 3799/0, integration 590/0 (incl. the boot smoke), smoke 21/0.

PR #551, opened 2026-06-26
security: SSRF redirect guard on remote transport + Secure auth cookies in prod

mgoldsborough:
Two findings from a security best-practices review of the runtime, both prod-reachable and grounded in code.

## H1 — SSRF via remote-MCP transport redirects (High)

The MCP SDK's HTTP transports follow redirects with fetch's default `redirect: "follow"`. A remote bundle URL can be tenant-supplied (a self-service `dcr`/`static` connector install), so a hostile/compromised server can answer the initial request with `30x Location: http://169.254.169.254/...` (cloud metadata / IMDS) or an in-cluster service, and the SDK's fetch silently follows it — turning our outbound fetch into an internal-network probe. `validateBundleUrl` only ran once, on the configured URL at startup; it never saw the redirect target.

**Fix:** `createSsrfGuardedFetch` (`src/tools/ssrf-guarded-fetch.ts`) wraps whatever fetch the transport uses (minting / OAuth-refresh / global) and follows redirects manually, **same-origin only**:
- A cross-origin bounce is refused outright. It is the SSRF pivot, **and** re-issuing the request to a new origin would leak the connection's credential headers (`Authorization` / minted fleet token / API key) to a different trust scope — WHATWG fetch strips those across origins for exactly this reason, and a manual loop must not regress it.
- Same-origin redirects (trailing-slash / path normalization) are followed, with the connection's own opts intact so an operator-vetted in-cluster `http://*.svc` fleet source keeps working. `validateBundleUrl` runs per-hop as a backstop.

This reuses the per-hop-validation pattern already proven in the headless OAuth probe (`workspace-oauth-provider.ts`). New unit suite covers IMDS/loopback/cross-origin rejection, same-origin follow (incl. fleet), and the hop cap.

## H2 — Auth cookies missing `Secure` in production (High)

`ctx.isLocalhost` was derived from `server.hostname` after `Bun.serve`. Bun binds the default `0.0.0.0` in production, and the check read `0.0.0.0` as localhost — so every cookie writer (`secure = !isLocalhost`) dropped `Secure` on `nb_session`, `nb_refresh`, `nb_oauth_state`, and `nb_composio_state` in the standard containerized deployment. The listen address is the production signal, not a localhost signal.

**Fix:** replace the misnamed, bind-derived flag with `ctx.secureCookies`, set once at context construction from `authConfigured`. A configured deployment is always fronted by a TLS-terminating edge (browser↔edge is HTTPS regardless of the container's plain-HTTP listen address), so the cookie carries `Secure`; dev mode (no auth, `http://localhost`) keeps it off. The signal is a deployment property — not the spoofable forwarded-scheme header, not the misleading listen address.

## Review

Reviewed by the chief-architect (kernel/trust-boundary lens) and platform-architect (trust-mesh/edge-topology lens) — both **APPROVE-WITH-CHANGES**, incorporated here:
- **Blocking (chief-architect):** the original loop forwarded credentials across cross-origin redirects. Fixed via the same-origin-only rule above.
- **Cookie signal (platform-architect):** confirmed `secureCookies = authConfigured` is correct for every supported topology — `public-origin.ts` already fails closed at boot for any deployed tenant not on HTTPS, so there is no auth-configured-but-plain-HTTP path. `X-Forwarded-Proto` was considered and rejected (wire value).
- **Fleet mesh:** confirmed no fleet/minted flow relies on following cross-origin redirects (connectors target stable ClusterIP service names directly; the OAuth authorize-302 path uses its own probe, not this transport fetch).

## Deferred — SSRF defense-in-depth (tracked follow-up, not this PR)

The string validator does not resolve DNS, so a public hostname whose A-record points at an internal IP still passes hop 0 (H1b). Per both reviews the authoritative fix is the network layer, not in-kernel DNS pinning:
- **infra:** add `169.254.0.0/16` + `127.0.0.0/8` to the `except` list in the tenant egress NetworkPolicy (`allow-https-egress`) to match the `apps` namespace policy — closes tenant-connector → IMDS.
- **infra:** set EKS node-group `metadata_options` (IMDSv2 required, hop limit 1) as defense-in-depth.
- **nice-to-have:** unify this redirect-guard loop with the OAuth probe's into one shared helper so the policy can't drift.

## Verification

`bun run check` (tsc) clean · biome clean · `bun run verify:static` exit 0 · `bun run test:unit` **3806 pass / 0 fail**.

Pre-merge: a smoke test against a real OAuth remote and one SSE remote is recommended (the guard now sits on those paths); I did not have a live external remote to exercise.

### DIFF
--- a/src/api/handlers.ts
+++ b/src/api/handlers.ts
@@ -1518,7 +1518,7 @@ export async function handleOidcAuthorize(provider: IdentityProvider): Promise<R
 export async function handleOidcCallback(
   request: Request,
   provider: IdentityProvider,
-  isLocalhost: boolean,
+  secureCookies: boolean,
   appOrigin?: string,
 ): Promise<Response> {
   if (!provider.capabilities.authCodeFlow || !provider.exchangeCode) {
@@ -1556,7 +1556,7 @@ export async function handleOidcCallback(
     const result = await provider.exchangeCode(code, pendingFlow.codeVerifier);
 
     const redirectUrl = appOrigin ?? url.origin;
-    const secure = !isLocalhost;
+    const secure = secureCookies;
 
     const sessionParts = [
       `nb_session=${result.accessToken}`,
@@ -1601,7 +1601,7 @@ export async function handleOidcCallback(
 export async function handleOidcRefresh(
   request: Request,
   provider: IdentityProvider,
-  isLocalhost: boolean,
+  secureCookies: boolean,
 ): Promise<Response> {
   if (!provider.capabilities.tokenRefresh || !provider.refreshToken) {
     return apiError(400, "not_configured", "OIDC auth not configured");
@@ -1624,7 +1624,7 @@ export async function handleOidcRefresh(
   try {
     const result = await provider.refreshToken(refreshToken);
 
-    const secure = !isLocalhost;
+    const secure = secureCookies;
     const sessionParts = [
       `nb_session=${result.accessToken}`,
       "HttpOnly",
--- a/src/api/routes/auth.ts
+++ b/src/api/routes/auth.ts
@@ -21,12 +21,12 @@ export function authRoutes(ctx: AppContext) {
 
   app.get("/v1/auth/callback", (c) => {
     if (!ctx.provider) return apiError(400, "not_configured", "Auth provider not configured");
-    return handleOidcCallback(c.req.raw, ctx.provider, ctx.isLocalhost, ctx.appOrigin);
+    return handleOidcCallback(c.req.raw, ctx.provider, ctx.secureCookies, ctx.appOrigin);
   });
 
   app.post("/v1/auth/refresh", limit, (c) => {
     if (!ctx.provider) return apiError(400, "not_configured", "Auth provider not configured");
-    return handleOidcRefresh(c.req.raw, ctx.provider, ctx.isLocalhost);
+    return handleOidcRefresh(c.req.raw, ctx.provider, ctx.secureCookies);
   });
 
   // --- Authenticated ---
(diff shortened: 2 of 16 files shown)


mgoldsborough:
Deferred SSRF defense-in-depth (egress NetworkPolicy + IMDSv2 + loop unification) tracked in #552.

mgoldsborough:
Thanks for the QA pass — tracing the guard through all three fetch rails (OAuth-refresh + minting/fleet both preserve `redirect`) is exactly the check that matters. Adjudication of the four suggestions:

**1. `http→https` in the redirect-hop comment — accepted, fixed (b6093f1).** Correct: the same-origin rule refuses scheme changes, so an http→https upgrade is *not* a hop the guard follows. Comment now reads "same-origin hops (trailing-slash / path normalization — a scheme or host change is cross-origin and refused)."

**2. Smoke-test a *redirecting* remote — accepted as pre-merge guidance.** Agreed this deliberate behavior change deserves a targeted check, not just a clean-remote run. Pre-merge plan: point at a remote that issues (a) a same-origin trailing-slash 301 → must follow and connect, and (b) a cross-origin 30x → must refuse. A bundle relying on a vanity-host cross-origin redirect will now break and must be reconfigured to its final URL; that's the intended trade (SSRF pivot + cross-origin credential leak).

**3. 106-line loop vs `redirect: "error"` — affirmed, no change.** Conscious trade. `redirect: "error"` (refuse all) is smaller but not equal-correctness — it breaks legitimate same-origin trailing-slash/path-normalization on first connect, forcing every such user to reconfigure. The chief-architect review explicitly weighed reject-all vs same-origin-follow and chose the latter for exactly that reason. Keeping it.

**4. Real-auth-over-`http://localhost` now gets Secure cookies — accepted, doc'd (b6093f1).** Correct, and it surfaced an imprecision in my own comment: `Secure` is dropped on a *non-localhost* http origin, not on localhost (modern browsers treat localhost as a secure context and accept Secure cookies there), so the real-auth-over-localhost case works. The `secureCookies` doc now states both.

Net: two doc fixes landed; no logic change (your "0 critical" stands). The H1b infra follow-up (egress NetworkPolicy + IMDSv2) remains tracked in #552.

PR #555, opened 2026-06-26
test: tolerate truncated concurrent stream rejections

### LINKED ISSUE #511: Flaky: chat-stream-concurrent test fails on truncated SSE under load
## Summary

`test/integration/chat-stream-concurrent.test.ts` — *"POST /v1/chat/stream — concurrency protection > concurrent stream requests produce exactly one successful run; the rest are rejected"* — flakes intermittently in CI.

## Symptom

The test fires 5 concurrent `POST /v1/chat/stream` for the same conversation and asserts that **every rejected request** carries the stable error code `"run_in_progress"` (line ~186). Under CI load, one of the concurrent SSE streams is occasionally **truncated** — the runtime logs `JSON Parse error: Unexpected EOF` — so it parses as a stream error with **no** `error` event, and `r.error` is `undefined`:

```
expect(received).toBe(expected)
Expected: "run_in_progress"
Received: undefined
  at test/integration/chat-stream-concurrent.test.ts:186
```

Passes reliably locally; only flakes under concurrent CI load.

## Root cause

The assertion requires every rejected concurrent request to deliver a clean `run_in_progress` error event. But a request that loses the concurrency race can legitimately have its SSE stream aborted/truncated under load, delivering no clean error event. The invariant the test actually cares about is *"exactly one winner, the rest rejected, no state corruption"* — not that every rejection arrives as a well-formed `run_in_progress` event.

## Fix

Loosen the rejection assertion (lines ~185-186) to accept a truncated/aborted stream as a valid rejection alongside `run_in_progress`: for an `sseError` result with no error code (truncated stream), treat it as rejected rather than asserting a specific code. Keep the core invariants:
- `winners.length >= 1`
- at least one request rejected
- total === 5

## Notes

- Localized to a single test file; no production code change.
- Good first issue: clear repro, well-scoped, self-contained.

buyicoder:
## Summary
- Allow losing concurrent SSE streams to be treated as rejected when they are truncated before emitting an error event.
- Keep the core concurrency invariants and strict `run_in_progress` assertion for HTTP 409 responses.

## Test plan
- `bun test test/integration/chat-stream-concurrent.test.ts`

Closes #511

### DIFF
--- a/test/integration/chat-stream-concurrent.test.ts
+++ b/test/integration/chat-stream-concurrent.test.ts
@@ -145,9 +145,10 @@ describe("POST /v1/chat/stream — concurrency protection", () => {
     //   b) open the stream and get an SSE error `run_in_progress` from the
     //      runtime.chat() reject path, or
     //   c) be the single winner, emitting a `done` event.
-    // Either (a) or (b) is a valid rejection shape — the invariant is that
-    // across all responses, at most one `done` event is produced and every
-    // other response is cleanly rejected with the stable error code.
+    // Either (a) or (b) is a valid rejection shape. Under CI load, a losing SSE
+    // stream can also be truncated before emitting its error event. The
+    // invariant is that across all responses, at most one `done` event is
+    // produced and every other response is rejected.
     const results = await Promise.all(
       Array.from({ length: 5 }, (_, i) =>
         fetch(`${baseUrl}/v1/chat/stream`, {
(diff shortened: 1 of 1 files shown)


mgoldsborough:
Thanks for the PR [buyicoder](https://github.com/buyicoder). taking a look now.

 Give us a star if you can :)

mgoldsborough (approved):
LGTM!

buyicoder:
Thanks for the review and merge! I starred the repo as well.

PR #564, opened 2026-06-26
web(conversations): scope the conversation list to the focused room

mgoldsborough:
## What

The conversations app rendered **every** chat the identity owns, across all rooms, with no room scoping — the undifferentiated "where did this run?" list. The runtime already walls each chat session to one workspace; the list should match.

This scopes the conversation list to the **focused room by default**, with a room selector at the panel head that toggles to **"All rooms"** for the deliberate cross-room view.

| Layer | Change |
|---|---|
| **conversations bundle** | Surface the `workspaceId` breadcrumb (already persisted on the JSONL line-1 header) through `ConversationMeta` → `IndexEntry` → the `list` result. Filter by room **server-side** in `ConversationIndex.list()` — beside the ownership filter, **before the limit** — via new `workspaceId` + `includeUnstamped` params on the `list` tool. |
| **host** | Include `isPersonal` on the active-room `hostContext.workspace` payload, so the app folds legacy chats with no stamped room into Personal. **The active room only** — no cross-room roster crosses the iframe boundary. |
| **ui** | Default room-scoped list + room selector (`{room} ▾` / `All rooms`); `loadList` passes the focused room (and `includeUnstamped` when it's personal) and reloads on scope/room change. Count lede and date buckets follow the active scope. |

## Behavior

- **In a room:** the list shows that room's chats. Because the filter runs before the limit, a room whose chats aren't in the global most-recent page still returns its set (no under-count).
- **"All rooms":** the full cross-room list (the previous behavior), now opt-in.
- Legacy chats with no stamped room show only when the focused room is Personal.
- Live `data.changed` refreshes and title updates respect the scope.

## Scope notes

- Pure presentation + a surfaced breadcrumb + a server-side filter — **no storage migration**, no change to where conversations live or how they're authorized (still owner-scoped identity tools).
- **Pre-migration deployments:** until `migrate-personal-workspaces` runs, `WorkspaceInfo.isPersonal` is `false` for every workspace, so legacy roomless chats match no scoped room and appear only under "All rooms". Acceptable degradation given the escape hatch; not a bug.
- **Search** stays cross-room for now (search results aren't room-tagged); the selector governs the list view. Follow-up.
- **Per-item room chips** in the All-rooms view are deferred to the next slice — they need the room roster, which must be gated so it reaches only first-party identity apps (not every sandboxed iframe).

## Verification

- `bun run verify` (format + lint + tsc + cycles + codegen + unit + web 575/0 + bundles) — green. Schema change regenerated via `bun run codegen`.
- New `list room filtering` unit tests (5 cases, incl. the pre-vs-post-pagination case). Conversations bundle `vite build` + UI `tsc` + web `vite build` — green.
- Bundle `dist/` is gitignored (rebuilt by the pipeline); source only here.

### DIFF
--- a/src/bundles/conversations/src/index-cache.ts
+++ b/src/bundles/conversations/src/index-cache.ts
@@ -48,6 +48,10 @@ export interface ListOptions {
   sortBy?: "created" | "updated";
   dateFrom?: string; // ISO 8601
   dateTo?: string; // ISO 8601
+  /** Scope to one room. Applied before pagination so the limit applies to the room's set. */
+  workspaceId?: string;
+  /** With `workspaceId`, include roomless (legacy) entries — they belong to the personal room. */
+  includeUnstamped?: boolean;
 }
 
 /**
@@ -155,6 +159,18 @@ export class ConversationIndex {
     // reflects the caller's visible set, not the global one.
     if (access) {
       items = items.filter((e) => e.ownerId === access.userId);
+    }
+
+    // Room filter — scope to one workspace BEFORE pagination, so the limit
+    // applies to the room's set rather than slicing a global page and then
+    // dropping out-of-room entries (which under-counts a room whose chats
+    // aren't in the global most-recent page). A roomless (legacy) entry
+    // belongs to the personal room, so it's included only when the caller
+    // asks for it via `includeUnstamped` (set when the focused room is personal).
+    if (options?.workspaceId) {
+      const wsId = options.workspaceId;
+      const includeUnstamped = options.includeUnstamped ?? false;
+      items = items.filter((e) => (e.workspaceId ? e.workspaceId === wsId : includeUnstamped));
     }
 
     // Search filter: case-insensitive substring on title + preview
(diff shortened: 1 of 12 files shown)


mgoldsborough:
Thanks — addressed in `7c53519f`.

**Critical #1 (post-pagination filtering): fixed.** Moved the room filter server-side into `ConversationIndex.list()`, beside the ownership filter and **before** the limit/slice, via new `workspaceId` + `includeUnstamped` params on the `list` tool (input schema + generated declarations regenerated via `bun run codegen`). The client-side `roomScoped` filter is gone; the UI passes the focused room (and `includeUnstamped` when it's personal) and reloads when the scope/room changes. "All rooms" omits the params. The limit now applies to the room's set, so a room whose chats aren't in the global most-recent page no longer comes back short/empty, and the count is accurate.

**Suggestion 2 (coverage): added.** New `list room filtering` unit tests in `index-cache.test.ts` (5 cases): room scoping, `includeUnstamped` folding roomless chats into Personal, a non-personal room *not* picking up roomless chats, no-filter returns all, and a **pre-vs-post-pagination** case (25 newer chats in one room + 1 older chat in another; asserts the older room's chat is absent from the global page-of-20 but present when room-scoped) — the case that would have caught the original bug.

**Suggestion 1 (pre-migration `isPersonal: false`): noted** in the PR body — acceptable degradation given the "All rooms" escape hatch.

PR #565, opened 2026-06-26
web(nav): room-centric sidebar — the room you're in is the parent

mgoldsborough:
## What

Restructures the sidebar so the **room you're in is the parent of its contents** — the §5.1 room-centric IA. The sidebar previously listed Conversations / Automations / Files as global top-level nav with workspaces as a sibling category below; but the runtime walls each session to one workspace and those views are room-scoped now (see the conversation-list room-scoping that just landed), so the old nav contradicted the model.

```
Focused on Personal (home):        Focused on Helix:
┌────────────────────────┐         ┌────────────────────────┐
│ ▼ Personal             │         │ ⌂ Personal             │ ← way home
│    💬 Conversations     │         ├────────────────────────┤
│    ⏰ Automations       │         │ ▼ Helix                │ ← promoted
│    📁 Files             │         │    💬 Conversations     │
│    🔌 Connectors        │         │    ⏰ Automations       │
├────────────────────────┤         │    📁 Files             │
│ WORKSPACES          +  │         │    🔌 Connectors        │
│   M Helix              │         │    👥 People            │
│   A Acme               │         │    ✓ Tasks              │
└────────────────────────┘         ├────────────────────────┤
                                   │ WORKSPACES          +  │
                                   │   A Acme               │
                                   └────────────────────────┘
```

## Changes

- **New `CurrentRoomNav`** — the focused room is a section at the top: its Conversations / Automations / Files / Connectors + its apps (capped at `MAX_INLINE_APPS` with a View-all overflow) nest under it. **Home = Personal** (the standalone Home item is dropped). When focused on a **shared** room, a compact **Personal way-home row** sits above it (click → `/`).
- **`WorkspaceSection` is now the OTHER rooms** — Personal (home) and the focused room (promoted above) are filtered out; clicking a row focuses it (it promotes into the current-room section). The inline-apps-under-the-focused-row logic moved into `CurrentRoomNav`.
- **`ShellLayout`** drops the top-level identity-app `NavItem`s (subsumed by `CurrentRoomNav`) and the unused threshold/`NavItem` machinery.

## Decisions / scope

- **Routing unchanged for v1:** identity views keep their top-level routes (`/conversations`, …), scoped by the focused room. Re-homing them under `/w/:slug` is deferred to the storage-relocation track. Connectors → `/w/<slug>/settings/connectors`.
[...]

PR #568, opened 2026-06-26
web(files): scope the file list to the active workspace

mgoldsborough:
## What

The files app listed **every** file the identity owns, across all rooms. The runtime walls each session to one workspace, so the list should match — the same room-scoping the conversation list got in #564.

| Layer | Change |
|---|---|
| **server** | Filter the `list` tool by room **before** the limit, via `workspaceId` + `includeUnstamped` (`FileEntry` already carries the `workspaceId` breadcrumb). Legacy roomless files belong to the personal room. |
| **ui** | Default room-scoped list + a room selector in the header (`{room}` / "All rooms"); `loadFiles` passes the focused room and reloads on scope change. The host already ships `isPersonal` on the active-room hostContext (from #564), so no host change. |

## Behavior

- **In a room:** the grid shows that room's files. Legacy files with no stamped room show only when the focused room is Personal.
- **"All rooms":** the full cross-room set, opt-in.
- Filtering runs before the limit, so a room whose files aren't in the global most-recent page isn't under-counted (the lesson from #564's QA).

## Scope notes

- Presentation + a server-side filter only — **no storage migration**, files stay identity-owned.
- **Search** stays cross-room for now (results aren't room-tagged); the selector governs the list. Follow-up.
- Per-item room chips deferred (same as #564) — they need the roster, gated to first-party identity apps.

## Verification

- `bun run verify` green: format, lint, tsc, **codegen**, cycles, unit, web, bundles. Schema change regenerated via `bun run codegen`.
- New `files room filtering` tests in `source.test.ts` (room scope, `includeUnstamped` → personal, all-rooms, and the **filter-before-limit** case). Files bundle build + web build green.

Part of the room-scoping correctness track (conversations #564 → files → automations next). Independent of the parked sidebar redesign (#565).

### DIFF
--- a/src/bundles/files/ui/src/Dashboard.tsx
+++ b/src/bundles/files/ui/src/Dashboard.tsx
@@ -4,28 +4,25 @@ import { DetailOverlay } from "./DetailOverlay";
 import { FileGrid } from "./FileGrid";
 import { collectTags, TYPE_FILTERS } from "./format";
 import { Header } from "./Header";
-import type { FileEntry, FilterKey, ListResult, RoomScope } from "./types";
+import type { FileEntry, FilterKey, ListResult } from "./types";
 
 const SEARCH_DEBOUNCE_MS = 300;
 
 export function Dashboard() {
   const synapse = useSynapse();
   const { pickFiles } = useFileUpload();
-  // The room the shell is focused on — the binding for the default room-scoped
-  // list. Pushed by the host via hostContext.
+  // The workspace the shell is focused on — the list scopes to it. Pushed by
+  // the host via hostContext.
   const { workspace } = useHostContext<{
     workspace?: { id: string; name: string; isPersonal?: boolean };
   }>();
   // Primitives (not the workspace object, whose identity churns per push) so
-  // `loadFiles` only re-runs when the room actually changes.
-  const roomId = workspace?.id;
-  const roomIsPersonal = workspace?.isPersonal === true;
+  // `loadFiles` only re-runs when the workspace actually changes.
+  const workspaceId = workspace?.id;
+  const workspaceIsPersonal = workspace?.isPersonal === true;
 
   const [files, setFiles] = useState<FileEntry[]>([]);
   const [totalCount, setTotalCount] = useState(0);
-  // Default to the focused room — the list matches where you are. "All rooms"
-  // is the deliberate cross-room escape hatch.
-  const [roomScope, setRoomScope] = useState<RoomScope>("current");
   const [activeFilter, setActiveFilter] = useState<FilterKey>("all");
   const [activeTag, setActiveTag] = useState<string | null>(null);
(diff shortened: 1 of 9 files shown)


mgoldsborough:
Addressed Critical #1 in the latest commit. The UI already read `result.data.totalCount` and `types.ts` already declared `totalCount: number` — only the *server* drifted to `total`, so I aligned the `list`/`search` handlers to return `totalCount` (matching the conversation list's convention). No UI change needed; the header count now renders the room-scoped total. No schema/codegen impact (the list output is informal JSON via `ok()`). Tests updated; `bun run verify` green.

Both suggestions left as-is, as flagged: search staying cross-room is a conscious scope cut (results aren't room-tagged — follow-up), and the pre-handshake flash is identical to the conversations bundle (#564), so it's consistent rather than new.

mgoldsborough:
Addressed suggestion 1: added a search test that asserts `handleSearch` returns the renamed `totalCount` field **and** pins the current cross-room behavior (a `report` search surfaces matches from both Helix and Acme). So when search does get room-scoped, it changes against an explicit test rather than silently. Suggestion 2 (the search seam) is a product call — see the thread; recommending we keep search cross-room for now to stay consistent with the conversation app and room-scope search across all three apps as one follow-up.

mgoldsborough:
Reworked per feedback — simpler, and no new vocabulary:

- **Dropped the scope toggle entirely.** Files are identity-owned (stored per-user, across workspaces), so the `list` tool returns everything by default. "Show this workspace's files" is just the active workspace — there's no second concept to choose. Removed the `RoomScope` state, the selector, and the "All workspaces" view; the UI always scopes the list to the active workspace, and the server filter does the work. Net **−59 lines**.
- **Dropped "room" for "workspace"** throughout (`filterByWorkspace`, schema/param descriptions, tests). "Room" was a stray framing from the design doc; the codebase already says "workspace" everywhere, so this just stops inventing a second word for an existing concept.

`bun run verify` green (incl. codegen + 14 files tests). Search still spans workspaces (unchanged); browse is now always the workspace you're in.

Note: the same two cleanups apply to the already-merged conversation list (#564 has the "All rooms" toggle + `RoomScope` on main) and the parked sidebar draft (#565 has `CurrentRoomNav`/"room" language) — I'll do those as follow-ups.

mgoldsborough:
Closing — wrong layer. Investigating the QA feedback surfaced that `FileEntry.workspaceId` is never stamped on any creation path (neither the `create` tool nor ingest), so this PR's UI filter had nothing real to filter on. More fundamentally, files are identity-owned by design (#299), which is the actual leak: a file uploaded in one workspace is reachable from another.

The fix is the storage relocation already specified in `research/SPEC-permission-boundaries.md` (§2.3 "the room owns the directory", step 5) — files move under `workspaces/<wsId>/files/<ownerId>/` so the workspace boundary is structural, not a list filter. Superseded by that work; reopening this branch isn't the path.

PR #569, opened 2026-06-29
feat(files): workspace-owned storage — files live under the workspace

mgoldsborough:
## What

Files move from identity-owned `users/<userId>/files/` to **workspace-owned** `workspaces/<wsId>/files/<ownerId>/` — the same "path is the wall" layout #557 gave conversations. The workspace becomes a real boundary for files, not a soft breadcrumb.

This fixes the leak: a file uploaded in one workspace (say, financial) was reachable from another (ops), because files were identity-owned and shared across all your workspaces. Now **a file created in one workspace is structurally unreachable from another.**

Implements the files half of `research/SPEC-permission-boundaries.md` §6 step 5.

## The model (per spec §2.3)

`(owner, workspace, visibility)` on every file:
- **`workspace` + `owner` are path-authoritative** — `workspaces/<wsId>/files/<ownerId>/`. The directory is the truth; the store backfills the fields from the path on read. Under app-level auth (no RLS), the boundary lives in the path where a query can't reach across it.
- **`visibility` defaults to `private`** (fail-closed; absent ⇒ private). **v1 is private-only** — `shared` is dormant groundwork, no read path consults it yet.

## Changes

| Area | Change |
|---|---|
| `src/files/paths.ts` (new) | `workspaceFilesDir()` + `parseFilePath()` — the single sanctioned path site (mirrors `conversation/paths.ts`). |
| `FileEntry` | `ownerId` + `workspaceId` required, `visibility?` added. Store backfills owner/workspace from the dir on read. |
| `getFileStore(wsId, ownerId)` | Replaces `getFileStore(userId)`. Callers pass the conversation's workspace (`chat`/`runTurn`) or the **membership-validated** request workspace (REST), falling back to the owner's personal workspace. |
| Creation | `handleCreate` + `ingest` stamp `(owner, workspace, visibility:private)`. |
| `check:file-paths` | Inverted — file dirs must be built via `workspaceFilesDir`; identity-owned `users/<userId>/files/` construction is forbidden. |
| `scripts/migrate-files-to-room.ts` (new) | Existing files → owner's personal workspace (no provenance → personal). Idempotent, dry-run default, under the migration lock. |

## Security model

- **Workspace wall:** REST file serving runs behind `optionalWorkspace` (validates `X-Workspace-Id` → membership; 403 for non-members). No header ⇒ personal workspace (fail-closed).
- **Owner-private:** the `<ownerId>/` partition means you only ever read your *own* files in a workspace you belong to. Another member's files are a different partition — invisible by construction, not by check.

## Verification

- `bun run verify` green — format, lint, tsc, cycles, **check:file-paths** (inverted guard), codegen, unit + web + bundle + integration (**591 pass / 0 fail**).
- New integration tests: **cross-workspace isolation** (a file in WS A is null/absent from WS B's store) + **migration round-trip**. Plus store-backfill unit coverage.

## Migration / rollout note

Like #557 for conversations, the frontend must send `X-Workspace-Id` consistent with where a file was created (the upload path and the read path now agree on the partition). Run `bun run migrate:files-to-room --write` to relocate existing files (→ each owner's personal workspace).

Automations relocation (the other half of step 5) follows as a separate PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,8 @@
     "format:check": "bunx @biomejs/biome format src/ web/src/ instrument/ scripts/",
     "sync-models": "bun run src/model/sync-models.ts",
     "migrate:skill-frontmatter": "bun run scripts/migrate-skill-frontmatter.ts",
-    "migrate:conversations-to-room": "bun run scripts/migrate-conversations-to-room.ts"
+    "migrate:conversations-to-room": "bun run scripts/migrate-conversations-to-room.ts",
+    "migrate:files-to-room": "bun run scripts/migrate-files-to-room.ts"
   },
   "devDependencies": {
     "@biomejs/biome": "^2.4.10",
--- a/scripts/check-file-paths.ts
+++ b/scripts/check-file-paths.ts
@@ -1,27 +1,39 @@
 #!/usr/bin/env bun
 /**
- * Lint: files are identity-owned and reached through one constructor.
+ * Lint: files are workspace-owned — never built at the identity level.
  *
- * Phase B moved files from the per-workspace store
- * (`{workDir}/workspaces/{wsId}/files/`) to the identity store
- * (`{workDir}/users/{userId}/files/`). To keep that single, two things are
- * enforced in `src/`:
+ * A file lives under the workspace it was created in, with the owner as a
+ * privacy sub-partition:
  *
- *   1. `createFileStore(...)` is called only in `src/runtime/runtime.ts` —
- *      via `Runtime.getFileStore(userId)` (the sanctioned identity-scoped
- *      constructor) and the host-resources resolver closure. Any other call
- *      site builds a store off a path the caller chose, which is how files
- *      drifted back into a workspace silo.
- *   2. `join(getWorkspaceScopedDir(...), ..., "files")` — building a
- *      workspace-scoped files dir by hand — is forbidden anywhere in `src/`.
+ *   workspaces/<wsId>/files/<ownerId>/   a member's files (private by default)
  *
- * Allowed: a `// lint-ok:file-path` marker on a line just above the call,
- * for the rare future case the constructor genuinely can't cover.
+ * The identity-owned `{workDir}/users/<userId>/files/` layout is gone. Any new
+ * occurrence of identity-level file construction — `getDataPath("files")` on an
+ * `IdentityContext`, or joining `"files"` directly onto a non-workspace root —
+ * is a regression: it drops the workspace wall the permission model depends on
+ * (§2.3).
  *
- * Scope: `src/**\/*.ts`. Tests and `scripts/` are out of scope (the
- * migration deliberately reads the old workspace-scoped layout).
+ * What this script flags:
+ *   - `x.getDataPath("files")` / `getDataPath("files")` — the identity store
+ *     constructor. Files are no longer identity-owned.
+ *   - `join(...)` whose `"files"` segment is NOT qualified by a workspace root
+ *     before it (neither a literal `"workspaces", <wsId>` pair nor a
+ *     workspace-scoped base like `this.workspacesRoot` / `getWorkspaceScopedDir()`).
  *
- * Exports its AST predicates for the self-test under `test/unit/scripts/`.
+ * What it allows:
+ *   - `src/files/paths.ts` — the single sanctioned construction (and parse)
+ *     site for the `workspaces/<wsId>/files/<ownerId>` layout. It *is* the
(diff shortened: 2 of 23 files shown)


mgoldsborough:
Both criticals were real — fixed in `f731d9bb`, each with a regression test that **fails if the fix is reverted** (the existing suite missed both):

**Critical #1 (file partition followed the request, not the conversation's room): fixed.** `chat()` now captures the authoritative `roomWsId` that `resolveChatStore` recovers from the conversation's path and builds the file store from it — so on a cross-room resume, attachments rehydrate from the conversation's actual room, not the request header. New `cross-room-file-resume.test.ts` resumes a conversation *unfocused* (request room = personal ≠ the conversation's room) and asserts every rehydration partition is the conversation's room (reverting the fix → it reads personal → file lost → test fails). `runTurn` was already internally consistent (file store + conversation store both use `provRoomWsId`), so it's untouched.

**Critical #2 (migration dropped disjoint registries): fixed.** `registry.jsonl` is now **merged** (source rows appended to an existing dest registry; `readRegistry` dedupes by id) instead of move-or-drop. New test seeds a pre-existing dest registry (the post-deploy-upload window) and asserts both the old and new entries survive (reverting → the old entry is orphaned → test fails). The "crash recovery" test was retargeted to a content-addressed blob, which is still the valid move-or-drop path.

**Suggestion 1 (withScope partial entries on scratch paths): documented** (a comment; no throw, since test/scratch stores rely on the lenient pass-through). **Suggestion 2 (lint substring heuristic): acknowledged**, left as-is.

`bun run verify` green (593 pass / 0 fail).

---

**One deferral I want your call on.** The downstream of #1 — the bundle resolver and `handleFileServe` reading a bare file id without a conversation — is *not* fixed here. The correct fix is a **membership-aware `FileLocator`** (resolve id → workspace, then verify the caller is a member, so it can't read files from a workspace they've *left*) — symmetric with the `ConversationLocator` #557 added. That's security-sensitive and I'd rather design it deliberately than bolt it onto this PR. So this PR ships the data-loss fixes; bare-id file resolution stays header-dependent (the web app sends `X-Workspace-Id`, so display works; header-less direct links / cross-room bundle reads are the gap). I'll open it as a scoped fast-follow — unless you'd prefer to hold #569 until the locator lands.

mgoldsborough:
Both criticals fixed in `5ff38398`. Verify green (594 pass / 0 fail).

**Critical #1 (upload partitioned by header, read by the conversation's room → silent loss): fixed.** Last round I made the read authoritative but left the write header-derived — the asymmetry you caught. Both upload handlers (`handlers.ts` chat-multipart + resource-upload) now resolve the conversation's authoritative room via a new `Runtime.resolveConversationRoomWsId` (probe the focused/personal room, then the locator; a new conversation falls back to the focused/personal workspace — mirroring `resolveChatStore`) and write into that partition. Write and read now agree on the conversation's room.
- **New e2e test through the real handler** (`cross-room-file-upload.test.ts`): uploads via `POST /v1/resources` on an *unfocused* resume and asserts the file lands in the conversation's room (`workspaceId === ROOM_A`), not the personal/header partition. You were right that no test drove the handler — this one does, and it **fails if the write-side fix is reverted** (the file lands in `ws_user_…`, the exact vanish).

**Critical #2 (NUL byte in `runtime.ts`): fixed.** The `getFileStore` cache-key separator was a literal NUL — replaced with a normal delimiter. `runtime.ts` is text again (ripgrep reads it; that's how I'd been hitting "binary file matches" and working around it with `grep -a` instead of investigating — my miss).

**Suggestion 1 (migration summary): fixed** — `merged` is in the totals and merge plans get their own "merged into …'s registry" line (not the "removed stale copy" message). **Suggestion 2 (handleCreate store+scope redundancy): acknowledged**, left as-is.

The one still-open item is the same as before — the **membership-aware `FileLocator`** for bare-id reads with *no* conversation (`handleFileServe` direct links / bundle reads). The chat-attachment flow is now fully correct on both write and read; the locator only matters for header-less direct serving, and I'm keeping it as a deliberate fast-follow unless you'd rather hold for it.

mgoldsborough:
Closing as a duplicate. #567 ("Files: workspace-owned storage — the path is the wall") landed the same work on main while this branch was in flight — this was branched off #566 before #567 merged and not reconciled. The implementations are independent, so this isn't mergeable on top of #567.

The QA findings on this PR are real bug *shapes*, so they're being audited directly against the merged #567 code on main; genuinely-live ones will be fixed there. The branch stays available for reference.

PR #574, opened 2026-06-29
fix(files): conversation files follow the conversation's workspace, not the request

mgoldsborough:
## What

#567 made files workspace-owned (`workspaces/<wsId>/files/<ownerId>/`), but several read/write paths still resolved a file's workspace from the **request header** instead of the **conversation's authoritative workspace**. A file attached in workspace A silently vanishes when the conversation is resumed focused elsewhere (or downloaded without the exact `?ws=`).

These were surfaced by the QA on the (now-closed, duplicate) #569 and **confirmed live on main via a direct audit of #567** — this is the focused fix.

## The fix (all four are partition-resolution, same root cause)

| Path | Was | Now |
|---|---|---|
| **Read** (`chat()`/runTurn rehydration) | file store built from `request.workspaceId ?? sessionWsId` (the returned `convWsId` was discarded) | built from the conversation's authoritative workspace (`convWsId`/`provWsId` from `resolveChatStore`) |
| **Write** (multipart + resource upload) | partitioned by the request header | resolves the conversation's workspace (new `Runtime.resolveConversationWorkspaceId`, mirroring `resolveChatStore`) and writes there |
| **Serve** (`GET /v1/files/:fileId`) | `?ws=` only → cross-workspace download 404s | accepts optional `conversationId`, follows the conversation's workspace |
| **Resolver test** | fixture closure ignored the `wsId` param (a regression dropping `ctx.workspaceId` stayed green); comments said "identity-owned" | honors `wsId`, asserts cross-workspace isolation, comments corrected. **Resolver behavior unchanged** |

## Security notes

- **Serve:** the file store is always rooted at the **caller's own `<ownerId>` partition**, so a forged `?ws=` or `conversationId` can only ever reach the *caller's own* files — not a cross-user boundary, so no membership check is load-bearing there (documented in-code).
- **Resolver:** confining a bundle's `files://` reads to its own workspace (`ctx.workspaceId`) is the **intended trust boundary** — an untrusted bundle must not follow the conversation across the wall. This PR keeps that behavior and only fixes the test that wasn't covering it.

## Verification

- `bun run verify` green: **590 pass / 0 fail**.
- Three new tests (cross-workspace resume rehydration, cross-workspace upload through the **real** `POST /v1/resources`, and file-serve-by-conversation through the **real** `GET /v1/files`) — **each fails if its fix is reverted** (revert-checked).
- Migration (`migrate-files-to-workspace`) was audited and is **not affected** — its `collapseRegistry` already merges disjoint source/dest entries safely.

## Frontend follow-up (not in this PR)

`FileAttachment.tsx` builds the download URL with `?ws=<activeWorkspaceId>`, not the conversation's workspace, and doesn't receive `conversationId`. The runtime now **accepts** `?conversationId=`; a small web change should **send** it (thread `conversationId` into `FileAttachment` + `fileUrl(fileId, { conversationId })`) so cross-workspace downloads resolve correctly.

### DIFF
--- a/src/api/handlers.ts
+++ b/src/api/handlers.ts
@@ -1701,16 +1701,21 @@ export function sanitizeFilename(name: string): string {
   return name.replace(/["\r\n\x00-\x1f]/g, "_");
 }
 
-/** Handle GET /v1/files/:fileId?ws=<wsId> — serve a stored file from the
- * workspace it lives in. Files are workspace-owned; the URL must carry the
- * workspace (`?ws=`) because a browser GET can't send the `X-Workspace-Id`
- * header. */
+/** Handle GET /v1/files/:fileId?ws=<wsId>&conversationId=<id> — serve a stored
+ * file from the workspace it lives in. Files are workspace-owned; the URL must
+ * carry the workspace (`?ws=`) because a browser GET can't send the
+ * `X-Workspace-Id` header. When `conversationId` is present the workspace is
+ * resolved from the conversation instead — the download then follows the
+ * conversation's workspace (the partition the chat read/write path uses), so a
+ * file stored under a cross-workspace conversation resolves without the client
+ * knowing the right `?ws=`. */
 export async function handleFileServe(
   fileId: string,
   runtime: Runtime,
   features: ResolvedFeatures,
   identity: UserIdentity | undefined,
   workspaceId: string | undefined,
+  conversationId: string | undefined,
 ): Promise<Response> {
   if (!features.fileContext) {
     return apiError(404, "not_found", "Not found");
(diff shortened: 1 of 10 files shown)


mgoldsborough:
Both criticals fixed in `d121e9d2`, plus Suggestion 1 — and writing the test you asked for surfaced a *fourth* path, which is exactly why it was worth insisting on.

**Critical #1 + #2 (fixed):** `RequestContext.fileWorkspaceId` — which scopes the agent's own identity-door `files__*` tools — was the third path still using the request header. It now follows the conversation's authoritative workspace (`convWsId` in chat, `provWsId` in runTurn).

**The deeper find (from the tool-invoking test):** the runtime change alone was *dead code* — `IdentityToolRouter.execute` rebuilds the per-call `RequestContext` before dispatching and was **dropping `fileWorkspaceId`**, so `files__list` got "no workspace in scope" even with the fix in place. The restamp now propagates it (orthogonal to `scope`, same shape as `conversationId`/`toolPromotion`). The new `cross-workspace-file-tool-resume.test.ts` drives a real `files__list` call on an unfocused resume and asserts it returns the attachment from workspace A — and **fails if the `fileWorkspaceId` fix is reverted** (returns `{ files: [], total: 0 }` from the personal partition).

**Suggestion 1 (done) — the root cause:** the probe-then-locate partition rule was copy-pasted across `resolveChatStore`, `resolveConversationWorkspaceId`, and the inline `fileWorkspaceId` — which is *why* each review round found another stray site. Consolidated into one private `resolveConversationLocation` all three flow through, so it physically can't drift again.

`bun run verify` green (591 pass / 0 fail). The frontend follow-up (thread `conversationId` into `FileAttachment` so downloads send it) is unchanged — still out of this PR.

mgoldsborough:
Clean review — thank you for the thorough survey (confirming every conversation-bearing path routes through the one `resolveConversationLocation` chokepoint, and that `handleToolCall` is correctly header-scoped, was exactly the assurance I wanted). Both optional polish items applied in `ecf5702f`: dropped the upload `"pending"` sentinel round-trip (`convId` is now `string | undefined` straight into the resolver; `ingestFiles` gets `?? "pending"`), and tightened the runTurn rehydration comment so it no longer frames `provWsId` as independently conversation-derived. No behavior change; `bun run verify` green (591 pass).

PR #577, opened 2026-06-29
web(files): send conversationId on file download so cross-workspace attachments resolve

mgoldsborough:
## What

The frontend completion of #574. That PR made the runtime resolve a file's workspace from an optional `?conversationId=` on `GET /v1/files/:fileId` (so a download follows the conversation's workspace, not the focused one). But the web client only ever sent `?ws=<activeWorkspaceId>` — so an image attachment in a conversation you've resumed while focused on a *different* workspace would 404.

This threads the conversation id through to the file URL:
- `fileUrl(fileId, workspaceId?, conversationId?)` adds a `conversationId` param (via `URLSearchParams`).
- `FileAttachment` takes an optional `conversationId` prop and passes it to `fileUrl` for the image thumbnail src.
- `MessageList` passes `conversationId` to each `FileAttachment`.
- `ChatPanel` supplies it from `useChatContext()`.

The server prefers `conversationId` (resolves the conversation's authoritative workspace) and falls back to `?ws=` when absent — so nothing regresses for files viewed in their own workspace.

## Verification

- `web` tsc clean, `bun run build` clean, `bun run test:web` green (575 pass).
- No behavior change when no conversation is in scope (e.g. the Files app) — `?ws=` is still sent.

Closes the only open follow-up from the files-workspace-storage work.

---

_QA polish addressed: added `fileUrl` unit coverage (ws/conversationId branches, active-workspace fallback, id encoding) — web suite now 580 pass._

### DIFF
--- a/web/src/api/client.test.ts
+++ b/web/src/api/client.test.ts
@@ -1,0 +1,39 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import { fileUrl, setActiveWorkspaceId } from "./client";
+
+describe("fileUrl", () => {
+  afterEach(() => {
+    setActiveWorkspaceId(null);
+  });
+
+  test("includes both ws and conversationId when given", () => {
+    const url = fileUrl("f1", "ws1", "conv1");
+    const query = new URLSearchParams(url.split("?")[1]);
+    expect(url).toContain("/v1/files/f1");
+    expect(query.get("ws")).toBe("ws1");
+    expect(query.get("conversationId")).toBe("conv1");
+  });
+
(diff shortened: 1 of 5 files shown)


mgoldsborough:
Superseded by #579. That PR threaded `conversationId` into the file-download URL so the server could resolve the attachment's workspace. #579 goes further on the same thread: file ids are globally unique, so a new `FileLocator` resolves the workspace from the bare id within the caller's own partitions, and the route becomes `GET /v1/files/:fileId` with **no** client-supplied coordinate at all. Closing this in favor of that approach.

PR #579, opened 2026-06-29
feat(files): serve files by bare id via a workspace locator

mgoldsborough:
## What

A browser `` GET can't send `X-Workspace-Id`, so the file-serve route needs to recover which workspace a file lives in. The shipped approach carried the **focused** workspace (`?ws=<activeWorkspaceId>`), which is wrong for a cross-workspace attachment — resume a conversation that lives in workspace A while focused on B, and its image 404s. The interim fix (#574) accepted `?conversationId=` and resolved the workspace from the conversation, but that conflates a file with a conversation it's only *referenced* by, and still threads a coordinate the client has to get right.

File ids are **globally unique** (`fl_<24 hex>` — `src/files/uri.ts` already states it), so the bare id is sufficient. This makes the route bare and resolves the workspace server-side:

```
GET /v1/files/:fileId        ← no ?ws=, no ?conversationId=
```

## How

- **`FileLocator`** (`src/files/locator.ts`), mirroring `ConversationLocator`: `locate(ownerId, fileId) → wsId` via a path-only readdir walk of the caller's own owner partitions, with a `fileId → wsId` memo.
- **The owner partition is both the gate and the search scope.** `ownerId` comes from the authenticated identity, never the URL; the store reads `(wsId, ownerId)`. There is no client-supplied coordinate to forge, and a request can only ever reach the caller's own bytes.
- **Memo, never the source of truth.** Kept current at the write sites — `getWorkspaceFileStore` wraps `saveFile` → `remember` and `deleteFile` → `forget`, so a freshly uploaded file resolves O(1). A stale hit self-heals: the serve handler drops it and re-resolves from disk once. Correct under `replicas > 1` (a miss costs a walk, not a wrong answer).
- **Owner-scoped by construction.** Reading a file *shared* by another owner (future `visibility: shared`) is a separate, visibility-checked path — never a widening of this locator to scan other owners. Pinned by a test.
- **A globally-unique id resolving in two partitions is refused, not guessed** → 404. Pinned by a test.
- Web `fileUrl(fileId)` drops its workspace/conversation args.

## Design validation

Validated against the chief-architect review before implementation. The review flagged that naively "mirroring `ConversationLocator`" is a trap — its `locate()` is *uncached* (fine for cold conversation loads, a hot-path problem for thumbnails), so this locator carries a real memo with write-time population and self-heal. The three mandatory conditions (real cache; owner-scoped contract walled from Phase-2 shared-read; clean not-found/duplicate semantics) are all implemented and tested.

## Verification

- `bun run verify` green — **596 pass, 12 skip, 0 fail**; all static path-guards (`check:file-paths`, `check:workspace-paths`, …) pass.
- New: `test/unit/files/file-locator.test.ts` (owner-scope wall, duplicate refusal, memo/forget, path-only resolution); `web/src/api/client.test.ts`.
- Rewrote `file-serve-by-conversation.test.ts` → `file-serve-by-id.test.ts` (bare-id resolves a workspace-A attachment with no coordinate); updated the serve fetches in `chat-file-visibility.test.ts`.

## Supersedes

Closes #577 — that PR threaded `conversationId` into the file URL; this removes the need for **any** client-supplied coordinate, so it's obsolete.


mgoldsborough:
Addressed the QA review in `0db24b58`:

**Critical #1 — self-heal branch untested → FIXED.** Added `self-heals a stale memo entry` to `test/integration/file-serve-by-id.test.ts`: poisons the memo with a wrong workspace (`remember(OWNER, fileId, "ws_does_not_exist")`), then asserts `GET /v1/files/:id` still returns 200 — proving the memo read fails, the entry is dropped, and the disk re-resolve recovers. This is the branch that guards availability under `replicas > 1`.

**Suggestion #1 — genuine 404 double-walks → FIXED.** Split the locator into `peek` (memo-only) + `resolve` (disk-authoritative). The handler peeks first; a pure miss now walks **once** and 404s with no retry. The disk re-resolve runs only after a *memo* hit fails to serve (the actual self-heal case), exactly as suggested.

**Suggestion #3 — memo keyed by `fileId` alone → FIXED.** Memo is now keyed by `(ownerId, fileId)`. The owner-scope wall holds in the memo too, not just on disk: one owner's lookup can neither read nor evict another's entry. Pinned by a new unit test (`memo is owner-scoped`).

**Suggestion #2 — memoize negative lookups → declined (deliberate).** The review rates it negligible (bounded to one user's own partitions), a broken `` doesn't refetch in a loop, and negative caching adds a staleness failure mode (an id valid later wouldn't resolve) for no real gain. Left as-is.

`bun run verify` green — 597 pass, 12 skip, 0 fail.

mgoldsborough:
Second QA round is clean — no Critical, no Delete Instead, 8/8 claims verified. Both suggestions are deferred by agreement (the review itself rates them "fine to leave"):

- **Memo fast-path skips the cold walk's duplicate-refusal** — benign by construction: a memo entry only ever comes from `remember` (write path, stamped with the real wsId) or from `resolve`'s walk, and the walk *refuses* a duplicate (returns undefined, doesn't memoize). So the memo can't hold a dup-bypassing value today. Revisit only if Phase-2 `visibility: shared` changes the one-id-one-partition invariant.
- **Unbounded memo growth** — mirrors the existing `ConversationLocator` (same unbounded Map + readdirSync walk); a bounded-LRU pass should cover both locators together, out of scope here.

No further changes. Ready for a reviewer.

PR #584, opened 2026-06-29
fix(runtime): seal a resumed conversation to its own workspace

mgoldsborough:
## The leak

A conversation is born in one workspace and its on-disk path is authoritative for life — but only **files** followed that binding (#574). In `_chatInner`, tool / skill / app / `## Workspace` context resolution keyed off the client's currently-focused workspace (`request.workspaceId`), not the conversation's.

So resuming a chat born in workspace **A** while focused on **B** answered the turn with B's tools, B's skills, B's house rules, and told the model *"you're in B"* — while the thread kept showing A's history. A single conversation could accumulate turns each answered in a **different workspace's context**. Observed in staging: switching workspaces mid-conversation and asking "what workspace am I in?" returned the *new* workspace.

This reverses a documented-but-wrong decision (the old "tool scope is a separate axis from the conversation workspace" comments). It's a UX/IA *and* isolation fix: the conversation is the unit of workspace binding, so everything it resolves should be too.

## The fix

`toolsWsId` and every workspace-scoped surface (skill pool, apps, overlays, workspace context, layer-3, connector skills) now key off the conversation's authoritative `convWsId` (read from its path by `resolveChatStore`), never the request header.

- For a **new** conversation `convWsId == request.workspaceId ?? sessionWsId`, so behavior is unchanged. They diverge **only** on a cross-workspace resume — exactly the leak.
- The wall stays internally consistent: tool namespacing and the routing reqCtx both flow from `toolsWsId`, so re-pointing that one variable moves the whole session coherently.
- A `narratedWsId` predicate (`convWsId !== sessionWsId`) keeps the personal workspace the silent session bridge — a personal-workspace chat still omits the `## Workspace` block.
- **Pin, not reject:** a cross-workspace resume resolves silently in the conversation's workspace, mirroring how files already behave. The boundary is derived from server-authoritative state, never a client-supplied header — the first-principles secure choice.
- `executeTask` is **untouched**: a task never resumes, so its focused workspace is always its provenance workspace and can't diverge.

## Tests

`test/integration/runtime/cross-workspace-context-resume.test.ts` — resume a workspace-A conversation (a) unfocused and (b) focused on workspace B, and assert the `## Workspace` block the model receives names **A**, not the focused workspace. On revert, the captured prompt literally shows `ID: ws_workspace_b` for the A-conversation (the reported bug) and the test fails.

`bun run verify` green (599 integration + 576 unit pass).

## Scope

Backend data-layer seal only. A companion web PR will re-scope the floating chat panel to clear on workspace switch — pure UX clarity now that resolution is server-authoritative.


mgoldsborough:
Knowingly out of scope here: `workspace.models` / `workspace.agents` overrides still resolve from the personal workspace (`sessionWorkspace`), not the conversation's. Not an isolation concern (own config, not cross-workspace data) — tracked as the Stage-6 per-conversation pin in #585.

mgoldsborough:
## QA adjudication — both suggestions upheld and addressed

**Suggestion 2 (tool-scope asserted only transitively) — fixed.** `cross-workspace-context-resume.test.ts` now captures `opts.tools` and asserts the workspace-namespaced surface directly: the resumed A-conversation's tools are `ws_workspace_a-…`, and the focused/personal workspace's tools (`ws_workspace_b-…` / `ws_user_…-…`) are absent. No longer relies on the narration block as the canary. (f47e5eb)

**Suggestion 1 (verify the offboarding claim) — the reviewer was right; the claim was hollow.** Investigated: no offboarding/severing path exists (`grep offboard` → 0 hits). `removeMember` (`workspace-store.ts:622-647`) mutates only the members list; resume is owner-gated only (`runtime.ts:1193`); files (#574) and automations have the identical gap. So a member removed from workspace A keeps indefinite resume access to conversations/files/automations they own in A — and #584 widens that for conversations specifically (before: a removed member resuming got the *focused/personal* tools; after: A's tools).

Actions:
- Corrected the AGENTS.md note — it claimed "severing is offboarding's job," which described a control that doesn't exist. Now states it as a **known, unclosed gap**.
- Filed **#586** with the real fix (membership re-check on resume, applied across conversations/files/automations) and the deny-vs-read-only product decision.

This is **not a blocker for #584**: the gap is pre-existing and sibling-consistent, and #584's primary effect — stopping context bleed between workspaces the user *is* a member of (the reported bug) — is a net security win. But it's now honestly documented and tracked rather than papered over.

mgoldsborough:
## Correction to my QA adjudication above

I was wrong that the "removed member retains resume access" was an *unbacked* gap. It's a **deliberate, E2E-tested Stage-1 decision** — `test/integration/stage-1-cross-workspace-conversation.test.ts`: *"a conversation created in workspace A survives Alice being removed from A."* Conversations outlive their workspace context by design; membership is enforced on the **focused** workspace at the HTTP boundary (non-member chat → 403), and the conversation stays owner-gated and continuable.

I had added a membership-on-resume gate to "close" it — which silently **reversed** that documented promise and broke its test (and a same-owner resume test). I've **reverted** the gate (c0b287d). #584 is back to the clean resolution seal, full `verify` green, all existing tests passing — including the Stage-1 "outlive" test.

What #584 actually changes about offboarding is narrower and real: resume now resolves the **conversation's** workspace tools instead of the focused workspace's, so a left-the-workspace member gets A's tools rather than the focused one's. Whether to revisit the "outlive" tradeoff for the tool dimension is a **product decision**, reframed in #586 — not something this PR should decide.

Net: #584 is unchanged in intent (fix the cross-workspace context leak) and now also carries the **direct tool-surface assertions** from QA suggestion #2. The only thing that moved is that I correctly stopped short of reversing a documented design.

PR #587, opened 2026-06-29
fix(web): re-scope the chat panel to the focused workspace

mgoldsborough:
## The bug

The chat panel is a persistent floating assistant. Switching the focused workspace (`/w/:slug`) left the open conversation on screen even though it belongs to the workspace you navigated away from. The conversation *list* already re-scopes on switch; the open panel did not.

## The change

The panel is now **workspace-scoped**. A conversation lives in exactly one workspace, so on a real workspace→workspace switch the open conversation clears and the panel resets to a fresh, empty draft in the newly-focused workspace. The panel stays open — the assistant is still there; it just no longer shows a conversation from the workspace you left.

- `ChatProvider` (`web/src/context/ChatContext.tsx`) watches the route-derived focus workspace and, on a workspace→workspace transition, calls the **narrow** `newConversation()` (a fresh draft slice) — **not** `chatStore.reset()`, which is the identity-change broad reset (nukes every cached slice). Other workspaces' slices stay cached.
- Triggered off the focus **transition**, not by comparing the open conversation's workspace (which would race with opening one). `null` focus (home/identity routes) is *held*, not reset, so `A→home→A` keeps context while `A→B` (and `A→home→B`) re-scopes.
- Opening a conversation from within its own workspace doesn't change the focus, so it is **not** cleared.

## Notes for the reviewer

- **The prompt's premise didn't hold against current main.** It assumed the panel mounts only under `/w/:slug` with no panel on home. Per `ChatChrome`'s invariant it's mounted **globally** by `ShellLayout` (deliberate, identical on home/overview/app views). So I re-scope on *switch* rather than relying on unmount, and did **not** reverse the global-mount design. The "no panel on home" sub-goal is a separate UX decision, deliberately out of scope.
- **This is UX/clarity, not safety.** The backend seal (#584) already resolves a resumed conversation in its own workspace server-side, so a stale cross-workspace panel resolved correctly even before this. This makes the UI match.
- `useChat`'s `_focusWorkspaceId` param stays unused — the focus is consumed in `ChatProvider` for the re-scope; the per-request workspace header still comes from the global active workspace. (The header-override follow-up noted in `useChat` is unrelated.)

## Tests

`web/src/__tests__/chat-panel-workspace-rescope.test.tsx`: switching workspaces clears the open conversation; a same-workspace re-render does **not**; returning to the original workspace starts fresh (no resurrection). Each switch case fails if the effect is reverted (verified). `bun run verify` green.

Companion to #584 (the backend seal).


mgoldsborough:
## QA adjudication

**Critical Issue 1 (untested null-focus held branch) — fixed.** Added two direct tests (3ae94b9): `A→home→A` keeps the conversation (null held, ref not advanced) and `A→home→B` re-scopes. Confirmed both **fail** if the `focusWorkspaceId === null` early return is dropped — exactly the "tidy" refactor flagged — and the other three pass. 5 tests, `bun run verify` green.

**Suggestion 1 (participantMap) — agreed it's debt, but the diagnosis was off; filed #588 with the corrected one.** `manage_users action:list` returns `userStore.list()` (the whole **org** directory), not workspace members, and is **org-admin-only** (`user-tools.ts:46`). So:
- It does **not** go stale per workspace — the map is workspace-independent, so #587 doesn't actually violate the workspace-scoped invariant here. That specific concern is a non-issue.
- The real debt: a **non-admin** member's fetch fails → empty map → every other participant shows a raw `userId`. Attribution silently only works for admins. Root cause: messages carry `userId` but not the author's name, so the client fetches the directory to compensate.
- Clean fix (server makes messages self-describing; web drops the fetch + map) tracked in **#588**. Out of scope for this PR by design.

No change to the diff beyond the added tests.

mgoldsborough:
## QA adjudication (round 2)

No critical issues — concur, mergeable. Both suggestions were fair optional polish; folded both in (9aac84d), comment-only, no logic change:

1. **Arrive-into-workspace clears (open-from-home edge).** Added the corollary to the effect comment: triggering off the focus *transition* (not the open conversation) means arriving into a workspace always yields a fresh draft, even if that workspace's conversation was opened from a home/identity route. Documented as the accepted, race-free tradeoff (consistent with the conversation list now being workspace-scoped).
2. **Test fidelity of the null-focus proxy.** Added a comment explaining that nulling `activeWorkspace` is faithful to a non-`/w/` home route: `focusWorkspaceId = pathname.startsWith("/w/") ? activeWorkspace?.id ?? null : null`, so both produce the `focusWorkspaceId === null` the effect branches on. Kept the harness as-is (a route-navigation rework is higher cost for no added coverage, per the reviewer's "low value").

5 tests still pass; format clean.

PR #589, opened 2026-06-30
refactor(api): require a workspace on the chat send routes

mgoldsborough:
## Why

`optionalWorkspace` let a chat turn arrive with no `X-Workspace-Id`, and the runtime silently defaulted to the caller's personal workspace. In the workspace-sealed model (#584) a chat turn belongs to exactly one workspace, and the header is the load-bearing coordinate that picks the **birth workspace** of a *new* conversation — so it should be required, not defaulted.

## What

- `POST /v1/chat`, `/v1/chat/stream`, `/v1/chat/start` → `requireWorkspace`: **400** if absent, **403** if a non-member. Per-route, placed after `chatBodyLimit`/`rl` so those middlewares still apply to a header-less request. `/cancel` is unchanged (owner-gated, no workspace).
- The runtime's `?? personalWorkspaceIdFor(...)` default is now the **embedded / dev-mode** path only (dev mode passes through without an identity; in-process `Runtime.chat()`/`startTurn()` callers may omit it). Comments updated to say so.

## No user-facing change

The first-party web client already sends `X-Workspace-Id` on every chat request (it's always focused on a workspace, personal at minimum). `/v1/chat` is the trusted same-origin shell endpoint; external MCP clients use `/mcp`, not this route.

## Scoped on purpose — not a blanket sweep

Only the chat **send** routes require it, because that's the only place the header is load-bearing. The other `optionalWorkspace` routes stay optional **by design**, not debt:
- **Reads** (`conversations`, `files`, resource reads) derive the workspace server-side from the owner-scoped locator (for files, deliberately — bare `files://<id>` URIs; a browser `` GET can't send a header).
- **Identity-door tool calls** (`conversations__*`, `files__*`) are identity-scoped; the wall already requires the workspace for `ws_<id>-` tool calls at the routing layer.

## Bonus: removes a Hono wildcard leak

Dropping the chat router's `.use("*", optionalWorkspace)` also removes a wildcard that was leaking that middleware **forward** onto `/v1/events` (mounted after chat). `/v1/events` is identity-scoped (`routes/events.ts` — authorizes by identity, filters fan-out by server-computed membership) and never gates on the header; its prior non-member 403 was *purely* the leak. It now behaves as documented (200, server-side filtering). The `bootstrap-workspace-header-leak` regression test is updated to assert enforcement on `/v1/chat` (which enforces by design) rather than `/v1/events` (which never did) — same wildcard-leak class the test already guards.

## Tests

`bun run verify` green. `bootstrap-workspace-header-leak.test.ts` updated and passing (bootstrap permissive → 200; `/v1/chat` + non-member header → 403).


mgoldsborough:
## QA adjudication

**Critical Issue 1 (the `/v1/events` forward-leak guard was lost) — fixed (8547845).** You're right: replacing the `/v1/events` assertion with `/v1/chat → 403` tested `requireWorkspace`, not the chat→events wildcard-leak class, and the corrected events behavior was prose-only. Restored a `GET /v1/events` + foreign-header case that asserts the response is **NOT** a 403 (the leak's exact signature — it resolves a 403 immediately; a clean identity-scoped response opens the SSE stream).

On asserting an exact 200: in this harness the SSE headers don't flush until the first byte, so a plain `await fetch` hangs (that's why I'd dropped it). The guard now connects with an `AbortController`, aborts after 1s, and asserts only `status !== 403` — the precise regression signature, no flush race. **Verified load-bearing:** re-adding `.use("*", optionalWorkspace)` to the chat router makes this case fail with a fast 403 (1ms), exactly your failure scenario; removed again, green.

**Suggestion 1 (docstring/inline mismatch) — fixed.** Reframed the file docstring around the Hono wildcard-leak *class* with both instances (`conversationEventRoutes`→bootstrap, `chatRoutes`→events) and the three per-endpoint pins (bootstrap 200, `/v1/chat` 403-by-design, `/v1/events` not-403), instead of the contradictory dual attribution.

`bun run verify` green (600 integration pass, including the 3-case leak test).

mgoldsborough:
Suggestion 1 (absent-header 400 not asserted end-to-end) — added (14f6dfa5). One-line `POST /v1/chat` with no header → 400 (`workspace_error`), pinning the headline 'require it' behavior through the real route, not just the resolver unit test. The leak test file now covers all of: bootstrap 200, `/v1/chat` absent→400 / non-member→403, `/v1/events` not-403. Green.

PR #592, opened 2026-06-30
fix(prompt): route every containment tag through one escaping primitive

mgoldsborough:
## Security fix — two prompt-injection breakouts shipping today

The system prompt fences untrusted content (bundle authors, tenant skills, workspace overlays) inside XML containment tags, escaping the closing form so a body cannot break out. That "escape the tag you open" rule was **disciplinary**, and it had silently failed at two sites that wrap untrusted bodies with **no escape at all**:

- **`<skill-instructions>`** (the matched-skill body) — live via `composeSystemSegments`. A skill body containing `</skill-instructions>` closes the fence early; everything after it is read as top-level system prompt. The existing test actively asserted the raw (unescaped) body as correct.
- **`<app-description>`** (the bundle description) — same breakout, one line.

Of ~11 opened containment tags, 9 escaped and these 2 did not — exactly the tag nobody remembered.

## Change

- New `wrapContained(tag: ContainmentTag, body)` — the open tag, escaped close, and trailing close all derive from **one argument**, so they cannot diverge; the tag is a **closed union**, so an unknown/typo tag is a compile error.
- The escape is **case-insensitive and whitespace-tolerant** (`</TAG>`, `</tag >`, `</ tag>`, `</tag\n>`) — the consumer is an LLM, a fuzzy parser, not a conforming XML reader, so an exact-substring `replaceAll` was itself evadable.
- All 11 sites routed through it (the 2 above gain an escape they never had).
- A new `check:code-style` pass (allow-list derived from the `ContainmentTag` union via AST) **fails CI** on any raw containment-tag open outside `wrapContained` — the invariant is now structural, not disciplinary.
- Stable tag names retained deliberately (a per-render nonce would bust the 1-hour prompt-cache prefix every turn).

## Tests

Flipped the test that asserted the breakout; added an `<app-description>` breakout case and an exhaustive table-driven evasion suite over the tag union, guarded by a compile-time `Record<ContainmentTag, true>` table so adding a tag without a test fails to compile. `bun run check` / `lint` / `check:code-style` pass; `prompt-injection.test.ts` is green (136 pass). From `research/architecture-designs/03-containment-escape.md`.

### DIFF
--- a/scripts/check-code-style.ts
+++ b/scripts/check-code-style.ts
@@ -133,30 +133,28 @@ function extractContainmentTags(source: ts.SourceFile): string[] {
 
 function checkContainmentTagOpens(): CheckResult {
   const rule = "containment-tags-via-wrapContained";
-  const file = join(SRC_ROOT, "prompt", "compose.ts");
-  const content = readFileSync(file, "utf-8");
-  const source = ts.createSourceFile(file, content, ts.ScriptTarget.Latest, true);
 
-  const tags = extractContainmentTags(source);
+  // The `ContainmentTag` union and `wrapContained` live in compose.ts; derive
+  // the allow-list from the union there so the rule and the type stay in
+  // lockstep. The check itself scans EVERY source file — a hand-rolled open in
+  // any other module (e.g. the `effective_context` tool's historical branch)
+  // is the same breakout class, so the guarantee is only structural if the
+  // scan is project-wide rather than one hardcoded file.
+  const composeFile = join(SRC_ROOT, "prompt", "compose.ts");
+  const composeContent = readFileSync(composeFile, "utf-8");
+  const composeSource = ts.createSourceFile(
+    composeFile,
+    composeContent,
+    ts.ScriptTarget.Latest,
+    true,
+  );
+  const tags = extractContainmentTags(composeSource);
   if (tags.length === 0) {
     return {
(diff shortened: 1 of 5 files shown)


mgoldsborough:
Follow-up `00cce760` since the first review pass:
- Fixed a **12th** containment open the original diff missed — the `effective_context` tool's historical branch (`tools/platform/compose.ts`) wrapped an untrusted `<layer3-skill>` body with no escape. Now routes through `wrapContained`.
- Broadened the `check:code-style` guard from one hardcoded file to **all of `src/`** (the `wrapContained`-body exclusion applies only in the defining file), so a hand-rolled open anywhere fails CI. Verified it flags the historical site when reverted.
- Widened the escape to tolerate `< /tag>` (whitespace between `<` and `/`), matching the lint detector; added that form to the evasion corpus and a behavioral test that the historical layer text escapes a breakout payload.

Test counts now: `prompt-injection.test.ts` 148 pass, `compose-effective-context.test.ts` 11 pass (the PR body's "136" predates this follow-up).

The `<conversation-summary>`/`<conversation-transcript>` fences (`compaction.ts`, `auto-title.ts`) flagged in review use a separate, weaker escape and a different tag vocabulary — tracking as a dedicated follow-up rather than expanding this PR's scope.

Issue #586, opened 2026-06-29
Ground truth: bind conversations/files/automations to their workspace (retire stale 'outlive') — offboarding must revoke access


mgoldsborough:
## Not a bug — a deliberate Stage-1 decision that #584 changes the impact of

Stage 1 made conversations **outlive their workspace context** on purpose. It's documented and E2E-tested: `test/integration/stage-1-cross-workspace-conversation.test.ts` — *"a conversation created in workspace A survives Alice being removed from A."* The model:

- A conversation is **owner-gated**, not membership-gated (`canAccess` is pure ownership; resume checks `ownerId` only, `runtime.ts:1193`).
- Membership is enforced on the **focused** workspace at the HTTP boundary (`X-Workspace-Id` → non-member 403). So a removed member **cannot chat focused on A**.
- But they **can** still read and continue conversations they own that originated in A (the "outlive" promise; step 6 of the test continues the conversation from another workspace).

## What #584 changes

Before #584, tools followed the **focused** workspace — so a removed member resuming an A-conversation got the *focused* workspace's tools, never A's. After #584, tools follow the **conversation's** workspace — so resuming an owned A-conversation now resolves **A's tools** (connectors, etc.), even for a member who has left A.

So #584 doesn't open a new *resume* path (that's the intended "outlive" behavior) — it changes **which tools** that resume grants: from the focused workspace's to the conversation's. For a left-the-workspace user, that's action access to A's connectors they arguably shouldn't retain.

## The decision (this is a product call, not a fix)

**A. Keep "outlive" as-is.** A removed member can continue their owned A-conversations and gets A's tools. Simple, consistent with the documented Stage-1 promise and with files (#574). Offboarding, if it matters, is handled by a deliberate archive/transfer of owned conversations on removal — a separate workflow, not a hot-path gate.

**B. Revisit "outlive" for the tool dimension.** On resume into a workspace the owner has left, deny (or downgrade to read-only with **no** workspace tools — identity tools only). Closes the connector-access concern but **reverses the documented Stage-1 promise** and needs its test + docstring rewritten. Must be decided consistently for files and automations too.

A prior attempt to just add a membership-on-resume gate (option B, deny) was reverted on #584 because it silently reversed the Stage-1 test. If we go with B, it should be its own change with the Stage-1 promise explicitly re-decided — not smuggled into the resolution-seal PR.

[...]

PR #595, opened 2026-06-30
fix(runtime): resuming a conversation requires current workspace membership

mgoldsborough:
Part of **#586** — the **conversations** slice. Files and automations carry the same offboarding shape and follow separately (#586 stays open for them).

## Why

A conversation is sealed to its workspace (#584): on resume its tools/skills/apps resolve in `convWsId`. But the resume gate was **ownership-only**, so a member removed from a workspace could resume conversations they own there and get that workspace's **tools/connectors** — ambient authority into a workspace they were offboarded from. That's the trust-boundary hole the chief-architect audit on #586 flagged: the wall's safety rests on "the workspace was membership-validated at session establishment," but resume validated only ownership.

The "conversations outlive their workspace context" stance that permitted this is a **vestige** of when conversations lived at flat top-level storage *outside* any workspace. They now live *inside* the workspace and are archived with it on delete — so active use should be bound to membership.

## What

Ownership is now necessary but not sufficient on resume:

- Both `chat()` and `startTurn()` re-check membership of the conversation's workspace and throw `ConversationWorkspaceAccessDeniedError` → **403** for a non-member (in `startTurn`, *before* it reserves a run).
- A per-**resume** check (once per conversation load, at session establishment — exactly where the wall says the workspace must be membership-validated), **not** the per-call scan the wall forbids.
- Personal workspaces are sole-member by construction, so they never gate.

**READ stays owner-gated** — a removed member can still load and read their own authored conversations (`findConversation`, the SSE stream). Only active continuation is gated. (Revoking read too would be an additive offboarding sweep + an audited export flow — deliberately not this change; see #586 Q1.)

## The decision (chief-architect verdict on #586)

This is the gate **#584 built and reverted** because it broke the Stage-1 "outlive" E2E test. That stance is now **retired** as a stale premise: the test is **inverted** to assert a removed member *cannot* resume (reads still pass), and the contradictory CHANGELOG "outlive" claim is dropped. The #584 seal stays — it's what *exposed* this, correctly.

## Tests

- `resume-requires-workspace-membership.test.ts` — `chat()` + `startTurn()` denial, current-member allow, personal-workspace exemption.
- `stage-1-cross-workspace-conversation.test.ts` — inverted: removed member's resume → 403; read assertions + the `canAccess` ownership-only secondary still pass.
- `conversation-access.test.ts` — setup now establishes Alice's membership (it relied on owner-only resume).
- `bun run verify` green (605 integration pass).


mgoldsborough:
Suggestion 1 (stale step-1 comment) — fixed (9ccfd89). The comment described the pre-#584 model (conversation stored under the personal workspace, header scopes only the briefing); the test's own assertions disprove it (file under `workspaces/<sharedA>/`, `workspaceId === sharedA`). Replaced with the accurate "born in the focused X-Workspace-Id (sharedA), sealed to it" framing. Comment-only; 2 pass.

PR #599, opened 2026-06-30
refactor(web): delete the dead speaker-label machinery (participantMap)

### LINKED ISSUE #588: Delete the dead speaker-label machinery (participantMap) — vestige of the removed multi-participant feature
## Verdict: dead code, not a bug — delete it

Audit confirms the speaker-label path is **unreachable** in the single-owner product. Stage 1 removed multi-participant conversations (`shareConversation` / `unshareConversation` / `addParticipant` / `removeParticipant` gone; `Conversation.visibility` / `participants` purged from the schema — CHANGELOG bullets ~120-123; commits `6594e23c`, `09fc0fa0`, PR #244/#570). Since then every persisted user message is stamped `userId = ownerId`, and the owner-only auth gate forces `ownerId === currentUserId`. So `MessageList`'s `shouldShowSpeaker` hits `msg.userId === currentUserId → return false` every time: the label never renders and `participantMap` is never consulted.

(So there is no live "non-admins see raw userIds" symptom either — the label doesn't render at all. This is purely vestigial.)

## Scope (frontend only — net deletion)

- `web/src/context/ChatContext.tsx`: remove the `participantMap` state, the `manage_users` mount effect that builds it, the `ChatConfigContextValue.participantMap` field, and both `configValue` entries. **Keep** `manage_users` itself — it's still used by the settings tabs (`UsersTab`, `OrgUsageTab`, `WorkspaceDetailPage`, `WorkspaceMembersTab`); only the ChatContext fetch goes.
- `web/src/components/ChatPanel.tsx`: drop `participantMap` from the destructure and the prop pass-down.
- `web/src/components/MessageList.tsx`: remove `shouldShowSpeaker`, the `participantMap` prop, the label render block, and the `participantColor` border tint — all dead once the only label producer is gone.
- Update `web/src/__tests__/chat-panel-workspace-rescope.test.tsx` (its mock comment references the `manage_users` mount fetch).

## Do NOT touch

The backend per-message `userId` field (`src/conversation/types.ts`) is **deliberate forward-compat** for the deferred Stage 4 sharing feature (see comments in `compose.ts`, `handlers.ts`). Leave it. This cleanup is frontend-only.

(Re-scoped from a server-side "self-describing messages" fix — that was based on the wrong premise that attribution was live.)

mgoldsborough:
Closes #588.

## Dead code, confirmed

The speaker-label path is unreachable. Stage 1 removed multi-participant conversations (`shareConversation` / `addParticipant` / the `participants` schema are gone), so every persisted message now has `userId === ownerId === currentUserId`. `MessageList`'s `shouldShowSpeaker` early-returns on `msg.userId === currentUserId` every time — the label never rendered and `participantMap` was never consulted. (It was also built from an **org-admin-only** `manage_users` fetch, so a non-admin's fetch 403'd into an empty map regardless.)

## Removed (net deletion, frontend only)

- **ChatContext** — the `participantMap` state, the `manage_users` mount fetch, the `ChatConfigContextValue.participantMap` field, and its `configValue` entries.
- **ChatPanel** — the `participantMap` + `currentUserId` props it only passed through to `MessageList` (and the now-unused `useChatConfigContext` import).
- **MessageList** — `shouldShowSpeaker`, the participant border tint, the speaker-label render block, and the now-unused `currentUserId` prop.
- **`web/src/lib/participant-colors.ts`** — its only consumer was the speaker label.
- The two tests that asserted the removed fetch / field.

## Kept

- The **backend** per-message `userId` field — deliberate forward-compat for the deferred Stage 4 sharing feature. This is a frontend-only deletion.
- `currentUserId` on the context — it stamps the outgoing message's `userId` in `chat-store` and drives the identity-change reset; only its dead pass *into MessageList* was removed.
- `manage_users` itself — still used by the settings tabs (Users, Org usage, Workspace members).

`bun run verify` green (596 integration + 581 web pass).


mgoldsborough:
Suggestion 1 (guard the surviving `userId` plumbing) — added (68df443). Comment at the stamp site (`chat-store.ts`) noting the per-message `userId` is intentional Stage-4 forward-compat with no UI consumer today, so it isn't mistaken for dead code and removed. Comment-only; tsc clean.

PR #601, opened 2026-07-01
fix(web): remove the composer footer (stale post workspace-wall)

mgoldsborough:
## What

Removes the chat **composer footer** — the two-line strip below the message input showing `Not viewing an app` and `TOOLS FROM: <workspace badges>`.

## Why it's a bug

The `TOOLS FROM` line rendered one badge per workspace the user is a **member of** — sourced from the full membership list (`WorkspaceContext.workspaces`, i.e. `manage_workspaces { action: "list" }`), never scoped to the conversation.

That was correct under the **pre-wall Stage 2 aggregator model**, where a chat aggregated tools from every workspace the identity belonged to (the component's own header comment says so; both files were born in `#272`). The **workspace wall** (#537) then sealed each conversation to **exactly one workspace + identity tools** (`runtime.listToolsForWorkspace`), but never touched this footer. So the line advertises tools from every workspace the user belongs to (e.g. 5 badges) when the agent can actually only reach **one**.

Rather than rescope the data source to the single bound workspace, the footer as a whole is low-value chrome (the second line just restates the route), so it's dropped entirely.

## Changes

- Delete `ComposerFooter.tsx`, its now-unused `ToolWorkspacesContext.tsx` seam (consumed only by the footer), and `ComposerFooter.test.tsx`.
- `Composer` becomes a thin `MessageInput` wrapper; drop the now-meaningless `hideFooter` prop and its `ChatPanel` pass-through.
- Drop the deleted file from the namespace audit-grep test's target list.

## Verification

- `bun run verify` → exit 0 (static + unit + web + integration; 596 pass / 12 skip / 0 fail)
- `cd web && bun run build` → exit 0


mgoldsborough:
Addressed the optional suggestion in `d5d14e43`: with the footer gone, `Composer` was pure 1:1 indirection (identical props, no-op wrapper div, speculative doc comment), so rather than keep the seam I collapsed it — `ChatPanel` now mounts `MessageInput` directly and the audit-grep test drops the `Composer.tsx` target. `bun run verify` green (596 pass / 0 fail), web build clean. No blocking items outstanding.

PR #602, opened 2026-07-01
web: compress and long-cache static SPA assets in Caddy

mgoldsborough:
## Problem

The web tier serves the built SPA **uncompressed and uncached**. The main chunk is **2,541,484 bytes (~2.5 MB) on the wire** — verified live: no `content-encoding` even when the client sends `Accept-Encoding: br, gzip`, and no `Cache-Control` on the content-hashed assets. It's a render-blocking resource (the app can't paint until it loads), so first open is slow on any non-fast link, and every subsequent open re-downloads the full bundle.

Measured on the current bundle:

| | Bytes | vs raw |
|---|--:|--:|
| raw (today) | 2,541,484 | 1.0× |
| gzip -9 | 669,556 | **3.8×** smaller |
| brotli -11 | 553,558 | **4.6×** smaller |

## Change

Scoped to the static-file `handle` block only:

- **`encode zstd gzip`** — 4–5× smaller render-blocking payload.
- **`Cache-Control: public, max-age=31536000, immutable` on `/assets/*`** — filenames are content-hashed (`index-<hash>.js`), so they're safe to cache forever; repeat opens hit browser cache. `index.html` falls through uncached (new deploys picked up immediately); `config.js` keeps its `no-store`.

Compression is **deliberately not** applied to the reverse-proxied `/v1` and `/mcp` routes — those carry the SSE run-event stream and the streamable-HTTP MCP transport, which must not be buffered by an encoder. That's why `encode` lives inside the static handler, not at the site level.

## Not in scope (follow-up)

This is the zero-risk 80/20 (compression). The bundle is 2.5 MB in the first place because **`streamdown` → `mermaid@11` → `d3` + `dagre`** (plus the `marked` / `unified` / `remark` / `rehype` markdown pipeline and `katex`) all land in **one eager chunk** — vite has no `build.rollupOptions.manualChunks` and nothing is dynamically imported, so mermaid/d3 ship to every user even though most turns render no diagram. Splitting the diagram/math renderer behind a dynamic import is the real size fix and wants its own PR + testing.

## Verification

- Caddyfile syntax is standard; `caddy validate` not run (no Caddy CLI locally) — please confirm on image build. Standard Caddy includes both gzip and zstd encoders.
- `bun run verify` does not cover the Caddyfile (config-only change).

### DIFF
--- a/web/Caddyfile
+++ b/web/Caddyfile
@@ -38,8 +38,16 @@
 		header Cache-Control "no-store"
 		respond `window.__NB_CONFIG__ = {"tenantId":"{$NB_TENANT_ID:}","environment":"{$NB_SENTRY_ENV:}","release":"{$NB_VERSION:}","sentry":{"dsn":"{$NB_SENTRY_DSN:}","enabled":"{$NB_SENTRY_ENABLED:false}","tracesSampleRate":"{$NB_SENTRY_TRACES_SAMPLE_RATE:0}"},"posthog":{"key":"{$NB_POSTHOG_KEY:}","enabled":"{$NB_POSTHOG_ENABLED:false}"}};` 200
 	}
+	# Static SPA assets. Compression + immutable caching are scoped to THIS handler
+	# only — deliberately not applied to the reverse-proxied /v1 and /mcp routes,
+	# whose SSE/streamable-HTTP responses must not be buffered by an encoder.
 	handle {
+		encode zstd gzip
 		root * /srv
+		# Content-hashed build output (index-<hash>.js/.css) is immutable — cache hard.
+		# index.html itself falls through uncached so a new deploy is picked up at once.
+		@immutable path /assets/*
+		header @immutable Cache-Control "public, max-age=31536000, immutable"
 		try_files {path} /index.html
 		file_server
 	}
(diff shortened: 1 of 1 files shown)


mgoldsborough:
Thanks for the thorough review — agree it's merge-ready, and I verified your findings hold (ran `caddy validate` on `caddy:2-alpine` and a runtime header check per path).

Addressed **suggestion #1** in `955f388`→latest: index.html now gets an explicit `Cache-Control: no-cache` so the "new deploy picked up at once" promise is real instead of relying on heuristic freshness. Done via a matcher mutually exclusive with `/assets/*`, so each response carries exactly one `Cache-Control` (asset is never both immutable and no-cache).

Verified on `caddy:2-alpine` (the Dockerfile base):
- config validates
- `/assets/*.js` → `immutable` + `Content-Encoding: gzip`
- `/`, SPA routes (`/w/...` via try_files), and `/favicon.png` → `no-cache` (favicon correctly not immutable)
- exactly 1 `Cache-Control` header on assets (no conflict)

**#2** (immutable header on a missing `/assets/<hash>` fallback) — left as-is; impact is nil (content-hashed URL, never re-requested) and the matcher split doesn't change it.

**#3** (CI `caddy validate` step) — good idea, out of scope for this PR; happy to open a follow-up so a future Caddyfile typo can't reach the image build.

PR #605, opened 2026-07-01
ci: validate web Caddyfile

### LINKED ISSUE #604: web: add a CI check that validates web/Caddyfile
## Background

`web/Caddyfile` configures the production web tier — SPA routing, the `/v1` and `/mcp` reverse proxies, gzip/zstd compression, and cache headers (see #602). There's currently **no validation** of this file in CI, so a syntax typo would sail through review and only surface when the image is built/deployed. A tiny CI step would catch that up front.

## Task

Add a CI check that runs `caddy validate` on `web/Caddyfile` and fails the build if it's invalid.

## Suggested approach (scaffolded — this is a good starting point, adapt as you see fit)

- **Where:** a new job in `.github/workflows/ci.yml`, gated with a `paths:` filter so it only runs when `web/Caddyfile` changes (no need to run it on every PR). A small standalone workflow is fine too.
- **How:** validate with the **same image the Dockerfile ships** (`web/Dockerfile` ends in `FROM caddy:2-alpine`), so CI matches production. The command that already works:

  ```bash
  docker run --rm -v "$PWD/web/Caddyfile:/etc/caddy/Caddyfile:ro" \
    caddy:2-alpine caddy validate --config /etc/caddy/Caddyfile --adapter caddyfile
  ```

  (GitHub-hosted runners have Docker available, so you can run this directly in a `run:` step.)

- **Gotcha to save you time:** the Caddyfile uses `{$PLATFORM_URL}` env placeholders. `caddy validate` handles unset placeholders fine (they default to empty), so **no env setup is needed** — it validates as-is.

## How to prove it works

1. With your job in place, the current `web/Caddyfile` should validate ✅.
2. Temporarily introduce a typo (e.g. change `encode zstd gzip` to `encode zztd gzip`), push, and confirm the CI job **fails** ❌.
3. Revert the typo.

## Acceptance criteria

- CI validates `web/Caddyfile` and **fails** on an invalid config, **passes** on a valid one.
- The check runs only when `web/Caddyfile` (or the workflow) changes.
- Uses `caddy:2-alpine` to match the deployed image.

Context: this came out of the #602 review — the reviewer validated the Caddyfile by hand on `caddy:2-alpine`; this issue makes that a permanent CI guard.

MFA-G:
## Summary
- add a focused Caddyfile CI workflow for changes to web/Caddyfile
- validate with caddy:2-alpine, matching the production web image
- include the workflow file in the path filter so updates to the check run it too

Closes #604

## Validation
- Not run locally: Docker is not installed in this environment (docker: command not found).

### DIFF
--- a/.github/workflows/caddyfile-ci.yml
+++ b/.github/workflows/caddyfile-ci.yml
@@ -1,0 +1,30 @@
+name: Caddyfile CI
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - "web/Caddyfile"
+      - ".github/workflows/caddyfile-ci.yml"
+  push:
+    branches: [main]
+    paths:
+      - "web/Caddyfile"
+      - ".github/workflows/caddyfile-ci.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  validate-caddyfile:
+    name: Validate Caddyfile
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: Validate web/Caddyfile
+        run: |
+          docker run --rm \
+            -v "$PWD/web/Caddyfile:/etc/caddy/Caddyfile:ro" \
+            caddy:2-alpine \
+            caddy validate --config /etc/caddy/Caddyfile --adapter caddyfile


mgoldsborough:
Hi @MFA-G - thanks for the submission. Taking a look now.

In the meantime, give us a star! ⭐  😄

mgoldsborough:
Ran the validation locally (the PR notes Docker wasn't available in the author's env), so here's the evidence that the guard actually fails closed — all against `caddy:2-alpine`, the exact image `web/Dockerfile:16` ships:

| Case | Command result |
|---|---|
| **Broken** Caddyfile (`encode zstd` → `encode zztd`) | `caddy validate` → **exit 1**, `Error: … module not registered: http.encoders.zztd, at /etc/caddy/Caddyfile:45` → step fails → **CI fails** ✅ |
| Valid Caddyfile (current `main`) | exit **0** ✅ |
| Valid Caddyfile with **`PLATFORM_URL` unset** (this workflow sets no env) | **passes** — the `{$PLATFORM_URL}` placeholder resolves to empty and validate still succeeds, so no false-fail on the good config ✅ |

Exact reproduction:

```bash
docker run --rm -v "$PWD/web/Caddyfile:/etc/caddy/Caddyfile:ro" \
  caddy:2-alpine caddy validate --config /etc/caddy/Caddyfile --adapter caddyfile
echo "exit=$?"
```

Mechanics also check out: single `docker run` step (no `|| true` / `continue-on-error`, so the exit code propagates), and the `paths:` filter includes the workflow file itself, so it self-validates on this PR. LGTM.

